PR #5, opened 2020-10-21
Fix clippy error of using sort_unstable instead of sort

bergwolf:
Using sort_unstable instead of sort for primitive type will get better performance.


imeoer (approved):
LGTM

jiangliu (approved):

PR #13, opened 2020-10-22
Fix: incorrect nydusd config

imeoer:
- Keep `backend` and `cache` fields in `device`, move other fields to outside;
- Fix incorrect JSON syntax rendering for Github markdown;

### DIFF
--- a/docs/nydusd.md
+++ b/docs/nydusd.md
@@ -69,15 +69,14 @@ mount -t virtiofs nydus /mnt
 
 #### Common Fields In Config
 
-``` JSON
+```
 {
   "device": {
     "backend": {
       // localfs | oss | registry
-      "type": "...",
+      "type": "localfs",
       "config": {
-        ...
-        // Access remote storage backend via P2P proxy, for example Dragonfly client address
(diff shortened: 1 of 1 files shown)


docs/nydusd.md
```diff
@@ -69,15 +69,14 @@ mount -t virtiofs nydus /mnt
...
+```
 {
   "device": {
     "backend": {
       // localfs | oss | registry
-      "type": "...",
+      "type": "localfs",
       "config": {
-        ...
-        // Access remote storage backend via P2P proxy, for example Dragonfly client address
```
changweige, line 80: Though I am not inclining toward adding such 'fake' comments as it is not Json grammar, we still merge this right now. Hopefully, oneday, we can split that part out from the examples and show a proper Json.

changweige (approved):

docs/nydusd.md
imeoer: Thanks, @changweige, it's a good suggestion, we can put these comments to outside, or support `yaml` or `toml` config file.

PR #22, opened 2020-11-04
Initial support for image format - Stargz

changweige:
Within this pach series, we aim at implementing the Stargz image format along with essential smoke test.


imeoer (approved):
LGTM

PR #25, opened 2020-11-18
Nydusify: refactor without containerd service

imeoer:
In previous, Nydusify use containerd service to pull, push image and manipulate image manifest, but containerd is a black box for image manipulation, we can't track the error of layer-leveled pulling, pushing conveniently, and it has performance issue for layer read/write/unpack manipulation when using it as a middle service. This patch removed the containerd service dependency for Nydusify:

- Support to show realtime progress bar for pulling, building and pushing;
- Remove `--source-auth` and `--target-auth` CLI options, change to use docker config for source/target registry auth automatically;
- Use `go-containerregistry` package to pull, push and manipulate image, manifest instead of containerd service, so that Nydusify has better compatibility;
- Support `--multi-platform` CLI option to add manifest index (multiple platforms, OCI & Nydus) for nydus image;


changweige (approved):
Looks good.

PR #36, opened 2021-01-04
update nydusify.md and add GO111MODULE flags for nydusify build

ydcool:
Signed-off-by: Dominic Yin <yindongchao@inspur.com>

PR #38, opened 2021-01-05
CI: use rust 1.47.0 environment

imeoer:
Use a fixed rust version in the CI environment to make testing stable,
we can move forward once decided to upgrade the rust version in project.

Fix CI issue on #36 


jiangliu (approved):
that's great, a recommend rustc version:)

PR #49, opened 2021-01-22
example: use standalone nydusd mode

bergwolf:
`--shared-daemon` is not working due to API incompatibility.
Let's use standalone mode for now until that is fixed.

issue #44


liubogithub:
looks good.

liubogithub (approved):

PR #59, opened 2021-02-05
rafs: backend related fixes

eryugey:
This PR fixes two issues
- only use backends when the backend feature is on, otherwise build would fail due to un-imported crates.
- initialize localfs backend's metrics when blob file is specified in config as well


changweige (approved):
Looks good

PR #64, opened 2021-02-20
rafs/prefetch: fix prfetch outage with layered image building

changweige:
Previously, when converting image, a prefetch list is input to
builder from stdin, which is later persisted to bootstrap prefetch
table.

During parsing the prefetch table, builder checks file's existance.
But for layered building, we can't guarantee that the source directory
of parent boostrap is located from local file system.

So builder no longger accecpts abosolute paths, but relative paths to
source rootfs.

It must start with "/"


imeoer (approved):
LGTM

PR #66, opened 2021-03-08
Adapt to new fuse-rs APIs

changweige:
Fuse-rs crate has enhanced its API. Therefore, we firstly adapt to those APIs to catch more metrics and make nydus error handling more convenient.


imeoer (approved):
LGTM!

PR #69, opened 2021-03-08
rafs: xattrs size should also be reflected thru i_blocks

changweige:
When rafs report its file disk usage, it should also include the space xattr is occupying.


imeoer (approved):
LGTM

PR #72, opened 2021-03-08
Change logger from stderrlog to flexi_logger

changweige:
Nydusd now can log both to standard error and specified logging file. Just specify the logging file by adding option `--log-file` when nydusd starts


imeoer (approved):
LGTM

PR #74, opened 2021-03-08
Enrich open-api yaml

changweige:
Nydus api server has evolved a lot. Therefore add more objects to open-api yaml


imeoer (approved):
LGTM

PR #75, opened 2021-03-08
Builder: support --output-json for check subcommand

changweige:
Add `--output-json` option for check subcommand, use it for Nydusify
image checker to get blob ids from bootstrap, and compare with the blob
ids recorded in Nydus image manifest, to ensure the image validity.


imeoer (approved):
LGTM

PR #78, opened 2021-03-09
Nydusify: make converter as a package

imeoer:
Refactor nydusify code to make converter as an independent package,
the package can be imported to any project which wants to implement
a nydus image converter, for example, buildkit. Abstract a provider
interface to adapt to different build environments, the provider includes
components:

  logger: output build progress for nydusify or buildkitd/buildctl;
  remote: create a remote resolver, it communicates with remote registry;
  source: responsible for getting image manifest, config, and mounting layer;

Nydusify provides a default implementation in pkg/converter/provider,
so we can use it in Nydusify directly, but we need to implement it in
buildkit or other any projects which want to import nydusify package.


changweige (approved):
Looks good

PR #79, opened 2021-03-09
Nydusify: make some improvements

imeoer:
Ignore directory size check in checker:

The directory size may be different between various fs environments,
so we need ignore the size check in checker when walking the rootfs
mountpoint.

Fix concurrent push issue:

Concurrently push blob with same digest using containerd
docker remote client will cause error:
`failed commit on ref: unexpected size x, expected y`
use digest leveled mutex lock to avoid the issue.

Adapt test cases for nydusify package:

Use default provider in nydusify to replace remote and source
accessor in e2e test, add a case to test build cache.


changweige (approved):
Looks good.

PR #80, opened 2021-03-09
Nydusify: simplify parser to allow importing by external

imeoer:
Nydusify parser parses OCI & Nydus image manifest, manifest index, and
image config into Parsed object, we can move it out from checker so that
source provider can use it.


changweige (approved):
Looks good.

PR #81, opened 2021-03-09
Nydusify: support init remote with auth

imeoer:
Support init remote instance with base64 encoded auth string
instead of specifying docker config, useful for integration with
other projects which want to convert nydus image.

Add a nydusify package usage example to show how to convert
nydus image.


changweige (approved):
Looks good

PR #85, opened 2021-03-16
github: add github release action

bergwolf:
To automatically build and upload static binaries upon release tag
push events.

When we push a new tag like `v10.10.10`, the action would build static binaries and create a new release like https://github.com/bergwolf/image-service/releases/tag/v10.10.10

Example run:
https://github.com/bergwolf/image-service/actions/runs/657438261

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,84 @@
+name: release
+
+on:
+  push:
+    tags:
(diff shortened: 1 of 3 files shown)


imeoer (approved):
LGTM!

imeoer:
We can tag a release next :)

PR #86, opened 2021-03-17
Support snapshotter reconnect to already running daemons

renzhengeek:
This patch set includes fixes as below:
- persist daemon info in database, so that snapshotter can reconnect to running daemons;
- output daemon's log to daemon's own log file;
-  umount daemon with API when daemon in shared mode.

Signed-off-by: 慕陶 <jihui.xjh@antfin.com>
Signed-off-by: Eric Ren <renzhen@linux.alibaba.com>


changweige (approved):
LGTM

PR #87, opened 2021-03-17
snapshotter: rename snapshotter name to be easily imported

changweige:
Current snapshotter module name makes it hard to be imported to
other projects like containerd.
So rename it


imeoer (approved):
LGTM

PR #90, opened 2021-03-19
Nydusify: add default platform to existed oci manifest

imeoer:
**Nydusify: add default platform to existed oci manifest**

```
Add default platform "linux/amd64" to existed manifest in
remote registry if pushing Nydus image as a manifest index
(multi-platform enabled), so that docker/containerd can
distinguish OCI and Nydus image using the platform field
in the manifest index.
```

**Nydusify: support --multi-platform option for checker**

```
Ensure target image represents as a manifest list, and it consists
of OCI and Nydus manifest, if the "--multi-platform" option of checker
be specified.
```


changweige (approved):
LGTM

PR #97, opened 2021-03-31
nydusd: add virtual mountpoint commandline option

bergwolf:
So that we can start nydusd with an initial fs backend and still be able
to attach more fs backends to it later on.


changweige (approved):
looks good

PR #106, opened 2021-04-13
snapshotter: add initial metrics support

renzhengeek:
To know about the image access IO stats/pattern such as read size, read latency and file operation distributions, we add this metrics feature that can provide prometheus-format metrics, and mainly focus on the filesystem metrics currently.
    
    - How to enable metrics
    
    o For grpc plugin binary
    Add "--enable-metrics" option in commandline
    
    o For containerd builtin snapshotter
    Add "enable_metrics = true" in containerd config file
    
    - How to get the metrics
    
    Two ways to get metrics as below:
    
    o metrics http server
    
    RootDir="/var/lib/containerd/io.containerd.snapshotter.v1.nydus/"
    metricsSock="$RootDir/metrics.sock"
    sudo curl --unix-socket $metricsSock http://localhost/metrics
    
    o metrics log file path:
    
    $RootDir/metrics.log


imeoer (approved):
LGTM

PR #109, opened 2021-05-07
Introduce secure-container support to nydus-snapshotter and various bug fixes

eryugey:
This PR mainly brings the secure-container support to nydus-snapshotter, which introduces a new "none" instance mode, and snapshotter returns the rafs bootstrap file instead of a mounted rafs dir, so that secure-container such as kata-containers could parse the bootstrap file directly.


imeoer (approved):
LGTM

PR #112, opened 2021-05-30
fix nydusd daemon process double wait and change p.Kill() to     p.Signal(syscall.SIGTERM) when destroy nydusd daemon process

luodw:
There are two process wait, when destroy daemon, if first wait function run first, the second wait function will fail and output error as below. It is happened occasionally.
image

1. https://github.com/dragonflyoss/image-service/blob/v0.1.0/contrib/nydus-snapshotter/pkg/process/manager.go#L124
2. https://github.com/dragonflyoss/image-service/blob/v0.1.0/contrib/nydus-snapshotter/pkg/process/manager.go#L188

The other question is here why not use p.Signal(syscall.SIGTERM) to gracefully exit?  p.Kill() will kill -9.
https://github.com/dragonflyoss/image-service/blob/v0.1.0/contrib/nydus-snapshotter/pkg/process/manager.go#L184

### DIFF
--- a/contrib/nydus-snapshotter/pkg/process/manager.go
+++ b/contrib/nydus-snapshotter/pkg/process/manager.go
@@ -121,8 +121,7 @@ func (m *Manager) StartDaemon(d *daemon.Daemon) error {
 		return err
 	}
 	d.Pid = cmd.Process.Pid
-	// make sure to wait after start
-	go cmd.Wait()
+	// process wait when destroy daemon and kill process
 	return nil
 
 }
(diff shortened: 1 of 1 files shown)


bergwolf:
@luodw Thanks for the fix! And you are right that nydus-snapshotter should just send `SIGTERM` to terminate nydusd, as being handled at https://github.com/dragonflyoss/image-service/blob/master/src/bin/nydusd/main.rs#L433

PS, please add your SOB line to the commit message so that it can pass DCO test. It would be better to amend the commit message to include the texts in the PR as well. So that we know better why we made the change later on. Thanks!

bergwolf (approved):
Thanks!

PR #113, opened 2021-06-04
snapshotter: change CleanupDatabase to CleanupDaemons

luodw:
As boltdb is shared by daemon and snapshot/blob, so it is not
suitable for method name CleanupDatabase when clean daemons

### DIFF
--- a/contrib/nydus-snapshotter/pkg/process/manager.go
+++ b/contrib/nydus-snapshotter/pkg/process/manager.go
@@ -250,7 +250,7 @@ func (m *Manager) Reconnect(ctx context.Context) error {
 
 	// cleanup database so that we'll have a clean database for this snapshotter process lifetime
 	log.L.Infof("found %d daemons running", len(daemons))
-	if err := m.store.CleanupDatabase(ctx); err != nil {
+	if err := m.store.CleanupDaemons(ctx); err != nil {
 		return errors.Wrapf(err, "failed to cleanup database")
 	}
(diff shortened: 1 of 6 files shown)


contrib/nydus-snapshotter/pkg/process/manager.go
```diff
@@ -250,7 +250,7 @@ func (m *Manager) Reconnect(ctx context.Context) error {
 
 	// cleanup database so that we'll have a clean database for this snapshotter process lifetime
 	log.L.Infof("found %d daemons running", len(daemons))
-	if err := m.store.CleanupDatabase(ctx); err != nil {
+	if err := m.store.CleanupDaemons(ctx); err != nil {
 		return errors.Wrapf(err, "failed to cleanup database")
```
xujihui1985, line 254: may be sync with the method name

luodw: you mean it should add lock in CleanupDaemons？Maybe I misunderstand and  create new a branch use cleanup-daemons.

xujihui1985: > errors.Wrapf(err, "failed to cleanup database") 
sorry, I mean the error message should sync with the method name, it's trivial

luodw: oh, I see, should fix next pr.

PR #114, opened 2021-06-06
snapshotter: change CleanupDatabase to CleanupDaemons

luodw:
As boltdb is shared by daemon and snapshot/blob, so it is not
suitable for method name CleanupDatabase when cleanup daemons


bergwolf (approved):
Thanks @luodw !

PR #115, opened 2021-06-10
Sync builder, metric and prefetch changes

imeoer:
nydus-image: delete blob when finish building
nydus-image: introduce BlobBufferWriter

Nydusify: ignore empty blob generated by builder
Nydusify: preferred use of existing oci manifest
Nydusify: fix lost blob layers in manifest

export latest read files to array instead of object
merge api /metrics/latest_read_files & /metrics/files
metrics: support latest read files metrics

snapshotter: fix unmount error on overlay snapshot dir

fix: performance degraded when disabling prefetch
blobcache/prefetch: user io also take limiter tokens

storage/oss: remove deadcode from oss - function create_bucket


changweige (approved):
Looks good. Thanks

PR #116, opened 2021-06-10
Builder: refactor code structure

imeoer:
Refactor code structure for builder core modules and introduce a new Builder trait, it accepts build context parameter, then builds source into bootstrap(for directory and stargz source) and blob(only for directory source). we can also implement Builder for diff builds source in the future.

And sync some improvement changes for nydusify.


changweige (approved):
Looks good. Thanks

PR #117, opened 2021-06-10
Cache: implement chunk cache in bitmap

imeoer:
```
This PR for the sync of main feature.

Introduce new ChunkMap trait, It has two implementations, IndexedChunkMap
and DigestedChunkMap.

The IndexedChunkMap is an implementation that uses a file as bitmap
(like HashMap<chunk_index, has_ready>). It creates or opens a file with
the name $blob_id.chunk_map which records whether a chunk has been cached
by the blobcache. This approach can be used to share chunk ready state
between multiple nydusd instances, which was not possible with the previous
implementation using in-memory hashmap.

The DigestedChunkMap is an implementation that uses a hash map
(HashMap<chunk_digest, has_ready>) to records whether a chunk has been
cached by the blobcache. it is used to be compatible with the previous old
nydus bootstrap format.

According to a simple performance test, IndexedChunkMap performs 17 times
faster than DigestedChunkMap.
```


changweige (approved):
Looks good. Thanks

PR #120, opened 2021-07-01
snapshotter: support oss and localfs backend as well

eryugey:
Currently, nydus snapshotter assumes the backend is registry, as
NewDaemonConfig() doesn't check backend type and always set 'Host',
'Repo' and 'Auth' fields that are only useful for registry backend.

So add oss and localfs backend configs to Backend config as well, so
images with such backends could be used correctly.

Signed-off-by: Eryu Guan <eguan@linux.alibaba.com>


imeoer (approved):
LGTM, thanks!

PR #122, opened 2021-07-15
log information when fuse failed to open

benjaminhuanghuang:
utils/fuse: print more detailed error when mounting fusedev

Description: 
    when I started nydusd in a docker running in MacOS, nydusd existed without any helpful information left in log.

Reason: 
    MacOS dose not support fuse, that causes nydysd failed.

Fix: 
1. Log error message when FUSE failed to open
    
2. This fix cause Rustc 1.49 crash with error 
     " thread 'rustc' panicked at 'called `Option::unwrap()` on a `None` value'"

    So I upgrade the rust tool chain to 1.50
    
3. The Clippy report error:
     error: this function's return value is unnecessarily wrapped by `Result`
       --> utils/src/metrics.rs:411:5

     I fixed it accordingly.

Signed-off-by: Benjamin Huang <benjaminhuanghuang@gmail.com>


liubogithub:
@benjaminhuanghuang 
Thanks for the fix.
DCO is not happy with missing SOB, can you please also put the above description and Signed-off-by in the commit log?
specifically, use `git commit --amend` and then make a force push `git push --force`

utils/src/fuse.rs
```diff
@@ -252,12 +252,25 @@ fn fuse_kern_mount(
...
-    let meta = mountpoint.metadata()?;
+        .open(FUSE_DEVICE)
+    {
+        Ok(v) => v,
+        Err(e) => {
+            error!("FUSE failed to open. {}", e);
+            return Err(e);
+        }
+    };
+    let meta = match mountpoint.metadata() {
```
changweige, line 267: Um. I can't understand why such an update is causing `rustc` crashed.
But I suggest to remove all the existed target directories like `target-fusedev` and `target-virtiofs` from workspace. Then try to re-compile nydus by executing `make` from project home direcotry.
Sometimes, rustc is not stable. :-(

rust-toolchain
```diff
@@ -1 +1 @@
-1.49.0
+1.50.0
```
changweige, line 1: If we truly have to upgrade toolchain. It is fine to me. But better that we can add a separated patch doing so.

utils/src/fuse.rs
```diff
@@ -252,12 +252,25 @@ fn fuse_kern_mount(
...
         .create(false)
         .read(true)
         .write(true)
-        .open(FUSE_DEVICE)?;
-    let meta = mountpoint.metadata()?;
+        .open(FUSE_DEVICE)
+    {
+        Ok(v) => v,
+        Err(e) => {
+            error!("FUSE failed to open. {}", e);
```
changweige, line 263: How about just using error mapping. For example:
```
let file = OpenOptions::new()
        .create(false)
        .read(true)
        .write(true)
        .open(FUSE_DEVICE).map_err(|e| {error!("bla-bla");e})?
```
That is neater and clean

utils/src/metrics.rs
```diff
@@ -484,7 +484,7 @@ pub fn export_files_stats(
             if !latest_read_files {
                 v.export_files_stats()
             } else {
-                v.export_latest_read_files()
+                Ok(v.export_latest_read_files())
```
changweige, line 487: Clippy of version 1.51 or higher does not complain about this anymore.

changweige (changes requested):

changweige:
@benjaminhuanghuang  Can you add a tittle line to your patch commit log?
For example: utils/fuse:  print more detailed error when mounting fusedev

Description:
    ....

imeoer:
@benjaminhuanghuang Thanks for this patch. I noticed that the [build](https://github.com/dragonflyoss/image-service/blob/master/misc/musl-static/Dockerfile) and [smoke test](https://github.com/dragonflyoss/image-service/blob/master/misc/nydus-smoke/Dockerfile) scripts still depend on 1.49.0, could you help change them together?

utils/src/fuse.rs
bhuang-rms: I built my patch in two different environment. rustc crashes in both of them.

rust-toolchain
bhuang-rms: Without upgrading the toolchain, this path can not pass compiling , is that acceptable?

utils/src/fuse.rs
bhuang-rms: yes, I agree

bhuang-rms:
> @benjaminhuanghuang Can you add a tittle line to your patch commit log?
> For example: utils/fuse: print more detailed error when mounting fusedev
> 
> Description:
> ....

The PR has a title. Because a PR is always composed by several commits. It is better to put the  summary text to the title of the PR.

utils/src/fuse.rs
benjaminhuanghuang: Yes, I agree

rust-toolchain
benjaminhuanghuang: Agree

utils/src/metrics.rs
liubogithub: 1.51.0 reports this error too, and even more Clippy errors, I think we can let this go and fix the others in a separate patch later.

changweige: True.

changweige (approved):
Looks good

PR #124, opened 2021-07-23
Nydusify: fix blob push issue when enabling build cache

changweige:
Push blob layer to cache image failed because the blob was deleted earlier:

```
Failed push layer to cache image: Open blob file: open tmp/blobs/x:
no such file or directory
```


imeoer (approved):
LGTM

PR #128, opened 2021-07-23
improve blobcache and rafs prefetch

changweige:
As shared chunk bitmap file is already introduced, in which 1 bit of value 1 indicates that a chunk is successfully downloaded.
Rafs/prefetch no longer has to validate the whole chunk by calculating its sha256, but just to probe if the corresponding it is ever set. This will save unnecessary read from disk.

More over try to deduce locking contention to improve performance.


imeoer (approved):
LGTM, Thanks.

PR #130, opened 2021-07-23
nydus-image inspects bootstrap to analyze rootfs

changweige:
Add a subcommand `inspect` to `nydus-image`.
With this capability, nydus-image can retrieve the meta part of nydus image like listing files, chunks of data distribution, global rafs information, internal management structure like blob table and prefetch table.

It provides two kinds of working mode
- interactive mode, in which a prompt is available and commands are processed interactively.
- request mode, in which `nydus-image inspect` directly receives a command and the output is json formatted, which is to be analyzed and gathered.


imeoer (approved):
LGTM

PR #134, opened 2021-08-04
Prefetch: fix overlapped prefetch merge requests

imeoer:
Two commits, actually:

**Prefetch: fix overlapped prefetch merge requests**

The left chunks whose size is smaller than 4MB will be fetched
after 4MB merge requests.

**Rafs: add ctime support**

Save ctime in inode and use it for ctime/mtime/atime.
There are applications that might check ctime/mtime so
we cannot just drop it and use runtime system time as ctime.

### DIFF
--- a/rafs/src/fs.rs
+++ b/rafs/src/fs.rs
@@ -453,9 +453,13 @@ impl Rafs {
             attr.gid = self.i_gid;
         }
 
-        attr.atime = self.i_time;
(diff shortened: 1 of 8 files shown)


jiangliu:
Should we name it as "ctime" instead of "mtime"?

jiangliu (approved):

PR #140, opened 2021-08-13
storage: add ability to use user-defined blob redirected host

eryugey:
When using registry backend, blob data may be stored on other storages,
    such as OSS, and registry will redirect download request to URL pointing
    to such storage, so backend downloads blob file from there.

    But this redirected link might not be optimal, e.g. we may want to use
    OSS cache endpoint. So introduce a new 'blob_redirected_host' config
    option in Registry backend config, and replace the redirected blob url
    from registry with this user-defined host to have better performance.

    Signed-off-by: Eryu Guan <eguan@linux.alibaba.com>


imeoer (approved):
LGTM!

PR #143, opened 2021-08-16
ctr-remote: add ctr-remote command

luodw:
There are some cases user use ctr to pull image and run container,
so here add ctr-remote command to pull nydus image refer to stargz
https://github.com/containerd/stargz-snapshotter/tree/main/cmd/ctr-remote


bergwolf (approved):
Thanks a lot!

contrib/ctr-remote/commands/rpull.go
```diff
@@ -0,0 +1,148 @@
+/*
+   Copyright The containerd Authors.
+
```
changweige, line 3: Is keeping license header here suitable?

jiangliu: I feel it's OK:)

luodw: As the code is refer to stargz mostly，so keep the license header here

PR #154, opened 2021-08-23
snapshotter: save daemon pid after starting it

bergwolf:
So that the pid is not zero after snapshotter restarts.

Signed-off-by: Peng Tao <bergwolf@hyper.sh>

### DIFF
--- a/contrib/nydus-snapshotter/pkg/process/manager.go
+++ b/contrib/nydus-snapshotter/pkg/process/manager.go
@@ -121,6 +121,11 @@ func (m *Manager) StartDaemon(d *daemon.Daemon) error {
 		return err
 	}
 	d.Pid = cmd.Process.Pid
+	err = m.store.Update(d)
+	if err != nil {
+		// Nothing we can do, just ignore it for now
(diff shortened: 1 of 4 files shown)


imeoer (approved):
LGTM

Bowser1704:
Oops, I fixed it by saving the daemon after starting the daemon in my local repo.

Bowser1704:
@bergwolf Hi, It seems that it doesn't fix this issue due to a duplicate key in `boltdb`.
see log behind.
```
 {"level":"error","msg":"fail to update daemon info (xxx) to db: object with key \"YmRoXgkQRyCmhCxst_LdOQ\" already exists","time":"2021-08-23T14:26:40.825054995+08:00"}
```

Bowser1704:
Looks like we should use `updateObject` not `putObject.`

https://github.com/dragonflyoss/image-service/blob/41250ba681019663380a466048f28d8134d6ea4d/contrib/nydus-snapshotter/pkg/store/database.go#L107-L118

https://github.com/dragonflyoss/image-service/blob/41250ba681019663380a466048f28d8134d6ea4d/contrib/nydus-snapshotter/pkg/store/database.go#L178

PR #155, opened 2021-08-23
Contrib: fix update daemon

Bowser1704:
using updateObject not putObject which will cause error due to duplicate key.

related pr https://github.com/dragonflyoss/image-service/pull/154

Signed-off-by: yuhongqi <yuhongqi@bytedance.com>


bergwolf (approved):
oops, thanks for catching it!

PR #157, opened 2021-09-02
daemon: Turn on the xattr feature of passthrough_fs

zhangjaycee:
The filesystem of overlayfs upperdir needs to support xattr.
Passthrough_fs supports xattr but not enabled by nydusd daemon. This
commit enables it.

Signed-off-by: Jiachen Zhang <zhangjiachen.jaycee@bytedance.com>


bergwolf (approved):
Thanks a lot!

PR #158, opened 2021-09-16
Blobcache handles single inflight backend IO

changweige:
If there is already a backend IO going fetching a chunk from remote storage. Blobcache does not have to issue another one for the same data chunk. Waiting is an ideal behavior saving backend width and reduce IO latency.

When there is several big files residing in Rafs, sequential read will gain performance profit.


imeoer (approved):
LGTM

PR #161, opened 2021-09-16
fuse: drop MS_NOSUID

changweige:
So that suid can take effect and programes like sudo and passwd can
still work on nydus rootfs.

Signed-off-by: Peng Tao <bergwolf@hyper.sh>


imeoer (approved):
LGTM

PR #162, opened 2021-09-16
nydus-image: fix disorder of writing chunks to blob

changweige:
In rafs disk layout, we should enforce a rule that chunks
are dumped/written to blob file according to order of inode
number.

When prefetch files list is specified, e.g. "/", the specified
files will be dumped before dumping other files. But the `readahead_files`
collection is not sorted by its key - inode index.

So fix this by sorting it before consuming it.

This problem might affect prefetch merging possibility.


imeoer (approved):
LGTM

PR #164, opened 2021-09-17
Enhance rafs inspector

changweige:
Thus to show permission bits and human-readable flags


imeoer (approved):

src/bin/nydus-image/inspect.rs
```diff
@@ -569,7 +570,7 @@ Blocks:             {blocks}"#,
     Flags:              {flags}"#,
                 version = sb.version(),
                 inodes_count = sb.inodes_count(),
-                flags = sb.flags()
+                flags = RafsSuperFlags::from_bits(sb.flags()).unwrap()
```
jiangliu, line 573: Is it better to use `RafsSuperFlags::from_bits_truncated()" instead of `unwrap()`?

PR #175, opened 2021-09-26
Fix cargo build artifacts and docker layers are not cached

changweige:
This patch will reduce GitHub/Action running time quite considerably.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ endef
 # Targets that are exposed to developers and users.
 build: .format fusedev virtiofs
 release: .format .release_version fusedev virtiofs
-static-release: .musl_target .format .release_version fusedev virtiofs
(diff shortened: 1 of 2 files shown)


jiangliu (approved):

Makefile
```diff
@@ -50,7 +50,7 @@ endef
 # Targets that are exposed to developers and users.
 build: .format fusedev virtiofs
 release: .format .release_version fusedev virtiofs
-static-release: .musl_target .format .release_version fusedev virtiofs
+static-release: .musl_target .format .release_version fusedev
```
bergwolf, line 53: `static-release` should include both `fusedev` and `virtiofs` targets. To only build fusedev，please add `fusedev-static-release` target instead.

PR #177, opened 2021-09-27
Multiple chunks based rafs read path and user io amplification

changweige:
Currently, rafs can only handle a bio/chunk down to blob layer. In case a target file range spans multiple chunks, it is clumsy and low efficiency.
Fetching data from backend for size smaller than 128KB takes similar time, so we introduce a feature to amplify user io to fetch more data from backend and store ti into blob cache to boost following rafs read.

From benchmark, with 1000 files of 8K size, read latency had decreased by 50%.


imeoer (approved):
LGTM

PR #183, opened 2021-10-12
snapshotter: introduce some config options to control snapshotter behavior

eryugey:
- Add ability to disable indexed chunk map to use digested chunk map
- Introduce log_dir config to specify nydus log dir path
- Introduce log_to_stdout config to dump nydusd log to stdout of snapshotter


changweige (approved):
Looks good to me

PR #184, opened 2021-10-13
feat(nydus-overlayfs): add nydus-overlayfs and adapt for kata

luodw:
As adapted for kata, Mounts interface has changed the return value and
it can not be mount in containerd. So here provide a binary
nydus-overlayfs for containerd to temp mount. refer to containerd https://github.com/containerd/containerd/blob/main/mount/mount_linux.go#L47


contrib/nydus-snapshotter/cmd/containerd-nydus-grpc/pkg/command/flags.go
```diff
@@ -46,6 +46,7 @@ type Args struct {
 	MetricsFile          string
 	EnableStargz         bool
 	DisableCacheManager  bool
+	EnableKata           bool
```
bergwolf, line 49: Do we still need `EnableKata`? for `runc` and `kata`, snapshotter should be able to just return the same mount slice structure IMO.

bergwolf: But we can just always return the nydus options. The fuse mount wrapper would ignore it anyway.

bergwolf: Or can ignore it. I see that it is not ignored right now.

luodw: > But we can just always return the nydus options. The fuse mount wrapper would ignore it anyway.

It sounds good, but if so, in runc, it will also depend on nydus-overlayfs and will increase the cost of use, maybe it is a balance

contrib/nydus-snapshotter/pkg/daemon/daemon.go
```diff
@@ -123,19 +124,46 @@ func NewDaemon(opt ...NewDaemonOpt) (*Daemon, error) {
...
+func copyBootstrapFile(dst, src string) error {
+	d, err := os.OpenFile(dst, os.O_CREATE|os.O_RDWR|os.O_TRUNC, 0644)
+	if err != nil {
+		return err
+	}
+	s, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	_, err = io.Copy(d, s)
```
changweige, line 136: Should close the two opened file ？

contrib/nydus-snapshotter/cmd/containerd-nydus-grpc/pkg/command/flags.go
```diff
@@ -207,6 +214,7 @@ func Validate(args *Args, cfg *config.Config) error {
 	cfg.MetricsFile = args.MetricsFile
 	cfg.EnableStargz = args.EnableStargz
 	cfg.DisableCacheManager = args.DisableCacheManager
+	cfg.EnableKata = args.EnableKata
```
changweige, line 217: Seems not a generic option name as `EnableKata`.
Let's try to think of a more generic one?

contrib/nydus-snapshotter/pkg/daemon/daemon.go
luodw: The code has updated

contrib/nydus-overlayfs/cmd/main.go
```diff
@@ -0,0 +1,144 @@
...
+		target: args[1],
+	}
+	if margs.fsType != "overlay" {
+		return nil, errors.New("fsType only support overlay")
+	}
+	if len(margs.target) == 0 {
+		return nil, errors.New("target can not be empty")
+	}
+	if args[2] == "-o" && len(args[3]) != 0 {
+		for _, opt := range strings.Split(args[3], ",") {
```
changweige, line 46: Just a nitt.
I am curious if we can use package like `"github.com/urfave/cli/v2"` to extract mount point and mount options out of command line arguments since it is a generic step

contrib/nydus-overlayfs/cmd/main.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	var (
+		flags        int
+		finalOptions []string
+	)
+	for _, o := range options {
+		if f, exist := flagsTable[o]; exist {
+			flags |= f
+		} else {
+			finalOptions = append(finalOptions, o)
```
changweige, line 100: What does `finalOptions` want to hold? Hold filesystem-specific data from user-space to kernel?
Then why it comes from mount options?
Shall we add an option to take filesystem-specific data? So it is more clear

contrib/nydus-snapshotter/snapshot/snapshot.go
```diff
@@ -567,60 +567,63 @@ func overlayMount(options []string) []mount.Mount {
...
-		source, err := o.fs.BootstrapFile(id)
-		if err != nil {
-			return nil, err
-		}
-
-		cfg, err := o.fs.NewDaemonConfig(labels)
-		if err != nil {
-			return nil, errors.Wrapf(err, fmt.Sprintf("remoteMounts: failed to generate nydus config for snapshot %s, label: %v", id, labels))
-		}
+	if s.Kind == snapshots.KindActive {
```
changweige, line 588: We need to consider back-compatibility here since some container engine/runtime still depends on `hasDaemon`
mode. So simply removing the if block code snippet is dangerous.

luodw: agree

contrib/nydus-snapshotter/snapshot/snapshot.go
```diff
@@ -567,60 +567,63 @@ func overlayMount(options []string) []mount.Mount {
 	}
 }
 
+func fuseMount(options []string) []mount.Mount {
+	return []mount.Mount{
+		{
+			Type:    "fuse.nydus-overlayfs",
```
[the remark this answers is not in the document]
luodw: I refer to this https://github.com/containers/fuse-overlayfs and https://github.com/containerd/containerd/blob/main/mount/mount_linux.go#L386, in order to stay the same，so named nydus-overlayfs.

contrib/nydus-snapshotter/snapshot/snapshot.go
changweige: Fine

bergwolf:
@luodw have you considered removing the overlay logic from the fuse mount helper? Then nydus snapshotter can return a mount slice like:

For View() mount slice:
```
0: type(fuse.nydus-overlayfs), source (boot.image), options(ro)
```

For Mount() mount slice:
```
0: type(fuse.nydus-overlayfs), source (boot.image), options(mountpoint=lower_A)

1: type(overlay) source(overlay) options(lowerdir=lower_A,upperdir=upper_A,workdir=worker_A)
```

Then containerd and runc shim can call mount.fuse to mount nydus, and continue to mount overlay in `mount.All()`(ref: https://github.com/containerd/containerd/blob/main/mount/mount.go#L33)

contrib/nydus-snapshotter/snapshot/snapshot.go
luodw: I find it is difficult to back-compatibility, because when hasDaemon is true, the returned value has changed to a new type  which is  unified value in runc and kata env. Even continue to use hasDaemon, the returned value also has to be changed.  @changweige @bergwolf What do you think?

contrib/nydus-overlayfs/cmd/main.go
```diff
@@ -0,0 +1,144 @@
...
+	if err := doMount(margs); err != nil {
+		return errors.Wrap(err, "doMount err")
+	}
+	return nil
+}
+
+func main() {
+	app := &cli.App{
+		Name:      "NydusOverlayfs",
+		Usage:     "Binary for containerd to run temp mount in kata + nydus env",
```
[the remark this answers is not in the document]
luodw: has removed

contrib/nydus-snapshotter/snapshot/snapshot.go
changweige: I will consider about this, perhaps tomorrow.
I am quite occupied today. :-(

docs/nydus-overlayfs.md
```diff
@@ -0,0 +1,32 @@
...
+]
+```
+
+Containerd will use `mount.fuse` or `mount.fuse3` when `Type` has prefix `fuse` or `fuse3`, and the `mount.fuse` command format is
+
+```shell
+sudo mount.fuse overlay ./foo/merged -o lowerdir=./foo/lower2:./foo/lower1,upperdir=./foo/upper,workdir=./foo/work -t nydus-overlayfs
+```
+
+when `mount.fuse` start, it will run below command to do the real mount
```
changweige, line 23: typo: start -> starts

PR #212, opened 2021-11-18
snapshotter: don't touch original nydusd auth if no auth in labels

changweige:
Nydusd must pull data from registry with auth if the repo is private.
Snapshotter only fetches auth from labels and if no auth in the labels,
it will use empty string to replace the original auth.

This causes nydusd lossing auth to access registry.

### DIFF
--- a/contrib/nydus-snapshotter/pkg/auth/keychain.go
+++ b/contrib/nydus-snapshotter/pkg/auth/keychain.go
@@ -25,7 +25,7 @@ var (
 	emptyPassKeyChain = PassKeyChain{}
 )
 
-// PassKeyChain is user/pass based key chain
+// PassKeyChain is user/password based key chain
 type PassKeyChain struct {
 	Username string
 	Password string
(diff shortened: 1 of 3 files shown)


contrib/nydus-snapshotter/pkg/auth/keychain.go
```diff
@@ -61,10 +61,20 @@ func (kc PassKeyChain) TokenBase() bool {
 
 // FromLabels find image pull username and secret from snapshot labels
 // if username and secret is empty, we treat it as empty string
-func FromLabels(labels map[string]string) PassKeyChain {
-	return PassKeyChain{
-		Username: labels[label.ImagePullUsername],
-		Password: labels[label.ImagePullSecret],
+func FromLabels(labels map[string]string) *PassKeyChain {
+	u, found := labels[label.ImagePullUsername]
+	if !found {
```
bergwolf, line 66: should check for empty username/password as well.

changweige: Sure. It now also checks if the values are void string.

bergwolf (approved):

PR #217, opened 2021-11-23
release: include an example nydusd config in the release tarball

bergwolf:
So that user can just use it in normal use cases.

### DIFF
--- a/misc/configs/nydusd-config.json
+++ b/misc/configs/nydusd-config.json
@@ -1,0 +1,27 @@
+{
+  "device": {
+    "backend": {
+      "type": "registry",
+      "config": {
+        "scheme": "http",
+        "timeout": 5,
+        "connect_timeout": 5,
+        "retry_limit": 0
+      }
+    },
+    "cache": {
+      "type": "blobcache",
+      "config": {
+        "work_dir": "/cache"
+      }
+    }
+  },
+  "mode": "direct",
+  "digest_validate": false,
+  "iostats_files": true,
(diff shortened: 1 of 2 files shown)


jiangliu (approved):

misc/configs/nydusd-config.json
```diff
@@ -0,0 +1,27 @@
...
+    "cache": {
+      "type": "blobcache",
+      "config": {
+        "work_dir": "/cache"
+      }
+    }
+  },
+  "mode": "direct",
+  "digest_validate": false,
+  "iostats_files": true,
```
changweige, line 21: We'd better not make `iostats_files` as  true in an example configuration since it introduces extra overhead.
How about just removing this item from nydusd-config.json?

bergwolf: How large is the extra overhead?

bergwolf: I thought it is pretty small in terms of memory and latency, isn't it?

changweige: It's related to how many files are packed in. Generally, one file costs extra 40 Bytes for stats counter plus the entry's name length.
As all counters are managed by a `HashMap`, locking acquisition  might introduce latency. But it might be tolerant.

bergwolf: Users can feel free to disable it. With it being open, it is much easier to get io status from nydusd. Useful to new users.

PR #232, opened 2021-12-01
improve passthroughfs when running together with rafs

bergwolf:
When running passthrougfs together with rafs in the same vfs instance, we need to make sure that necessary passthroughfs options are enabled rather than using the default vfs options.


imeoer (approved):
LGTM

PR #234, opened 2021-12-03
Introduce nydus-v2.0/rafsv6

jiangliu:
This patchset is a little big(huge), which does:
1) refactor the storage/blob subsystem
2) introduce Rafsv6, which is compatible with EROFS
3) introduce new tools nydus-stat, nydus-cached


liubogithub (approved):

imeoer (approved):
LGTM!

PR #236, opened 2021-12-08
app: fix incorrect crate name in README.md

liubin:
The crate name should be nydus-app.


changweige (approved):
Looks good. Thanks :-)

PR #243, opened 2021-12-20
nydus-snapshotter: fix recreate daemon config/socket dir

luodw:
when create multiple containers with same image, nydus-snapshotter will
create daemon config/socket dir per container, most of them can not be
deleted even though the image is deleted.

### DIFF
--- a/contrib/nydus-snapshotter/pkg/filesystem/nydus/fs.go
+++ b/contrib/nydus-snapshotter/pkg/filesystem/nydus/fs.go
@@ -344,6 +344,10 @@ func (fs *filesystem) createNewDaemon(snapshotID string, imageID string) (*daemo
 		d   *daemon.Daemon
 		err error
 	)
+	d, _ = fs.manager.GetBySnapshotID(snapshotID)
+	if d != nil {
+		return nil, errdefs.ErrAlreadyExists
+	}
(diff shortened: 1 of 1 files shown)


bergwolf (approved):
The change makes sense. But I'm wondering why containerd would ever delete an image when it is still referenced by some snapshots.

bergwolf:
Discussed with @luodw offline, so the issue is that when we create more containers with the same image in non-shared daemon mode, we would create extra directories in `WithSocketDir` and `WithConfigDir` etc.,  which is a side effect of failing `fs.manager.NewDaemon(d)`. Defly a bug and need to be fixed.

bergwolf:
Thanks @luodw !

contrib/nydus-snapshotter/pkg/filesystem/nydus/fs.go
```diff
@@ -373,7 +377,10 @@ func (fs *filesystem) createSharedDaemon(snapshotID string, imageID string) (*da
 		d            *daemon.Daemon
 		err          error
 	)
-
+	d, _ = fs.manager.GetBySnapshotID(snapshotID)
```
changweige, line 380: Though, seems it is also possible that a race window exists where two goroutines are creating daemons with those directories created.

PR #249, opened 2021-12-30
cargo: update fuse-backend-rs dependency

bergwolf:
To update virtio-queue to the released 0.1.0 version.


eryugey:
The second patch updates virtio-queue but summary mentions fuse-backend-rs

bergwolf:
The two commits both update fuse-backend-rs. The first one fetches virtio-queue changes, and the second one fetches FileVolatileSlice. I kept them separate just in case that `FileVolatileSlice` breaks anything.

imeoer (approved):
LGTM

PR #250, opened 2022-01-04
A series of changes for supporting diff build cache

imeoer:
In buildkit/harbor-acceld scenario, nydus-image need to dump blobs and bootstraps for every layer, the last bootstrap file of layer will be used as final metadata of a Nydus image, and other bootstraps of lower layer will be used as cache files in buildkit to speed up image build on next.

Commits:

Builder: fix blob race in diff build
Nydusify: relax blob list check in checker
Builder: fix invalid blob reference in diff build
Builder: some fixups for diff build cache
Builder: implement diff build cache
Builder: introduce artifact storage


changweige (approved):
Looks good

PR #251, opened 2022-01-05
snapshotter: pass auth from docker config

bergwolf:
So that if there is a docker config.json setup, we can reuse it.


imeoer:
LGTM!

imeoer (approved):

PR #253, opened 2022-01-05
fix ctr remote build and add CI

bergwolf:
We missed testing contrib code in CI and ended up being broken by the dependabot LOL.

### DIFF
--- a/contrib/nydus-snapshotter/Makefile
+++ b/contrib/nydus-snapshotter/Makefile
@@ -10,7 +10,7 @@ PROXY := GOPROXY="${GOPROXY}"
 endif
 
 .PHONY: build
-build: check
+build:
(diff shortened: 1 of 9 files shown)


imeoer (approved):
LGTM!

contrib/nydus-snapshotter/Makefile
```diff
@@ -21,19 +21,14 @@ clear:
...
+test: build
 	go vet $(PACKAGES)
-
-check: vet
 	golangci-lint run
-
-.PHONY: test
-test: vet
-	go test -v -mod=mod -cover ${PACKAGES}
+	go test -v -cover ${PACKAGES}
```
changweige, line 28: With this patch, it's no longer easy for us to just perform static check

bergwolf: `make test` is your friend ;)

PR #254, opened 2022-01-06
workflow: add workflow to convert top docker hub images

bergwolf:
So that we can use them in CI and ref to users to test with.


imeoer (approved):
LGTM!

PR #257, opened 2022-01-06
readme: add brief introduction for nydus-overlayfs

changweige:
An introduction to `nydus-overlyfs` helper was lacked.

In addition, let's move harbor acceleration service introduction to a separated section. Keep the table only to tools belonging to this project repo.


imeoer (approved):
LGTM

PR #258, opened 2022-01-06
nydusify: utilize golangci-lint to perform static check

changweige:
Add golangci-lint rules definition into

This also partly addresses #255

### DIFF
--- a/contrib/nydusify/pkg/parser/parser.go
+++ b/contrib/nydusify/pkg/parser/parser.go
@@ -164,8 +164,8 @@ func (parser *Parser) PullNydusBootstrap(ctx context.Context, image *Image) (io.
 	return reader, nil
 }
 
-func (p *Parser) matchImagePlatform(desc *ocispec.Descriptor) bool {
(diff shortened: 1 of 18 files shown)


contrib/nydusify/pkg/parser/parser.go
```diff
@@ -164,8 +164,8 @@ func (parser *Parser) PullNydusBootstrap(ctx context.Context, image *Image) (io.
 	return reader, nil
 }
 
-func (p *Parser) matchImagePlatform(desc *ocispec.Descriptor) bool {
-	if p.interestedArch == desc.Platform.Architecture && desc.Platform.OS == "linux" {
+func (parser *Parser) matchImagePlatform(desc *ocispec.Descriptor) bool {
+	if parser.interestedArch == desc.Platform.Architecture && desc.Platform.OS == "linux" {
```
imeoer, line 168: I like this suggestion from golangci-lint, we are never using single char as a variable name.

imeoer (approved):

PR #259, opened 2022-01-06
nydusd: remove mountpoint check in mount func

luodw:
There is no need to check whether the mouintpoint is existed and
in some cases like faas, it need to mount different bootstrap to
same mouintpoint at different time

### DIFF
--- a/src/bin/nydusd/daemon.rs
+++ b/src/bin/nydusd/daemon.rs
@@ -293,9 +293,6 @@ pub trait NydusDaemon: DaemonStateMachineSubscriber {
     // NOTE: This method is not thread-safe, however, it is acceptable as
     // mount/umount/remount/restore_mount is invoked from single thread in FSM
     fn mount(&self, cmd: FsBackendMountCmd) -> DaemonResult<()> {
-        if self.backend_from_mountpoint(&cmd.mountpoint)?.is_some() {
-            return Err(DaemonError::AlreadyExists);
-        }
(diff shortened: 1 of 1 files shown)


src/bin/nydusd/daemon.rs
```diff
@@ -293,9 +293,6 @@ pub trait NydusDaemon: DaemonStateMachineSubscriber {
     // NOTE: This method is not thread-safe, however, it is acceptable as
     // mount/umount/remount/restore_mount is invoked from single thread in FSM
     fn mount(&self, cmd: FsBackendMountCmd) -> DaemonResult<()> {
-        if self.backend_from_mountpoint(&cmd.mountpoint)?.is_some() {
```
imeoer, line 296: Why did we check the existence before? cc @changweige

changweige: We want nydus to enforce a certain rule to handle APIs from clients.
It's not easy to justified whether extra MOUNT API request is misbehaved  or not from clients.
Moreover, I didn't think `fuse-backend-rs` crate had a clear semantics for this behavior then.

@luodw Shall we try to umount and than mount again?

imeoer: Agree, the mount is a POST API, its semantic should also not guarantee idempotency.

luodw: ok，umount and than mount again also meet requirement.

PR #270, opened 2022-01-11
Migrate snapshotter to containerd org

changweige:
Containerd community has accepted `nydus-snashotter` as a non-core sub-project. So we are going to remove the code out.

Addresses #269


Makefile
```diff
@@ -185,19 +175,18 @@ docker-nydus-graphdriver-static:
...
+				nydus-overlayfs-test docker-nydus-graphdriver-test
 
 # https://www.gnu.org/software/make/manual/html_node/One-Shell.html
 .ONESHELL:
 docker-example: all-static-release
 	cp ${current_dir}/target-fusedev/${ARCH}-unknown-linux-musl/release/nydusd misc/example
 	cp ${current_dir}/target-fusedev/${ARCH}-unknown-linux-musl/release/nydus-image misc/example
 	cp contrib/nydusify/cmd/nydusify misc/example
-	cp contrib/nydus-snapshotter/bin/containerd-nydus-grpc misc/example
 	docker build -t nydus-rs-example misc/example
```
bergwolf, line 190: docker-example is broken without `nydus-snapshotter`.

changweige: True. How about releasing nydus-snapshotter in https://github.com/containerd/nydus-snapshotter and provide built binaries in its release page and download nydus-snapshotter when `make docker-example`

bergwolf: Better if we can provide a docker image in https://github.com/containerd/nydus-snapshotter so that it is easier for users to deploy the remote snapshotter too.

changweige:
Let's wait for the first release of nydus-snapshotter first for a while

Makefile
changweige: Could we mark this comment as solved since a docker image already is added into https://github.com/containerd/nydus-snapshotter

imeoer:
@changweige It's still working on?

Makefile
bergwolf: We still need to figure out how to run a demo example here. But I agree it can be fixed later.

changweige: Big agree

changweige:
> @changweige It's still working on?

I already remove *wip* from the title since I think it is not work in progress now. Let's review it.

imeoer (approved):

bergwolf (approved):

PR #272, opened 2022-01-11
Builder: fix no bootstrap output in v6 diff build

imeoer:
Issue: if we enable v6 build by specifying these option group:

`--rafs-version=6`, `--source-type=diff` and `--diff-bootstrap-dir`

There is no have bootstrap output in bootstrap directory, this
causes that failed to build v6 image in harbor-acceld or buildkit.

Fix: call release() method after all bootstrap write operations.


changweige (approved):
LGTM!

PR #287, opened 2022-01-27
nydusd: fix thread-num validation

liubin:
The acceptable range for thread-num should be [1-2024], but now
this option can be set larger than 1024.


imeoer (approved):

changweige (approved):
Though there is a typo within changelog, LGTM.

PR #308, opened 2022-02-23
cache: also commit current region for internal IO

changweige:
Skip committing to region for internal IO is causing
discontinuous region commision since an intermediate chunk
is skipped from merged requet with many continuous chunks.

### DIFF
--- a/storage/src/cache/filecache/cache_entry.rs
+++ b/storage/src/cache/filecache/cache_entry.rs
@@ -504,7 +504,7 @@ impl FileCacheEntry {
     // There are some assumption applied to the `bios` passed to `read_iter()`.
     // - The blob address of chunks in `bios` are continuous.
     // - There is at most one user io request in the `bios`.
-    // - The user io request may not aligned on chunk boundary.
+    // - The user io request may not be aligned on chunk boundary.
     // - The user io request may partially consume data from the first and last chunk of user io
     //   request.
     // - Optionally there may be some prefetch/read amplify requests following the user io request.
@@ -546,7 +546,9 @@ impl FileCacheEntry {
             // - the data in the file cache is uncompressed.
             // - data validation is disabled
             if is_ready && !self.is_compressed && !self.need_validate {
-                // Silently drop prefetch requests.
+                // Internal IO should not be committed to local cache region, just
+                // commit this region without pushing any chunk to avoid discontinuous
+                // chunks in a region.
                 if req.tags[i].is_user_io() {
                     state.push(
                         RegionType::CacheFast,
@@ -555,6 +557,8 @@ impl FileCacheEntry {
                         req.tags[i].clone(),
                         None,
                     )?;
+                } else {
+                    state.commit()
                 }
             } else if self.is_stargz || !self.is_direct_chunkmap || is_ready {
                 // Case to try loading data from cache
@@ -581,7 +585,7 @@ impl FileCacheEntry {
                 } else {
                     BlobIoTag::Internal(chunk.compress_offset())
                 };
-                // NOTE: Only this request region can steak more chunks from backend with user io.
+                // NOTE: Only this request region can steal more chunks from backend with user io.
                 state.push(
                     RegionType::Backend,
                     chunk.compress_offset(),
@@ -961,7 +965,7 @@ impl RegionType {
     }
 }
 
-/// A continuous region in cache file or backend storage/blob, it may contains several chunks.
+/// A continuous region in cache file or backend storage/blob, it may contain several chunks.
 struct Region {
     r#type: RegionType,
     status: RegionStatus,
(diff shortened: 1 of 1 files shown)


storage/src/cache/filecache/cache_entry.rs
```diff
@@ -555,6 +557,8 @@ impl FileCacheEntry {
                         req.tags[i].clone(),
                         None,
                     )?;
+                } else {
+                    state.commit()
```
bergwolf, line 561: What does `commit` mean? The only place a merge state is committed, is when there are non-user IO in the fast cache case. The committed status is for the entire merge state, but it only matters in the above-dropped non-user IO case.

storage/src/cache/filecache/cache_entry.rs
```diff
@@ -1058,8 +1070,12 @@ impl FileIoMergeState {
         tag: BlobIoTag,
         chunk: Option<BlobIoChunk>,
     ) -> Result<()> {
-        if self.regions.is_empty() || !self.joinable(region_type) {
+        if self.regions.is_empty()
+            || !self.joinable(region_type)
+            || self.status == FileIoMergeStateStatus::Committed
```
bergwolf, line 1075: So it plays with the merge state status on a per-region manner but the status itself is per state. So sometimes a merge state is committed, and sometimes it is open, depending on the sequence when the user-IO is seen in the list?

storage/src/cache/filecache/cache_entry.rs
```diff
@@ -546,7 +546,9 @@ impl FileCacheEntry {
             // - the data in the file cache is uncompressed.
             // - data validation is disabled
             if is_ready && !self.is_compressed && !self.need_validate {
-                // Silently drop prefetch requests.
+                // Internal IO should not be committed to local cache region, just
```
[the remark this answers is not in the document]
changweige: Yes. At present, internal IO class only comprises *prefetch io*. It defines internal IO as what does not have any pre-allocated memory buffer.

storage/src/cache/filecache/cache_entry.rs
changweige: I should add comment to describe the newly introduced method `commit` to close the understanding gap.
The *MergeState* status is only for last region within the merge state since it is hard for `MergeState` alone to tell if a chunk is skipped to be pushed into, although the chunks in merge requests(req) must be continuous. So we **Must** give the merge state extra hint.

Because `push` method of `MergeState` checks if input chunk is continuous with the last one in regions.

storage/src/cache/filecache/cache_entry.rs
changweige: Yes. It is decided by caller where chunk's type and if it hits local cache matter.
For user-IO, it can be pushed to current region(the last on within merge state), which will be dispatched to read data from local cache file. But for internal IO that also hits local cache, we must skip read data from cache for internal IO since it will consume data buffers provided from Rafs.

storage/src/cache/filecache/cache_entry.rs
```diff
@@ -1039,14 +1043,22 @@ impl Region {
     }
 }
 
+#[derive(PartialEq)]
+enum FileIoMergeStateStatus {
+    Open,
+    Committed,
```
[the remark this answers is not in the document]
changweige: Sure.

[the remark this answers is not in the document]
bergwolf: Ok, now I see what the PR does :)

storage/src/cache/filecache/cache_entry.rs
```diff
@@ -1039,14 +1043,25 @@ impl Region {
...
+    // Last region can take in more io chunks
+    Open,
+    // Last region can no longer take in more io chunks, a new region has to be
+    // created for following chunks.
+    Committed,
+}
+
 struct FileIoMergeState {
     regions: Vec<Region>,
+    status: FileIoMergeStateStatus,
```
bergwolf, line 1057: Can you change `status` to `last_region_joinable: bool` as a marker to tell if the last region in a merge state can join more requests? Also, a few comments to help explain why we need it is even better. Thanks!

storage/src/cache/filecache/cache_entry.rs
```diff
@@ -1058,8 +1073,12 @@ impl FileIoMergeState {
         tag: BlobIoTag,
         chunk: Option<BlobIoChunk>,
     ) -> Result<()> {
-        if self.regions.is_empty() || !self.joinable(region_type) {
+        if self.regions.is_empty()
+            || !self.joinable(region_type)
```
bergwolf, line 1077: I'd prefer folding the `last_region_joinable` check to `joinable()`.

bergwolf (commented):
Overall lgtm. Just a few minor comments.

storage/src/cache/filecache/cache_entry.rs
changweige: Sure. It also makes sense.

bergwolf (approved):
Thanks!

jiangliu (approved):

PR #310, opened 2022-02-24
test: use is_ok/is_err replace match for Result return type

liubin:
In test cases, we can use is_ok or is_err against to Result type,
instead of using a match.


imeoer (approved):
LGTM

bergwolf (approved):

PR #312, opened 2022-02-25
rafs: only append more chunks to last desc when amplify io

changweige:
Amplified IO chunks should be appended to the last Desc rather
than creating a new desc comprising several chunks.
Otherwise it will consume IO buffers provided by upper layer.
Moreover, two descs can't be merged before issuing to backend storage.


jiangliu (approved):

changweige:
This patch still depends other patches fixing several bugs within cache.
I am going to close it for now.

PR #314, opened 2022-02-28
cache: return Ok(false) when io inflight tracer timeouts

changweige:
It is hard to say the waiting IO can't succeed after a previous
io timeouts.
So give the second io a chance to perform its own backend IO.

### DIFF
--- a/storage/src/cache/state/mod.rs
+++ b/storage/src/cache/state/mod.rs
@@ -36,6 +36,7 @@ use std::any::Any;
 use std::io::Result;
 
 use crate::device::BlobChunkInfo;
+use crate::StorageResult;
 
 pub use blob_state_map::BlobStateMap;
 pub use digested_chunk_map::DigestedChunkMap;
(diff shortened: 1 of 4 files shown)


bergwolf (approved):

storage/src/cache/state/mod.rs
```diff
@@ -58,7 +58,7 @@ pub trait ChunkMap: Any + Send + Sync {
     /// Check whether the chunk is ready for use, and mark it as pending if not ready yet.
     ///
     /// The function returns:
-    /// - Err if timeout
+    /// - Ok(false) Waiting for inflight backend IO timeouts.
```
jiangliu, line 61: Does the caller need to distinguish the case and 
`Ok(false) and also marks the chunk as pending, either set_ready_and_clear_pending() or clear_pending() must be called to clear the pending state`?

changweige: Yes. Let's make it retry if waiting for inflight IO timeouts.

jiangliu (approved):

PR #319, opened 2022-03-03
Add capability of prefetch for rafs v6 format

changweige:
Rafs v6 now has the capability of prefetch with this patch set merged if it would.
By design, it places a prefetch table very closely to blob devices table on v6 extended block when creating v6 nydus image.
The `nydus-image` command line user interface is exactly the same with v5 style. Each file that needs to be prefetched will have its `nid` persisted into the prefetch table.
When nydusd starts, the prefetching files will be found from the prefetch table and the files' contents will be downloaded in ahead of time.

### DIFF
--- a/rafs/src/fs.rs
+++ b/rafs/src/fs.rs
@@ -464,9 +464,7 @@ impl Rafs {
         let prefetch_all = self.prefetch_all;
 
         let _ = std::thread::spawn(move || {
-            if sb.meta.is_v5() {
-                Self::do_prefetch_v5(reader, prefetch_files, prefetch_all, sb, device);
-            }
+            Self::do_prefetch(reader, prefetch_files, prefetch_all, sb, device);
         });
     }
 
@@ -479,7 +477,7 @@ impl Rafs {
         self.sb.superblock.root_ino()
     }
 
-    fn do_prefetch_v5(
+    fn do_prefetch(
         mut reader: RafsIoReader,
         prefetch_files: Option<Vec<PathBuf>>,
         prefetch_all: bool,
--- a/rafs/src/metadata/direct_v6.rs
+++ b/rafs/src/metadata/direct_v6.rs
@@ -917,7 +917,37 @@ impl RafsInode for OndiskInodeWrapper {
         &self,
         descendants: &mut Vec<Arc<dyn RafsInode>>,
     ) -> Result<usize> {
-        todo!()
+        if !self.is_dir() {
+            return Err(enotdir!());
+        }
+
+        let mut child_dirs: Vec<Arc<dyn RafsInode>> = Vec::new();
+
+        // EROFS packs dot and dotdot, so skip them two.
+        self.walk_children_inodes(2, &mut |inode: Option<Arc<dyn RafsInode>>,
+                                           name: OsString,
+                                           ino,
+                                           offset| {
+            // Safe to unwrap since it must have child inode.
+            if let Some(child_inode) = inode {
+                if child_inode.is_dir() {
+                    trace!("Got dir {:?}", child_inode.name());
+                    child_dirs.push(child_inode);
+                } else if !child_inode.is_empty_size() && child_inode.is_reg() {
+                    descendants.push(child_inode);
+                }
+                Ok(PostWalkAction::Continue)
+            } else {
+                Ok(PostWalkAction::Continue)
+            }
+        })
+        .unwrap();
+
+        for d in child_dirs {
+            d.collect_descendants_inodes(descendants)?;
+        }
+
+        Ok(0)
     }
 
     fn alloc_bio_vecs(&self, offset: u64, size: usize, user_io: bool) -> Result<Vec<BlobIoVec>> {
--- a/rafs/src/metadata/direct_v6.rs
+++ b/rafs/src/metadata/direct_v6.rs
@@ -878,7 +878,8 @@ impl RafsInode for OndiskInodeWrapper {
 
     /// Check whether the inode is a hardlink.
     fn is_hardlink(&self) -> bool {
-        todo!()
+        let inode = self.disk_inode();
+        inode.nlink() > 1 && self.is_reg()
     }
 
     /// Get inode number of the parent directory.
--- a/rafs/src/metadata/layout/v6.rs
+++ b/rafs/src/metadata/layout/v6.rs
@@ -300,8 +300,11 @@ pub struct RafsV6SuperBlockExt {
     s_blob_table_size: u32,
     /// chunk size
     s_chunk_size: u32,
+    s_prefetch_table_offset: u64,
+    s_prefetch_table_size: u32,
+    s_padding: u32,
     /// Reserved
-    s_reserved: [u8; 232],
+    s_reserved: [u8; 216],
 }
 
 impl_bootstrap_converter!(RafsV6SuperBlockExt);
@@ -393,6 +396,18 @@ impl RafsV6SuperBlockExt {
         u64
     );
     impl_pub_getter_setter!(blob_table_size, set_blob_table_size, s_blob_table_size, u32);
+    impl_pub_getter_setter!(
+        prefetch_table_size,
+        set_prefetch_table_size,
+        s_prefetch_table_size,
+        u32
+    );
+    impl_pub_getter_setter!(
+        prefetch_table_offset,
+        set_prefetch_table_offset,
+        s_prefetch_table_offset,
+        u64
+    );
 }
 
 impl RafsStore for RafsV6SuperBlockExt {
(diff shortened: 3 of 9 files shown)


src/bin/nydus-image/core/bootstrap.rs
```diff
@@ -540,9 +540,22 @@ impl Bootstrap {
...
+            (0, 0)
+        };
+
         let orig_meta_addr = bootstrap_ctx.nodes[0].offset;
         let meta_addr = if blob_table_size > 0 {
-            align_offset(blob_table_offset + blob_table_size, EROFS_BLOCK_SIZE as u64)
+            align_offset(
+                blob_table_offset + blob_table_size + prefetch_table_size as u64,
+                EROFS_BLOCK_SIZE as u64,
+            )
```
bergwolf, line 558: What if `blob_table_size` is zero but `prefetch_table_size` is non-zero?

src/bin/nydus-image/core/bootstrap.rs
```diff
@@ -610,7 +623,28 @@ impl Bootstrap {
...
+            bootstrap_writer
+                .seek_offset(prefetch_table_offset as u64)
+                .context("failed seek prefetch table offset")?;
+
+            pt.store(&mut bootstrap_writer).unwrap();
+
+            bootstrap_writer.file.seek(SeekFrom::Start(ext_sb_offset))?;
+            ext_sb
+                .store(&mut bootstrap_writer)
+                .context("failed to revise extended SB")?;
```
bergwolf, line 644: why ext_sb has to be written multiple times?

src/bin/nydus-image/core/bootstrap.rs
```diff
@@ -610,7 +623,28 @@ impl Bootstrap {
...
+        }
+
+        // EROFS does not have inode table, so we lose the chance to decide if this
         // image has xattr. So we have to rewrite extended super block.
         if ctx.has_xattr {
             ext_sb.set_has_xattr();
```
bergwolf, line 650: Same here.

storage/src/cache/filecache/cache_entry.rs
```diff
@@ -221,6 +221,15 @@ impl BlobCache for FileCacheEntry {
         bios: &[BlobIoDesc],
     ) -> StorageResult<usize> {
         let mut bios = bios.to_vec();
+        bios.iter_mut().for_each(|b| {
+            if let Some(ref chunks_meta) = self.meta {
+                // TODO: the first blob backend io triggers chunks array download.
+                if let BlobIoChunk::Address(_blob_index, chunk_index) = b.chunkinfo {
```
bergwolf, line 227: Not related to this PR, but it looks a bit wired to construct `BlobIoChunk::Address` and convert it to `BlobIoChunk::Base` for v6 IO desc. Looks like we can just create `BlobIoChunk::Base` and remove `BlobIoChunk::Address`. @jiangliu wdyt?

rafs/src/fs.rs
```diff
@@ -464,9 +464,7 @@ impl Rafs {
         let prefetch_all = self.prefetch_all;
 
         let _ = std::thread::spawn(move || {
-            if sb.meta.is_v5() {
-                Self::do_prefetch_v5(reader, prefetch_files, prefetch_all, sb, device);
-            }
+            Self::do_prefetch_v5(reader, prefetch_files, prefetch_all, sb, device);
```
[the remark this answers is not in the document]
changweige: Yes. This method works for both v5 and v6. It is misnamed somehow.
Let's rename it to `do_prefetch()` then

rafs/src/metadata/direct_v6.rs
```diff
@@ -916,7 +917,37 @@ impl RafsInode for OndiskInodeWrapper {
...
+        self.walk_children_inodes(2, &mut |inode: Option<Arc<dyn RafsInode>>,
+                                           name: OsString,
+                                           ino,
+                                           offset| {
+            // Safe to unwrap since it must have child inode.
+            if let Some(child_inode) = inode {
+                if child_inode.is_dir() {
+                    trace!("Got dir {:?}", child_inode.name());
+                    child_dirs.push(child_inode);
+                } else if !child_inode.is_empty_size() && child_inode.is_reg() {
```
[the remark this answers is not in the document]
changweige: An empty regular file will also be packed into nydus image, then it have a size of zero.
Moreover, for rafs v5, symlink has size of zero but non-zero size for symlink size.
For rafs v6, symlink size is also represented by `i_size`. So we have to strict the condition here.

rafs/src/metadata/layout/v6.rs
```diff
@@ -412,7 +427,10 @@ impl Default for RafsV6SuperBlockExt {
             s_blob_table_offset: u64::to_le(0),
             s_blob_table_size: u32::to_le(0),
             s_chunk_size: u32::to_le(0),
-            s_reserved: [0u8; 232],
+            s_reserved: [0u8; 216],
+            s_padding: 0,
+            s_prefetch_table_offset: 0,
+            s_prefetch_table_size: 0,
```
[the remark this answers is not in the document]
changweige: As the default value for each field is **zero**. I didn't figure out the necessity to convert endianness then.
But we can keep that consistency with other fields.

rafs/src/metadata/layout/v6.rs
```diff
@@ -412,7 +427,10 @@ impl Default for RafsV6SuperBlockExt {
             s_blob_table_offset: u64::to_le(0),
             s_blob_table_size: u32::to_le(0),
             s_chunk_size: u32::to_le(0),
-            s_reserved: [0u8; 232],
+            s_reserved: [0u8; 216],
```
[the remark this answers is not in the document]
changweige: Sure:)

src/bin/nydus-image/core/bootstrap.rs
changweige: It' impossible. If `blob_table_size` is zero, current image has no associated blob generated. Then prefetch table has nothing to prefetch. At least, the prefetch source comes from a blob.

src/bin/nydus-image/core/bootstrap.rs
changweige: Because, at the very beginning, I didn't want to touch current code too much.
So it updates extended block when it has to since some fields/info of it change. If the first `store` didn't happen, inodes region can't relocate to its offset from bootstrap. As the re-writing won't bring too much latency or overhand, it is somewhat like a `seek`.

rafs/src/fs.rs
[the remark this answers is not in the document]
changweige: fixed

rafs/src/metadata/layout/v6.rs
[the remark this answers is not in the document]
changweige: fixed

rafs/src/metadata/layout/v6.rs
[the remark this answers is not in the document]
changweige: fixed

src/bin/nydus-image/core/bootstrap.rs
bergwolf: Looks like the ext_sb can be seeked to via `ext_sb_offset` and be written once we finishes setting `ext_sb`. Does it not work somehow?

> If the first store didn't happen, inodes region can't relocate to its offset from bootstrap. 

Does the inode region depend on ext_sb location? I don't quite get it as ext_sb is fixed sized. We can always seek over it.

rafs/src/metadata/direct_v6.rs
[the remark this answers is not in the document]
changweige: Done

src/bin/nydus-image/core/bootstrap.rs
changweige: Done.

bergwolf (approved):
Thanks a lot!

rafs/src/metadata/layout/v6.rs
```diff
@@ -1640,6 +1658,75 @@ impl RafsXAttrs {
     }
 }
 
+#[derive(Clone, Default, Debug)]
+pub struct RafsV6PrefetchTable {
+    /// List of inode numbers for prefetch.
+    /// Note: It's not inode index of inodes table being stored here.
+    pub inodes: Vec<u64>,
```
hsiangkao, line 1665: Hi, we could use 32bit nid here, since blob.meta won't be too large for nydus v6 format.
it can support 4*32 = 128GB bootstrap size, which I think it's enough

changweige: already changed to u32

imeoer (approved):

jiangliu (approved):

PR #327, opened 2022-03-11
fix up release action

bergwolf:


imeoer (approved):

.github/workflows/release.yml
```diff
@@ -95,6 +99,7 @@ jobs:
         name: nydus-artifacts
         path: nydus-static
     - name: upload artifacts
+      if: ${{ github.event_name == 'schedule' }}
```
changweige, line 102: What is this PR trying to address?
Should the catching event is `push`?  Otherwise, the artifacts tarballs are overwritten every day?

bergwolf: oops, you are right!

jiangliu (approved):

PR #328, opened 2022-03-11
Misc: unify go version to 1.17

imeoer:


contrib/ctr-remote/commands/rpull.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/labels"
 	"github.com/containerd/containerd/log"
-	"github.com/dragonflyoss/image-service/contrib/nydus-snapshotter/pkg/label"
+	"github.com/containerd/nydus-snapshotter/pkg/label"
```
changweige, line 30: snapshotter has been relocated for quite a while. How the build succeed with the wrongly imported package url?

bergwolf (approved):

contrib/ctr-remote/commands/rpull.go
imeoer: An old committed package version is defined in `go.sum`.

jiangliu (approved):

PR #335, opened 2022-03-17
bugfix: storage: Fix the race problem when decompressing with lz4

MercyMM:
Lz4 does not support concurrent decompression of the same data into
the same piece of memory. There will be multiple containers mmap the
same file, causing the buffer to be shared between different
processes. This will cause data errors due to race issues when
decompressing with lz4. We solve this problem by creating a temporary
memory to hold the decompressed data.

Because this process will only be executed when the blob.meta file is
created for the first time, which means that a machine will only
execute the process once when the blob.meta is created for the first
time, the memory consumption and performance impact are relatively
small.

Signed-off-by: gexuyang <gexuyang@linux.alibaba.com>

### DIFF
--- a/storage/src/meta/mod.rs
+++ b/storage/src/meta/mod.rs
@@ -558,12 +558,25 @@ impl BlobMetaInfo {
                 return Err(eio!("failed to read blob metadata from backend"));
             }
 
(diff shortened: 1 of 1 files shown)


imeoer (approved):

bergwolf (approved):
The `cargo deny` failure is unrelated.

PR #337, opened 2022-03-17
fix up UT

bergwolf:


src/bin/nydus-image/builder/diff.rs
```diff
@@ -913,6 +914,11 @@ pub mod tests {
 
     #[test]
     fn test_walk_diff() {
+        // Skip mount test for non-root.
+        if getpid().as_raw() != 1 {
```
imeoer, line 918: Check non-root using `getpid`? ^0^

imeoer (approved):

jiangliu (approved):

PR #340, opened 2022-03-18
refact: use mio for daemon exit signal monitoring

ccx1024cc:
Nydusd responds to SYS_TERM and SYS_INT based on event poll. In fact, it needs a way to communicate between threads. Event poll is platform-coherent. 
This PR replace raw eventfd with MIO, which is a I/O library for focusing on non-blocking APIs.

### DIFF
--- a/src/bin/nydusd/main.rs
+++ b/src/bin/nydusd/main.rs
@@ -25,12 +25,13 @@ use std::sync::{
     mpsc::channel,
     Arc, Mutex,
 };
-use std::thread;
+use std::thread::{self, spawn};
 use std::{io, process};
 
 use clap::{App, Arg};
 use event_manager::{EventManager, EventSubscriber, SubscriberOps};
 use fuse_backend_rs::api::{Vfs, VfsOptions};
+use mio::Waker;
 use nix::sys::signal;
 use rlimit::{rlim, Resource};
 use vmm_sys_util::eventfd::EventFd;
@@ -57,7 +58,7 @@ mod upgrade;
 
 lazy_static! {
     static ref EVENT_MANAGER_RUN: AtomicBool = AtomicBool::new(true);
-    static ref EXIT_EVTFD: Mutex::<Option<EventFd>> = Mutex::<Option<EventFd>>::default();
+    static ref EXIT_NOTIFIER: Mutex<Option<Arc<Waker>>> = Mutex::default();
 }
 
 fn get_default_rlimit_nofile() -> Result<rlim> {
@@ -91,12 +92,12 @@ fn get_default_rlimit_nofile() -> Result<rlim> {
 }
 
 pub fn exit_event_manager() {
-    EXIT_EVTFD
+    EXIT_NOTIFIER
         .lock()
         .expect("Not poisoned lock!")
         .as_ref()
         .unwrap()
-        .write(1)
+        .wake()
         .unwrap_or_else(|e| error!("Write event fd failed when exiting event manager, {}", e))
 }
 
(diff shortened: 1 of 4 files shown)


jiangliu:
Linux `EventFd` is missing on macos, and it seems like a common requirements. So should we enhance vmm-sys-utils to provide EventFd for macos?
https://github.com/rust-vmm/vhost/issues/110

src/bin/nydusd/main.rs
```diff
@@ -20,17 +20,19 @@ use std::convert::TryInto;
 use std::fs::File;
 use std::io::{Read, Result};
 use std::ops::DerefMut;
+use std::sync::Mutex;
```
jiangliu, line 23: Line 23 and line 26 seems redundant?

ccx1024cc:
> The eventfd is not required for nydus, but a implementation for communication of threads. So this PR remove the coherent between eventfd and nydusd. The mio maybe a better way to provide abstract API.
PS: It provides different ways to run by os, which is not cared by nydus.

src/bin/nydusd/main.rs
```diff
@@ -491,3 +489,12 @@ fn main() -> Result<()> {
 
     Ok(())
 }
+
+fn start_exit_monitor() {
+    let mut monitor = NydusDaemonSubscriber::new().unwrap();
```
jiangliu, line 494: Should we `unwrap()` here or return error to caller?

src/bin/nydusd/main.rs
ccx1024cc: Mutex is required by lazy_inited global variable,(line 62). PS: Maybe not a good choice, changed later.
Line 26 required by grpc server.

jiangliu: I mean you removed `Mutex` at line 26 and add it back at line 23. So the change is unnecessary:)

src/bin/nydusd/main.rs
```diff
@@ -20,17 +20,19 @@ use std::convert::TryInto;
...
     mpsc::channel,
-    Arc, Mutex,
+    Arc,
 };
-use std::thread;
+use std::thread::{self, spawn};
 use std::{io, process};
 
 use clap::{App, Arg};
 use event_manager::{EventManager, EventSubscriber, SubscriberOps};
```
jiangliu, line 33: Are these still needed after switch to mio?

ccx1024cc: "spawn" is still needed for starting monitor so far(line 496).

bergwolf: The http thread is still using eventfd to exit itself. Are you planning to switch it (via the new micro_http crate) in a different PR?

src/bin/nydusd/main.rs
```diff
@@ -462,7 +459,7 @@ fn main() -> Result<()> {
         info!("api server running at {}", apisock);
     }
 
-    *EXIT_EVTFD.lock().unwrap().deref_mut() = Some(exit_evtfd);
+    start_exit_monitor().unwrap_or_else(|e| error!("failed to start exit monitor: {}", e));
```
bergwolf, line 462: Should fail the main process here if exit monitor fails to start.

src/bin/nydusd/main.rs
```diff
@@ -491,3 +488,14 @@ fn main() -> Result<()> {
 
     Ok(())
 }
+
+fn start_exit_monitor() -> Result<()> {
```
jiangliu, line 492: How about "start_daemon_monitor"?

src/bin/nydusd/main.rs
```diff
@@ -91,12 +92,12 @@ fn get_default_rlimit_nofile() -> Result<rlim> {
 }
 
 pub fn exit_event_manager() {
```
imeoer, line 94: We'd better use as few global variables as possible, how about wrapping an exit manager as a struct implementation, and putting it to FuseDaemon/VirtiofsDaemon? (This is not a blocker for this PR at all)

src/bin/nydusd/main.rs
ccx1024cc: > The http thread is still using eventfd to exit itself. Are you planning to switch it (via the new micro_http crate) in a different PR?

Yes. In a later PR.

src/bin/nydusd/main.rs
ccx1024cc: > We'd better use as few global variables/functions as possible, how about wrapping an exit manager as a struct implementation, and putting it to FuseDaemon/VirtiofsDaemon? (This is not a blocker for this PR at all)

Absolutely right.  Event manager maybe removed, together with global shared memory - EXIT_MANAGER_RUN in later PR.

bergwolf:
LGTM! Please fold the two commits into a single one.

jiangliu (approved):

PR #346, opened 2022-03-24
fix v6 panic due to a trace!

liubogithub:
With this, v6 prefetch with a dynamic prefetch file list is able to work as expected.

### DIFF
--- a/rafs/src/metadata/direct_v6.rs
+++ b/rafs/src/metadata/direct_v6.rs
@@ -344,9 +344,7 @@ pub struct OndiskInodeWrapper {
 
 impl OndiskInodeWrapper {
     fn disk_inode(&self) -> &dyn RafsV6OndiskInode {
-        let i = self.mapping.disk_inode(self.offset);
-        trace!("erofs disk inode loaded nid={} {:?}", i.ino(), i);
-        i
+        self.mapping.disk_inode(self.offset)
     }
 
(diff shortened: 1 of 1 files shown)


jiangliu (approved):

imeoer (approved):

rafs/src/metadata/direct_v6.rs
```diff
@@ -344,9 +344,7 @@ pub struct OndiskInodeWrapper {
 
 impl OndiskInodeWrapper {
     fn disk_inode(&self) -> &dyn RafsV6OndiskInode {
-        let i = self.mapping.disk_inode(self.offset);
-        trace!("erofs disk inode loaded nid={} {:?}", i.ino(), i);
```
changweige, line 348: Rafs v6 can load inode number by `ino()`, why it is being removed?

liubogithub:
```
Rafs: remove a trace! which is too verbose
disk_inode() is called too frequently and floods into logs when
log_level is "trace".
```

I got tons of 
`{"msg":"erofs disk inode loaded nid=1 Ondisk Inode { format: 5, ino: 1, mode: 16877, size: 336, union: 5, nlink: 22, xattr count: 0 }","level":"TRCE","ts":"2022-03-24T13:28:36.315034071+08:00","sid":"liubo-test","shimv2":109284}` in my log.

PR #349, opened 2022-03-24
storage: update BlobPrefetchRequest fields type from u32 to u64

MercyMM:
Blobfs relies on the BlobPrefetchRequest structure that field of "offset"
    and "len" are both u32, the maximum support is 4G. But the maximum blob file
    will exceed 4G, so change the define of "offset" and "len" to u64.

### DIFF
--- a/blobfs/src/sync_io.rs
+++ b/blobfs/src/sync_io.rs
@@ -48,12 +48,6 @@ impl BlobFs {
             blob_id_full_path
         );
 
-        debug_assert!(
-            parent
-                == Path::new(self.cfg.ps_config.root_dir.as_str())
-                    .join(self.bootstrap_args.blob_cache_dir.as_str())
-        );
-
         let blob_file = Self::open_file(
             libc::AT_FDCWD,
             &blob_id_full_path.as_path(),
--- a/blobfs/src/sync_io.rs
+++ b/blobfs/src/sync_io.rs
@@ -24,7 +24,18 @@ use storage::device::BlobPrefetchRequest;
 
 impl BlobFs {
     #[cfg(feature = "virtiofs")]
-    fn get_blob_id_and_size(&self, inode: Inode) -> io::Result<(String, i64)> {
+    fn check_st_size(blob_id: &Path, size: i64) -> io::Result<()> {
+        if size < 0 {
+            return Err(einval!(format!(
+                "load_chunks_on_demand: blob_id {:?}, size: {:?} is less than 0",
+                blob_id, size
+            )));
+        }
+        Ok(())
+    }
+
+    #[cfg(feature = "virtiofs")]
+    fn get_blob_id_and_size(&self, inode: Inode) -> io::Result<(String, u64)> {
         // locate blob file that the inode refers to
         let blob_id_full_path = self.pfs.readlinkat_proc_file(inode)?;
         let parent = blob_id_full_path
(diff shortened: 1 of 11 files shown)


blobfs/src/sync_io.rs
```diff
@@ -18,13 +18,26 @@ use std::ffi::CStr;
...
 use storage::device::BlobPrefetchRequest;
 
 impl BlobFs {
     #[cfg(feature = "virtiofs")]
-    fn get_blob_id_and_size(&self, inode: Inode) -> io::Result<(String, i64)> {
+    fn check_st_size(blob_id: &PathBuf, size: i64) -> io::Result<()> {
+        if size > 0 {
+            return Err(einval!(format!(
+                "load_chunks_on_demand: blob_id {:?}, size: {:?} is less than 0",
+                blob_id, size
```
liubogithub, line 33: Is it supposed to be `if size < 0`?

blobfs/src/sync_io.rs
```diff
@@ -60,31 +67,32 @@ impl BlobFs {
...
-                foffset
-            ))
-        })?;
-        let len = (size - offset as i64).try_into().map_err(|_| {
-            einval!(format!(
-                "blobfs: load_chunks_on_demand: len {} is larger than u32::MAX",
-                (size - offset as i64)
-            ))
-        })?;
+        if size < offset {
```
liubogithub, line 84: what about `if size == offset`?

blobfs/src/sync_io.rs
```diff
@@ -60,31 +67,32 @@ impl BlobFs {
...
+            return Err(einval!(format!(
+                "blobfs: blob {:?} offset {:?} is larger than size {:?}",
```
liubogithub, line 86: Can you please add "load_chunks_on_demand" in the error message to make it easier to locate where goes wrong?

storage/src/cache/worker.rs
```diff
@@ -304,11 +304,8 @@ impl AsyncWorkerMgr {
                     e
                 );
             }
-        } else if offset < u32::MAX as u64 && size < u32::MAX as u64 {
-            let _ = cache.reader().prefetch_blob_data_range(
-                offset as u32,
-                std::cmp::min(size as u32, u32::MAX - offset as u32),
-            );
+        } else {
+            let _ = cache.reader().prefetch_blob_data_range(offset, size);
```
liubogithub, line 308: By design the prefetch table is supposed to have (offset, size) < u32::MAX.
Can we leave a `warn!` or `info!` here about (offset, size) is > u32::MAX?

blobfs/src/sync_io.rs
MercyMM: It is a mistakes.  I will fix it

blobfs/src/sync_io.rs
MercyMM: ok

blobfs/src/sync_io.rs
MercyMM: the BlobPrefetchRequest.len will be 0,  storage mod should handle it correctly, let me check to make sure.

storage/src/cache/worker.rs
MercyMM: After this patch, u64 is no by design. This does not affect default behavior. If it does not cause potential bugs, I feel that warn! is useless.

eryugey:
LGTM!

eryugey:
Besides there're clippy warnings that should be fixed.

imeoer (approved):

liubogithub (approved):

PR #351, opened 2022-03-25
Builder: support tar build

imeoer:
Introduce a new build mode named tar build in nydusify and builder (nydus-image).

In order to correspond to the OCI image layer, we need a new build mode, that is
to convert the OCI-formatted tar stream directly into a nydus-formatted tar stream.

The nydus tar stream contains blob and bootstrap files, and the workflow is divided
into two steps, Convert and Merge.

Convert: converts each layer of the OCI tar stream to a nydus tar stream.
Merge: reads the bootstrap files from each layer of the nydus tar stream and merges
them into a final bootstrap.

Support merge sub-command in builder, it will be used in [converter](https://github.com/containerd/nydus-snapshotter/pull/34) package, and this
package will be imported by acceleration-service and buildkit to support the conversion
or build of nydus image.

### DIFF
--- a/contrib/nydusify/pkg/checker/rule/bootstrap.go
+++ b/contrib/nydusify/pkg/checker/rule/bootstrap.go
@@ -8,7 +8,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"reflect"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -54,6 +53,16 @@ func (rule *BootstrapRule) Validate() error {
 		return nil
 	}
 
+	// Parse blob list from blob layers in Nydus manifest
+	blobListInLayer := map[string]bool{}
+	layers := rule.Parsed.NydusImage.Manifest.Layers
+	for i, layer := range layers {
+		if i != len(layers)-1 {
+			blobListInLayer[layer.Digest.Hex()] = true
+		}
+	}
+
+	// Parse blob list from blob table of bootstrap
 	var bootstrap bootstrapDebug
 	bootstrapBytes, err := ioutil.ReadFile(rule.DebugOutputPath)
 	if err != nil {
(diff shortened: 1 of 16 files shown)


hsiangkao (approved):

src/bin/nydus-image/merge.rs
```diff
@@ -0,0 +1,183 @@
...
+                        "too many layers {}, limited to {}",
+                        layer_idx,
+                        u16::MAX
+                    ))?;
+                    node.overlay = Overlay::UpperAddition;
+                    match node.whiteout_type(WhiteoutSpec::Oci) {
+                        Some(_) => {
+                            // Insert removal operations at the head, so they will be handled first when
+                            // applying to lower layer.
+                            nodes.insert(0, node);
```
bergwolf, line 153: nits: `Insert whiteouts at the head, xxxx`

src/bin/nydus-image/merge.rs
```diff
@@ -0,0 +1,183 @@
...
+                        current_flags,
+                    );
+                }
+            } else {
+                // Keep final bootstrap following superblock flags of source bootstraps.
+                current_flags.set_to_ctx(ctx);
+                flags = Some(current_flags);
+            }
+
+            let parent_blobs = rs.superblock.get_blob_infos();
```
bergwolf, line 107: Why not simply bail if `len(parent_blobs) > 1`? Then the `parent_blob_added` logic can be dropped.

contrib/nydusify/pkg/checker/rule/bootstrap.go
```diff
@@ -47,42 +38,5 @@ func (rule *BootstrapRule) Validate() error {
...
-	layers := rule.Parsed.NydusImage.Manifest.Layers
-	for i, layer := range layers {
-		if i != len(layers)-1 {
-			blobListInLayer = append(blobListInLayer, layer.Digest.Hex())
-		}
-	}
-
-	// Blob list recorded in manifest annotation should be equal with
-	// the blob list recorded in blob table of bootstrap
-	if !reflect.DeepEqual(bootstrap.Blobs, blobListInLayer) {
```
bergwolf, line 77: We should still validate that all the blobs in bootstrap are in the image manifest, to make sure we don't reference any non-existing blobs in bootstrap.

bergwolf (commented):
lgtm!

src/bin/nydus-image/merge.rs
imeoer: Fixed

src/bin/nydus-image/merge.rs
imeoer: It is assumed that the `nydus-image create` at each layer and `nydus-image merge` commands use the same chunk dict bootstrap, so there is at most one blob from the parent bootstrap and all other blobs from the chunk dict bootstrap. Added a comment.

contrib/nydusify/pkg/checker/rule/bootstrap.go
imeoer: Good point! fixed.

src/bin/nydus-image/merge.rs
imeoer: Added more detailed comment.

bergwolf (approved):
Thanks!

PR #355, opened 2022-03-27
RFC: add FsCacheHandler to support fscache

jiangliu:
Introduce FsCacheHandler to cooperate with linux fscache driver to support fscache based blob cache.


jiangliu:
Now we have the `fscache` branch, so close this.

PR #358, opened 2022-03-29
blobfs: ensure blob cache dir exist

MercyMM:
When blob_cache_dir does not exist, create it instead of return error.

Signed-off-by: gexuyang <gexuyang@linux.alibaba.com>

### DIFF
--- a/blobfs/src/lib.rs
+++ b/blobfs/src/lib.rs
@@ -29,6 +29,7 @@ use std::any::Any;
 #[cfg(feature = "virtiofs")]
 use std::ffi::CStr;
 use std::ffi::CString;
+use std::fs::create_dir_all;
 #[cfg(feature = "virtiofs")]
 use std::fs::File;
 use std::io;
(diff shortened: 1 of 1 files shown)


jiangliu:
Please help to add commit message:)

MercyMM:
>

MercyMM:
> 
done

blobfs/src/lib.rs
```diff
@@ -168,6 +169,20 @@ pub struct BlobFs {
...
+        if path.as_os_str().is_empty() {
+            return Err(einval!("path is empty"));
+        }
+        if !path.exists() {
+            create_dir_all(path).map_err(|e| {
+                error!("create dir error: {:?}", path);
+                e
+            })?;
+        }
+
```
imeoer, line 182: If the path exists, it may not be a directory, should we throw an error?

MercyMM: if the path is not a directory, it will throw an error in Blobfs mount function. So we just ensure path exist here.

imeoer (approved):

blobfs/src/lib.rs
```diff
@@ -168,6 +169,20 @@ pub struct BlobFs {
...
 
 impl BlobFs {
+    fn ensure_path_exist(path: &Path) -> io::Result<()> {
+        if path.as_os_str().is_empty() {
+            return Err(einval!("path is empty"));
+        }
+        if !path.exists() {
+            create_dir_all(path).map_err(|e| {
+                error!("create dir error: {:?}", path);
+                e
```
liubogithub, line 179: can we use use macro einval!(e) which shows more verbose message including file!() and line!()

MercyMM: macro einval! is more convenient, but it will lose real error code return from create_dir_all().

liubogithub: I'm fine with it, but we need to make `error!("create dir error: {:?}", path);` more verbose so that we know where goes wrong exactly.

liubogithub (approved):

PR #359, opened 2022-03-30
Nydusify: ignore arch for single manifest

imeoer:
For a single manifest image, we just ignore the arch, so that allowing
to do a default conversion on a different arch's host, for example
converting an arm64 image on an amd64 host.

### DIFF
--- a/contrib/nydusify/pkg/parser/parser.go
+++ b/contrib/nydusify/pkg/parser/parser.go
@@ -118,7 +118,7 @@ func (parser *Parser) pullIndex(ctx context.Context, desc *ocispec.Descriptor) (
 }
 
 func (parser *Parser) parseImage(
-	ctx context.Context, desc *ocispec.Descriptor, onlyManifest *ocispec.Manifest,
+	ctx context.Context, desc *ocispec.Descriptor, onlyManifest *ocispec.Manifest, ignoreArch bool,
 ) (*Image, error) {
 	var manifest *ocispec.Manifest
 	var err error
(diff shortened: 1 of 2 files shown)


ccx1024cc (approved):

contrib/nydusify/pkg/parser/parser.go
```diff
@@ -118,7 +118,7 @@ func (parser *Parser) pullIndex(ctx context.Context, desc *ocispec.Descriptor) (
 }
 
 func (parser *Parser) parseImage(
-	ctx context.Context, desc *ocispec.Descriptor, onlyManifest *ocispec.Manifest,
+	ctx context.Context, desc *ocispec.Descriptor, onlyManifest *ocispec.Manifest, ignoreArch bool,
```
bergwolf, line 121: There is no way for `nydusify` to specify `ignoreArch`?

bergwolf: How is it going to be used? Please add an example in the docs.

imeoer: Updated comment. No way to specify. For a single manifest image, it's an automatic behavior to ignore the image arch, attempt to continue converting, and give a warning message. For the manifest list, nydusify selects host specified arch to convert, but the user still can specify an interesting platform (by --platform), the ultimate goal is that nydusify can convert any arch's image on any arch's host.

contrib/nydusify/pkg/parser/parser.go
```diff
@@ -136,12 +136,22 @@ func (parser *Parser) parseImage(
...
+			logrus.WithError(err).Warn("Ignore image arch")
+		} else {
+			return nil, err
+		}
 	}
 
 	// Just give user a simple hint telling option was ignored.
 	if config.Architecture != parser.interestedArch {
-		return nil, errors.Errorf("Specified %s architecture was not found", parser.interestedArch)
+		err = errors.Errorf("Found arch %s, but the specified target arch (--platform) is %s", config.Architecture, parser.interestedArch)
```
jiangliu, line 149: Should this be move into `else{}` branch?

imeoer: The same error message should be returned if no ignoreArch.

bergwolf (approved):

PR #374, opened 2022-04-06
Enhance storage/filecache to enable read amplification

jiangliu:
Enhance storage/filecache to enable read amplify, this is also a candidate for 2.0

### DIFF
--- a/storage/src/cache/fscache/mod.rs
+++ b/storage/src/cache/fscache/mod.rs
@@ -148,27 +148,26 @@ impl BlobCacheMgr for FsCacheMgr {
     }
 
     fn gc(&self, id: Option<&str>) {
-        let mut reclaim = Vec::new();
-
         if let Some(blob_id) = id {
-            reclaim.push(blob_id.to_string());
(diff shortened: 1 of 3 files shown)


storage/src/cache/fscache/mod.rs
```diff
@@ -148,27 +148,27 @@ impl BlobCacheMgr for FsCacheMgr {
...
-            let mut guard = self.blobs.write().unwrap();
-            if let Some(entry) = guard.get(key) {
-                if Arc::strong_count(entry) > 1 {
-                    continue;
+            drop(guard);
+
+            for key in reclaim.iter() {
+                let mut guard = self.blobs.write().unwrap();
+                if let Some(entry) = guard.get(key) {
+                    if id.is_none() && Arc::strong_count(entry) > 1 {
```
hsiangkao, line 166: id should be always none here?

jiangliu: My mistake, `id.is_none() &&` should be removed

hsiangkao (approved):

PR #376, opened 2022-04-07
Fix testcase break due to read amplification

hsiangkao:
Fix testcase break due to read amplification

### DIFF
--- a/storage/src/meta/mod.rs
+++ b/storage/src/meta/mod.rs
@@ -861,7 +861,7 @@ mod tests {
             state: Arc::new(state),
         };
 
-        let vec = info.get_chunks_uncompressed(0x0, 0x1001).unwrap();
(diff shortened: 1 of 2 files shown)


storage/src/meta/mod.rs
```diff
@@ -861,7 +861,7 @@ mod tests {
             state: Arc::new(state),
         };
 
-        let vec = info.get_chunks_uncompressed(0x0, 0x1001).unwrap();
+        let vec = info.get_chunks_uncompressed(0x0, 0x1001, 1).unwrap();
```
jiangliu, line 864: Use "0" instead of "1", which means batch operation is disabled?

hsiangkao: done.

jiangliu (approved):

PR #377, opened 2022-04-07
storage: support direct IO for fscache

jiangliu:
Enhance cachefiles to support direct IO for fscahce, but direct IO is
not supported for filecache yet. We may enhance it to support filecache
on demand.

Signed-off-by: Liu Jiang <gerry@linux.alibaba.com>


hsiangkao (approved):
Test ok, and looks good!

PR #384, opened 2022-04-13
nydusd: graceful state machine thread shutdown

bergwolf:
When exiting, make sure we wait for the state machine thread to quit.

A few changes were applied:
1. exit_daemon() should call daemon.stop() so that it can result in
    different daemon action depending on the current state. Then in a normal
    workflow, nydusd would just umount and quit. But when in Interrupted,
    nydusd just dies directly without doing any umount etc.
2. Quit the state machine thread upon EXIT (TerminateFuseService) event.
3. Wait for the state machine thread together when waiting for the fuse
    service threads.
4. Nydusd should just die when getting EXIT message in INIT state.
5. Only send Stop event if state is not INTERRUPTED or STOPPED.
    Otherwise there is no state machine thread to handle it.
6. Init and Interrupted should always be able to stop as well.

### DIFF
--- a/api/src/http.rs
+++ b/api/src/http.rs
@@ -429,6 +429,7 @@ pub fn start_http_thread(
                     Err(e) if e.kind() == std::io::ErrorKind::Interrupted => continue,
                     Err(e) => {
                         error!("http server poll events failed, {}", e);
+                        exit_api_server(api_notifier, &to_api);
                         return Err(e);
                     }
                     Ok(_) => {}
--- a/api/src/http.rs
+++ b/api/src/http.rs
@@ -419,10 +419,12 @@ pub fn start_http_thread(
         .spawn(move || {
             // Must start the server successfully or just die by panic
             server.start_server().unwrap();
+            let mut events = Events::with_capacity(100);
+            let mut do_exit = false;
+
             info!("http server started");
 
-            let mut events = Events::with_capacity(100);
-            'wait: loop {
+            loop {
                 match pool.poll(&mut events, None) {
                     Err(e) if e.kind() == std::io::ErrorKind::Interrupted => continue,
                     Err(e) => {
@@ -434,10 +436,7 @@ pub fn start_http_thread(
 
                 for event in &events {
                     match event.token() {
-                        EXIT_TOKEN => {
-                            exit_api_server(api_notifier.clone(), &to_api);
-                            break 'wait;
-                        }
+                        EXIT_TOKEN => do_exit = true,
                         REQUEST_TOKEN => match server.requests() {
                             Ok(request_vec) => {
                                 for server_request in request_vec {
@@ -461,6 +460,11 @@ pub fn start_http_thread(
                         },
                         _ => unreachable!("unknown poll token."),
                     }
+                }
+
+                if do_exit {
+                    exit_api_server(api_notifier, &to_api);
+                    break;
                 }
             }
             info!("http-server thread exits");
--- a/src/bin/nydusd/api_server_glue.rs
+++ b/src/bin/nydusd/api_server_glue.rs
@@ -10,9 +10,6 @@ use std::sync::mpsc::{Receiver, Sender};
 use std::sync::Arc;
 use std::thread::JoinHandle;
 
-use nix::sys::signal::{kill, SIGTERM};
-use nix::unistd::Pid;
-
 use nydus::{FsBackendType, NydusError};
 use nydus_api::http::{
     ApiError, ApiMountCmd, ApiRequest, ApiResponse, ApiResponsePayload, ApiResult, DaemonConf,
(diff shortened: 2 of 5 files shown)


jiangliu:
Is this target 2.0? There will be heavy conflicts with fscache branch.

bergwolf:
@jiangliu Yes, it fixes hot-upgrade and is targeting at 2.0 where hot-upgrade is broken with the introduction of mio.

src/bin/nydusd/api_server_glue.rs
```diff
@@ -216,7 +216,22 @@ impl ApiServer {
...
+                error!("exit fuse service failed {:}", e);
+                ApiError::DaemonAbnormal(e.into())
+            })?;
+
+        // Ensure both fuse and state machine threads have be terminated thus this
+        // nydusd won't race fuse messages when upgrading.
+        d.wait()
+            .map(|_| {
+                info!("fuse service exited by http request");
+                ApiResponsePayload::Empty
```
imeoer, line 229: Is it required?

api/src/http.rs
```diff
@@ -417,12 +417,16 @@ pub fn start_http_thread(
...
+            let mut do_exit = false;
+
             info!("http server started");
 
-            let mut events = Events::with_capacity(100);
-            'wait: loop {
+            loop {
                 match pool.poll(&mut events, None) {
                     Err(e) if e.kind() == std::io::ErrorKind::Interrupted => continue,
                     Err(e) => {
```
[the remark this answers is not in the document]
bergwolf: The PR handles EXIT_TOKEN. If poll fails, looks like we should also call exit_api_server()?

src/bin/nydusd/api_server_glue.rs
bergwolf: Yes, it is. We need to reply properly to API callers.

api/src/http.rs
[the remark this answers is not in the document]
bergwolf: fixed

src/bin/nydusd/api_server_glue.rs
imeoer: The return value of `d.wait` seems not to be used, but that's okay.

imeoer (approved):

src/bin/nydusd/daemon.rs
```diff
@@ -233,9 +233,14 @@ pub trait NydusDaemon: DaemonStateMachineSubscriber {
     fn stop(&self) -> DaemonResult<()> {
         self.on_event(DaemonStateMachineInput::Stop)
     }
+    /// close the current FUSE connection to properly shutdown
+    /// the FUSE server daemon.
     fn disconnect(&self) -> DaemonResult<()>;
-    fn as_any(&self) -> &dyn Any;
+    /// close the FUSE server without closing the FUSE connection
+    /// so that another FUSE server daemon can take over the same
+    /// FUSE connection and continue serving incoming FUSE requests.
```
liubogithub, line 241: is it supposed to be "continue to serve"?

src/bin/nydusd/api_server_glue.rs
```diff
@@ -216,7 +216,22 @@ impl ApiServer {
                 info!("exit daemon by http request");
                 ApiResponsePayload::Empty
             })
-            .map_err(|e| ApiError::DaemonAbnormal(e.into()))?;
+            .map_err(|e| {
+                error!("exit fuse service failed {:}", e);
+                ApiError::DaemonAbnormal(e.into())
+            })?;
+
+        // Ensure both fuse and state machine threads have be terminated thus this
```
liubogithub, line 224: typo, "have been terminated"

src/bin/nydusd/fusedev.rs
```diff
@@ -229,18 +230,22 @@ impl NydusDaemon for FusedevDaemon {
...
-                .join()
-                .map_err(|e| {
-                    DaemonError::WaitDaemon(
-                        *e.downcast::<Error>()
-                            .unwrap_or_else(|e| Box::new(eother!(e))),
-                    )
-                })?
-                .map_err(DaemonError::WaitDaemon)?;
+        loop {
+            let handle = self.threads.lock().unwrap().pop();
```
liubogithub, line 234: This changes where threads.lock protects,  it'll unlock  `threads` after pop(), why is that?

bergwolf:
Mark `wip` as I'm still testing it.

src/bin/nydusd/fusedev.rs
bergwolf: If we block on `handle.join` with the lock at hand, we cannot add more threads to the waiting list, which is necessary for the upgrade use cases, where we just start the state machine thread and then start more fuse service threads later on.

jiangliu:
We need to merge this into fscache branch too once it gets merged:)

bergwolf:
This is ready to be reviewed/merged.

src/bin/nydusd/fusedev.rs
liubogithub: Please correct me if I'm wrong.

In case of upgrade usecase, `d.wait()` is called by old binary's daemon right before old binary quits itself, so why would be more fuse service threads coming in to serve old binary?

With that being said, it looks like early unlock is fine because there is unlikely a race here.

bergwolf: It is not about the old binary's daemon. The new daemon will start and block on waiting for the state machine thread. And then the control path manager will send fd to the new daemon and ask it to take over the service. Then the new daemon will need to start fuse server threads and add them to the waiting list. That's why we need to drop the lock here.

PR #386, opened 2022-04-14
http: fix waker dead improperly due to lifetime mismatch

hsiangkao:
The mio waker must be alive to get the notification delivered,
otherwise the inflight notification may get dropped silently.

Signed-off-by: Liu Jiang \<gerry@linux.alibaba.com\>

### DIFF
--- a/api/src/http.rs
+++ b/api/src/http.rs
@@ -399,8 +399,8 @@ pub fn start_http_thread(
     let socket_path = PathBuf::from(path);
 
     let mut pool = Poll::new()?;
-    let waker = Waker::new(pool.registry(), EXIT_TOKEN)?;
-    let waker = Arc::new(waker);
+    let waker = Arc::new(Waker::new(pool.registry(), EXIT_TOKEN)?);
+    let waker2 = waker.clone();
     let mut server = HttpServer::new(socket_path).map_err(|e| {
         if let ServerError::IOError(e) = e {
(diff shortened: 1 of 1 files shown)


api/src/http.rs
```diff
@@ -399,8 +399,8 @@ pub fn start_http_thread(
     let socket_path = PathBuf::from(path);
 
     let mut pool = Poll::new()?;
-    let waker = Waker::new(pool.registry(), EXIT_TOKEN)?;
-    let waker = Arc::new(waker);
+    let waker = Arc::new(Waker::new(pool.registry(), EXIT_TOKEN)?);
+    let _waker = waker.clone();
```
jiangliu, line 403: Should we rename it to wake2 to keep consistence with #385 ?

hsiangkao: I'm afraid it has to be called "_waker..." at least, since it seems unused.

hsiangkao: updated.

jiangliu: There's a `drop(wake2)`.

hsiangkao: Yeah, it's an explicit use, but not sure if static analyzer changes the semantics in the future.
Anyway, I've renamed it to waker2.

imeoer (approved):

jiangliu (approved):

PR #388, opened 2022-04-14
Add support for zstd compression algorithm for blob file.

pigletDips:
… be tested.

### DIFF
--- a/docs/nydus-design.md
+++ b/docs/nydus-design.md
@@ -169,6 +169,8 @@ bitflags! {
         const HAS_XATTR = 0x0000_0020;
         // Data chunks are compressed with gzip
         const COMPRESS_GZIP = 0x0000_0040;
+        // Data chunks are compressed with zstd
+        const COMPRESS_ZSTD = 0x0000_0080;
     }
 }
 ```
--- a/utils/Cargo.toml
+++ b/utils/Cargo.toml
@@ -18,6 +18,7 @@ lz4-sys = "1.9.2"
 serde = { version = ">=1.0.27", features = ["serde_derive", "rc"] }
 serde_json = ">=1.0.9"
 sha2 = "0.9.1"
+zstd = "0.11"
 
 nydus-error = { version = "0.2", path = "../error" }
 
--- a/utils/src/compress/lz4_standard.rs
+++ b/utils/src/compress/lz4_standard.rs
@@ -15,7 +15,7 @@ pub(super) fn lz4_compress(src: &[u8]) -> Result<Vec<u8>> {
     }
 
     let mut dst_buf = Vec::with_capacity(compress_bound as usize);
-    let dec_size = unsafe {
+    let cmp_size = unsafe {
         LZ4_compress_default(
             src.as_ptr() as *const c_char,
             dst_buf.as_mut_ptr() as *mut c_char,
@@ -23,12 +23,12 @@ pub(super) fn lz4_compress(src: &[u8]) -> Result<Vec<u8>> {
             compress_bound,
         )
     };
-    if dec_size <= 0 {
-        return Err(eio!("decompression failed"));
+    if cmp_size <= 0 {
+        return Err(eio!("compression failed"));
     }
 
-    assert!(dec_size as usize <= dst_buf.capacity());
-    unsafe { dst_buf.set_len(dec_size as usize) };
+    assert!(cmp_size as usize <= dst_buf.capacity());
+    unsafe { dst_buf.set_len(cmp_size as usize) };
 
     Ok(dst_buf)
 }
--- a/utils/src/compress/zstd_standard.rs
+++ b/utils/src/compress/zstd_standard.rs
@@ -1,0 +1,14 @@
+use std::io::Result;
+
+use zstd::{
+    bulk::{compress, decompress_to_buffer},
+    DEFAULT_COMPRESSION_LEVEL,
+};
+
+pub(super) fn zstd_compress(src: &[u8]) -> Result<Vec<u8>> {
+    compress(src, DEFAULT_COMPRESSION_LEVEL)
+}
+
+pub(super) fn zstd_decompress(src: &[u8], dst: &mut [u8]) -> Result<usize> {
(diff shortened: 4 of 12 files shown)


docs/nydus-design.md
```diff
@@ -169,6 +169,8 @@ bitflags! {
         const HAS_XATTR = 0x0000_0020;
         // Data chunks are compressed with gzip
         const COMPRESS_GZIP = 0x0000_0040;
+        // Data chunks are compressed with zstd
+        const COMPRESS_GZIP = 0x0000_0080;
```
liubogithub, line 173: obvious typo.

liubogithub (commented):
The patch needs a better commit log other than "... be tested".

utils/src/compress/lz4_standard.rs
```diff
@@ -15,20 +15,20 @@ pub(super) fn lz4_compress(src: &[u8]) -> Result<Vec<u8>> {
     }
 
     let mut dst_buf = Vec::with_capacity(compress_bound as usize);
-    let dec_size = unsafe {
+    let cmp_size = unsafe {
```
liubogithub, line 18: These changes are irrelevant to the zstd support patch, can you please move it into a separate one?

utils/src/compress/lz4_standard.rs
```diff
@@ -15,20 +15,20 @@ pub(super) fn lz4_compress(src: &[u8]) -> Result<Vec<u8>> {
...
             src.as_ptr() as *const c_char,
             dst_buf.as_mut_ptr() as *mut c_char,
             src.len() as i32,
             compress_bound,
         )
     };
-    if dec_size <= 0 {
-        return Err(eio!("decompression failed"));
+    if cmp_size <= 0 {
+        return Err(eio!("compression failed"));
```
liubogithub, line 27: Ditto.

utils/src/compress/lz4_standard.rs
```diff
@@ -15,20 +15,20 @@ pub(super) fn lz4_compress(src: &[u8]) -> Result<Vec<u8>> {
...
     }
 
-    assert!(dec_size as usize <= dst_buf.capacity());
-    unsafe { dst_buf.set_len(dec_size as usize) };
+    assert!(cmp_size as usize <= dst_buf.capacity());
+    unsafe { dst_buf.set_len(cmp_size as usize) };
```
liubogithub, line 31: Ditto.

liubogithub (changes requested):

utils/Cargo.toml
```diff
@@ -18,6 +18,7 @@ lz4-sys = "1.9.2"
 serde = { version = ">=1.0.27", features = ["serde_derive", "rc"] }
 serde_json = ">=1.0.9"
 sha2 = "0.9.1"
+zstd-sys = "2.0.1"
```
imeoer, line 21: Thanks for the first contribution! I wonder why not use `zstd` crate directly?

docs/nydus-design.md
pigletDips: Ok

utils/src/compress/lz4_standard.rs
pigletDips: Ok

utils/src/compress/lz4_standard.rs
pigletDips: Ok

utils/Cargo.toml
pigletDips: The main reason is that the lz4 using the `lz4-sys` crate, so I use the `zstd-sys` here, and feels like it would be more flexible. Should I change to `zstd`? Thanks for your review!

utils/src/compress/lz4_standard.rs
pigletDips: Thanks for your review!

utils/Cargo.toml
imeoer: `zstd` -> `zstd-safe` -> `zstd-sys`

Seems no flexible requirement here, using a safe crate directly should be better. :)

pigletDips: Ok.

utils/src/compress/zstd_standard.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    let mut dst_buf = Vec::with_capacity(compress_bound as usize);
+    let cmp_size = unsafe {
+        ZSTD_compress(
+            dst_buf.as_mut_ptr() as *mut c_void,
+            compress_bound,
+            src.as_ptr() as *const c_void,
+            src.len(),
+            ZSTD_defaultCLevel(),
```
imeoer, line 24: For chunks `<=1MB`, is the default compression level optimal for performance? @liubogithub @hsiangkao wdyt?

liubogithub: TBH, I have no idea, a few performance tests is needed to prove here.
IMO it can be also updated later once we get some reasonable numbers here.

hsiangkao: > For chunks `<=1MB`, is the default compression level optimal for performance? @liubogithub @hsiangkao wdyt?

My own suggestion is to revise chunk compression approach later by introducing another compressed segment concept, so each compressed segment can roughly have 1MB size.
I don't quite like the current compression approach, it's somewhat ineffective.

hsiangkao: but I think currently we can live with that, since currently compression algorithm is selected on the fs basis. so we have no other choice, so the compression level is not a on-disk chance. We can live with it for now, and benchmark even further later.

imeoer (approved):

utils/src/compress/zstd_standard.rs
imeoer: Ok, let's keep going on.

hsiangkao:
any chance to merge it? so I can test zstd performance as well.

imeoer:
LGTM!

liubogithub (approved):

PR #389, opened 2022-04-14
Add support for fscache protocol v9

hsiangkao:
Update fscache protocol to v9

### DIFF
--- a/src/bin/nydusd/fs_cache.rs
+++ b/src/bin/nydusd/fs_cache.rs
@@ -439,8 +439,8 @@ impl FsCacheHandler {
             };
             format!("copen {},{}", hdr.id, -libc::EALREADY)
         } else {
-            match self.create_data_blob_object(&config, hdr.id, msg.fd) {
-                Err(s) => s,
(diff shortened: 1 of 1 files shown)


src/bin/nydusd/fs_cache.rs
```diff
@@ -104,10 +110,10 @@ impl TryFrom<&[u8]> for FsCacheMsgHeader {
 /// from the fscache driver.
 #[derive(Default, Debug, Eq, PartialEq)]
 struct FsCacheMsgOpen {
-    fd: u32,
-    flags: u32,
     volume_key: String,
     cookie_key: String,
+    fd: u32,
+    flags: u32,
```
jiangliu, line 116: Seems this change is unnecessary, have encountered any issue here?

jiangliu (approved):

src/bin/nydusd/fs_cache.rs
hsiangkao: My own assumption is that is more friendly to CPU cache.

PR #391, opened 2022-04-18
storage: fscache: hide useless chunk map

hsiangkao:
At least, fscache doesn't have any real requirement on chunk map.
Let's drop it directly instead.


hsiangkao:
outdated.

PR #393, opened 2022-04-19
fscache: implement blob unregister api

imeoer:
```
fscache: implement blob unregister api

Add `DELETE /api/v2/blobs?domain_id=?` api to support remove (unregister)
fscache blob cache config, including bootstrap and its blob items.

fscache: refactor blob cache state

Simplify BlobCacheState methods to make code cleaner.
```

### DIFF
--- a/api/src/http.rs
+++ b/api/src/http.rs
@@ -106,7 +106,7 @@ pub struct BlobCacheList {
 
 #[derive(Clone, Deserialize, Debug)]
 pub struct BlobObjectParam {
-    pub blob_id: String,
+    pub domain_id: String,
 }
 
 #[derive(Debug)]
@@ -259,7 +259,7 @@ pub enum HttpError {
     // Blob cache management related errors (v2)
     /// Failed to create blob object
     CreateBlobObject(ApiError),
-    /// Failed to create blob object
+    /// Failed to delete blob object
     DeleteBlobObject(ApiError),
     /// Failed to list existing blob objects
     GetBlobObjects(ApiError),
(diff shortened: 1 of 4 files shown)


api/src/http.rs
```diff
@@ -106,7 +106,7 @@ pub struct BlobCacheList {
 
 #[derive(Clone, Deserialize, Debug)]
 pub struct BlobObjectParam {
-    pub blob_id: String,
+    pub domain_id: String,
```
jiangliu, line 109: It would be better to match blobs by both domain_id and blob_id. 
It means matching all blobs with a common domain_id prefix when the blob_id is empty.

src/bin/nydusd/blob_cache.rs
```diff
@@ -101,6 +101,42 @@ struct BlobCacheState {
...
 
+impl BlobCacheState {
+    fn new() -> Self {
+        Self {
+            id_to_config_map: HashMap::new(),
+        }
+    }
+
+    fn remove(&mut self, domain_id: &str) {
+        let scoped_blob_prefix = format!("{}-", domain_id);
```
jiangliu, line 112: The algorithm to generate the blob key is
```
        let scoped_blob_id = if domain_id.is_empty() {
            blob_info.blob_id().to_string()
        } else {
            domain_id + "-" + blob_info.blob_id()
        };
```
So we should keep in consistence and the `domain_id` may be empty.

jiangliu: The `domain_id` may be used as tenant_id when support multi-tenants.
Or the `domain_id` may be empty when support single tenant.

hsiangkao: for now, for bootstrap, domain_id should equal to blob_id.
I think we could live with it for now, and then think further together with kernel approach evolves...

jiangliu:
Let improve with future PRs。

jiangliu (approved):

PR #394, opened 2022-04-19
doc: add fscache usage with snapshotter

imeoer:


docs/nydus-fscache.md
```diff
@@ -44,17 +44,21 @@ CONFIG_EROFS_FS_ONDEMAND=y
 3. Build nydusd with \
 ``cargo build --target x86_64-unknown-linux-gnu --features=fusedev --release --target-dir target-fusedev --bin nydusd``
```
hsiangkao, line 45: since nydus-image is needed for containerd-nydus-grpc, so we need to build it as well

imeoer: Thanks for the reply, it is not a required option, removed that, and added an image conversion section.

hsiangkao (approved):

jiangliu (approved):

PR #395, opened 2022-04-20
doc: refine fscache user guide with containerd runtime

hsiangkao:

### DIFF
--- a/docs/nydus-fscache.md
+++ b/docs/nydus-fscache.md
@@ -32,9 +32,9 @@ CONFIG_EROFS_FS_ONDEMAND=y
 
 9.  ``[ -c /dev/cachefiles ] && echo ok``
 
-## Get the fscache-supported nydusd
+## Get ctr-remote and the fscache-supported nydusd
 
-1. Make sure you have installed _rust 1.52.1_ version
+1. Make sure you have installed _rust 1.52.1_ version and golang.
 
 2. Check out the latest nydus source code with \
 ``git clone https://github.com/dragonflyoss/image-service.git -b fscache``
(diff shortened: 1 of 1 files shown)


docs/nydus-fscache.md
```diff
@@ -85,20 +92,49 @@ For more information on how to configure containerd to use nydus snapshotter ple
...
+7. Run container with [ctr-remote](../contrib/ctr-remote)
+
+``` shell
 # pull nydus image
-ctr-remote images rpull hsiangkao/ubuntu:20.04-rafs-v6-docker
+contrib/ctr-remote/bin/ctr-remote images rpull docker.io/hsiangkao/ubuntu:20.04-rafs-v6-docker
 
 # run nydus image
-ctr-remote run --rm -t --snapshotter=nydus hsiangkao/ubuntu:20.04-rafs-v6-docker ubuntu /bin/bash
+ctr run --rm -t --snapshotter=nydus docker.io/hsiangkao/ubuntu:20.04-rafs-v6-docker ubuntu /bin/bash
```
imeoer, line 131: We can also run with `ctr-remote`, because `ctr-remote` has already been compiled, so just use it.

hsiangkao: ctr doesn't installed by default? I think using the default tool is always perferred, no?

imeoer: Well, if we assume that `ctr` is already installed, we can use it. `ctr-remote` is actually the same as ctr except a new `rpull` subcommand.

hsiangkao: Yeah, I agree, is there some cases that `ctr` didn't install together with `containerd`?

Issue #348, opened 2022-03-24
Optimize bootstrap size for rafs v6


hsiangkao:
For now, bootstrap sizes of nydus rafs v6 are still far worse than the sizes of what erofs-utils generates.
We should support effective tail-packing inline with a real block allocator for rafs v6 nydus, see:
https://git.kernel.org/pub/scm/linux/kernel/git/xiang/erofs-utils.git/tree/lib/cache.c

liubogithub:
@pigletDips is working on it.

PR #396, opened 2022-04-20
Optimize bootstrap size (WIP)

pigletDips:
Add `dirents_offset: u64` field in `struct Node` to record dirents offset for directory files, to allow dirents don't have to be located at the exact block next to its inode.
Add `available_blocks: Vec<VecDeque<u64>>` field in `struct BootstrapContext`, to record used blocks containing left-over space.


pigletDips:
I change the implementation of `set_v6_offet`, so the CI won't pass.

liubogithub:
#348

PR #398, opened 2022-04-22
Use Caret to specify dependency versions.

wllenyj:
When compiling dragonball-sandbox, an error is reported:                                                        
```                                                                                                             
Error:   --> /home/runner/.cargo/git/checkouts/image-service-7b544e3856b79dd7/cba6aa9/storage/src/utils.rs:27:26
   |                                                                                                            
27 |         match preadv(fd, iovec, offset as off64_t).map_err(|_| last_error!()) {                            
   |                          ^^^^^ types differ in mutability                                                  
   |                                                                                                            
   = note: expected mutable reference `&mut [IoSliceMut<'_>]`                                                   
                      found reference `&[IoVec<&mut [u8]>]`                                                     
```                                                                                                             
This is because when dragonball-sandbox compiles storage, nix uses `0.24.0`.                                    
                                                                                                                
The `preadv` of `0.24.0` is signed as below (https://docs.rs/nix/0.24.0/nix/sys/uio/fn.preadv.html),            
different from the 0.23.1 we are using                                                                          
now (https://docs.rs/nix/0.23.1/nix/sys/uio/fn.preadv.html).                                                    
```                                                                                                             
pub fn preadv(                                                                                                  
    fd: RawFd,                                                                                                  
    iov: &mut [IoSliceMut<'_>],                                                                                 
    offset: off_t                                                                                               
) -> Result<usize>                                                                                              
```                                                                                                             
                                                                                                                
So we should use `Caret requirements` to avoid version compatibility issues.                                    
See                                                                                                             
`https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html`                                        
for details.                                                                                                    
                                                                                                                
According to cargo's definition of [API                                                                         
compatibility](https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/semver.md).                 
Cargo uses the convention that only changes in the left-most non-zero component                                 
are considered incompatible.                                                                                    
                                                                                                                
This patch modifies `Comparison requirements` to `Caret requirements`,                                          
and the version number is based on the version that Cargo.lock now depends on.                                  
                                                                                                                
Signed-off-by: wllenyj <wllenyj@linux.alibaba.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,31 +23,31 @@ path = "src/lib.rs"
 rlimit = "0.3.0"
 log = "0.4.8"
 libc = "0.2"
-vmm-sys-util = ">=0.9.0"
+vmm-sys-util = "0.9.0"
 clap = "2.33"
 flexi_logger = { version = "0.17" }
 # pin regex to fix RUSTSEC-2022-0013
-regex = ">=1.5.5"
-serde = { version = ">=1.0.27", features = ["serde_derive", "rc"] }
+regex = "1.5.5"
+serde = { version = "1.0.110", features = ["serde_derive", "rc"] }
 serde_json = "1.0.51"
 serde_with = { version = "1.6.0", features = ["macros"] }
 sha2 = "0.9.1"
 lazy_static = "1.4.0"
 xattr = "0.2.2"
-nix = ">=0.23.0"
+nix = "0.23.1"
 anyhow = "1.0.35"
-base64 = { version = ">=0.12.0" }
+base64 = "0.13.0"
 rust-fsm = "0.6.0"
 vm-memory = { version = "0.7.0", features = ["backend-mmap"], optional = true }
 chrono = "0.4.19"
 openssl = { version = "0.10.38", features = ["vendored"] }
 hyperlocal = "0.8.0"
-tokio = { version = ">=1.13.1", features = ["macros"] }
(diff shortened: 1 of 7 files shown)


bergwolf:
Does it affect `Cargo.lock` or not? I'd expect `Cargo.lock` to be updated with Cargo.toml.

wllenyj:
> Does it affect `Cargo.lock` or not? I'd expect `Cargo.lock` to be updated with Cargo.toml.

This patch modifies the version number is based on the version that Cargo.lock now depends on. So does not affect `Cargo.lock`.  
Tested with `make build` on my machine.

wllenyj:
@bergwolf What do you think about this pr, if possible, I will modify other Cargo.toml together.

bergwolf:
@wllenyj We can change them as long as `cargo deny` doesn't complain. There are some [comparison requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#comparison-requirements) added due to security reasons. They have an explicit comment like [this one ](https://github.com/dragonflyoss/image-service/blob/master/Cargo.toml#L29)so it is easy to identify.

wllenyj:
@bergwolf I got.  so we need to combine the two problems together to solve.

bergwolf (approved):
Thanks, @wllenyj !

PR #400, opened 2022-04-24
add global log level option for nydusify

liubin:
Add --log-level option and -D as global options for
pack/convert/check subcommands.

The -D is the shortcut for --log-level debug,
and will have the highest priority.

### DIFF
--- a/contrib/nydusify/cmd/nydusify.go
+++ b/contrib/nydusify/cmd/nydusify.go
@@ -35,6 +35,8 @@ var versionGitCommit string
 var versionBuildTime string
 var maxCacheMaxRecords uint = 50
 
+const defaultLogLevel = logrus.InfoLevel
+
 func isPossibleValue(excepted []string, value string) bool {
 	for _, v := range excepted {
 		if value == v {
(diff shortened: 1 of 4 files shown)


contrib/nydusify/cmd/nydusify.go
```diff
@@ -528,3 +539,19 @@ func main() {
...
+	logLevel, err := logrus.ParseLevel(lvl)
+	if err != nil {
+		logrus.Warnf("failed to parse log level(%s): %+v\ndefault log level(info) will be used", lvl, err)
+		logLevel = defaultLogLevel
+	}
+
+	logrus.SetLevel(logLevel)
+
+	// global `-D` has the highest priority
+	if c.Bool("D") {
```
imeoer, line 554: We can check this at the beginning of the func.

liubin: Check it last to let it overwrite the `--log-level`.

imeoer: How about this:

```
// global `-D` has the highest priority
if c.Bool("D") {
	return logrus.SetLevel(logrus.DebugLevel)
}

lvl := c.String("log-level")
logLevel, err := logrus.ParseLevel(lvl)
if err != nil {
	logrus.Warnf("failed to parse log level(%s): %+v\ndefault log level(info) will be used", lvl, err)
	logLevel = defaultLogLevel
}
logrus.SetLevel(logLevel)
```

liubin: Or which should have the higher priority?

imeoer (approved):
LGTM!

PR #402, opened 2022-04-25
fscache: add prefetch_all

nan1994:
If prefetch_config.enable = true, prefetch all data after open data blob.

tested with a configuration json like below:
```
{
  "type": "bootstrap",
  "config": {
    "backend_type": "registry",
    "backend_config": {
      "scheme": "https"
    },
    "cache_type": "fscache"
  },
  "fs_prefetch": {
    "enable": true,
    "threads_count": 2,
    "merging_size": 131072,
    "bandwidth_rate": 0
  }
}
```

### DIFF
--- a/api/src/http.rs
+++ b/api/src/http.rs
@@ -95,6 +95,8 @@ pub struct BlobCacheEntry {
     /// Domain id for the blob, which is used to group cached blobs into management domains.
     #[serde(default)]
     pub domain_id: String,
+    #[serde(default)]
+    pub fs_prefetch: Value,
 }
 
 /// Configuration information for a list of cached blob objects.
--- a/src/bin/nydusd/blob_cache.rs
+++ b/src/bin/nydusd/blob_cache.rs
@@ -277,7 +277,10 @@ impl BlobCacheMgr {
                 cache_compressed: false,
                 cache_config: entry.blob_config.cache_config.clone(),
                 cache_validate: false,
-                prefetch_config: Default::default(),
+                prefetch_config: match serde_json::from_value(entry.fs_prefetch.clone()) {
+                    Ok(fs_prefetch) => fs_prefetch,
+                    Err(_e) => Default::default(),
+                },
             },
         });
 
--- a/src/bin/nydusd/fs_cache.rs
+++ b/src/bin/nydusd/fs_cache.rs
@@ -4,6 +4,7 @@
 
 //! Handler to cooperate with Linux fscache subsystem for blob cache.
 
+use std::cmp;
 use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::fs::{File, OpenOptions};
@@ -18,6 +19,7 @@ use std::sync::{Arc, Barrier, Mutex, MutexGuard};
 use mio::unix::SourceFd;
 use mio::{Events, Interest, Poll, Token, Waker};
 use storage::cache::BlobCache;
+use storage::device::BlobPrefetchRequest;
 use storage::factory::BLOB_FACTORY;
 
 use crate::blob_cache::{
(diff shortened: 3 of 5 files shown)


api/src/http.rs
```diff
@@ -95,6 +95,7 @@ pub struct BlobCacheEntry {
     /// Domain id for the blob, which is used to group cached blobs into management domains.
     #[serde(default)]
     pub domain_id: String,
+    pub fs_prefetch:Value,
```
jiangliu, line 98: A space between "fs_prefetch" and "Value"?

src/bin/nydusd/blob_cache.rs
```diff
@@ -277,7 +277,7 @@ impl BlobCacheMgr {
                 cache_compressed: false,
                 cache_config: entry.blob_config.cache_config.clone(),
                 cache_validate: false,
-                prefetch_config: Default::default(),
+                prefetch_config: serde_json::from_value(entry.fs_prefetch.clone()).unwrap(),
```
jiangliu, line 280: Please avoid using `unwrap()` when possible. We should explicitly check `serde_json::from_value(entry.fs_prefetch.clone())` is valid here. Otherwise an invalid user input from the API server may cause nydusd crash thus DoS attack.

src/bin/nydusd/fs_cache.rs
```diff
@@ -416,16 +418,52 @@ impl FsCacheHandler {
...
                     state.id_to_config_map.insert(hdr.object_id, config.clone());
+                    let _ = self.do_prefetch(&config, blob.clone());
                     format!("copen {},{}", hdr.msg_id, blob_size)
                 }
             }
         }
     }
 
+    pub fn do_prefetch(
+        //prefetch_files: Option<Vec<PathBuf>>,
```
jiangliu, line 434: Please remove this line.

src/bin/nydusd/fs_cache.rs
```diff
@@ -416,16 +418,52 @@ impl FsCacheHandler {
...
+        &self,
+        config: &BlobCacheConfigDataBlob,
+        blob: Arc<dyn BlobCache>,
+    ) {
+        let blob_info = config.blob_info().deref().clone();
+        let factory_config = config.factory_config().deref().clone();
+        if factory_config.cache.prefetch_config.enable {
```
jiangliu, line 441: How about `if !factory_config.cache.prefetch_config.enable { return; }`?

src/bin/nydusd/fs_cache.rs
```diff
@@ -416,16 +418,52 @@ impl FsCacheHandler {
...
+            let size = factory_config.cache.prefetch_config.merging_size as u64;
+            let count = blob_info.compressed_size() / size;
+            let mut blob_rep = Vec::with_capacity(count as usize);
+            let mut pre_offset = blob_info.readahead_offset();
```
jiangliu, line 445: Seems you are trying to prefetch all data of the blob, so why using `blob_info.readahead_offset()` instead of `0` here?

src/bin/nydusd/fs_cache.rs
```diff
@@ -416,16 +418,52 @@ impl FsCacheHandler {
...
         }
     }
 
+    pub fn do_prefetch(
+        //prefetch_files: Option<Vec<PathBuf>>,
+        &self,
+        config: &BlobCacheConfigDataBlob,
+        blob: Arc<dyn BlobCache>,
+    ) {
+        let blob_info = config.blob_info().deref().clone();
```
jiangliu, line 439: Why do we need to clone blob_info and factory_config here?

hsiangkao:
we need to prioritize this feature.

hsiangkao:
We should support whether ``fs_prefetch`` is or isn't given, otherwise, with old nydus-snapshotter it will print
```
ERRO[2022-05-06T03:52:14.336433135Z] failed to mount /3/fs, failed to shared mount: failed to erofs mount: request to bind erofs blob: error code: UNDEFINED, error message: ParseBody(Error("missin
g field `fs_prefetch`", line: 1, column: 650))
```

src/bin/nydusd/fs_cache.rs
```diff
@@ -416,16 +418,49 @@ impl FsCacheHandler {
...
 
+    pub fn do_prefetch(&self, config: &BlobCacheConfigDataBlob, blob: Arc<dyn BlobCache>) {
+        let blob_info = config.blob_info().deref();
+        let factory_config = config.factory_config().deref();
+        if !factory_config.cache.prefetch_config.enable {
+            return;
+        }
+        let size = factory_config.cache.prefetch_config.merging_size as u64;
+        let blob_size=blob_info.compressed_size();
+        let count = blob_size / size + 1;
```
jiangliu, line 441: When blob_size could be divided by size, count will be one larger than expected. So should it be
```
let count = (blob_size + size - 1) / size
```
?

src/bin/nydusd/fs_cache.rs
```diff
@@ -416,16 +418,49 @@ impl FsCacheHandler {
...
+        let mut blob_req = Vec::with_capacity(count as usize);
+        let mut pre_offset = 0 as u64;
```
jiangliu, line 443: let mut pre_offset = 0u64；

src/bin/nydusd/fs_cache.rs
```diff
@@ -416,16 +418,49 @@ impl FsCacheHandler {
...
         }
     }
 
+    pub fn do_prefetch(&self, config: &BlobCacheConfigDataBlob, blob: Arc<dyn BlobCache>) {
+        let blob_info = config.blob_info().deref();
+        let factory_config = config.factory_config().deref();
+        if !factory_config.cache.prefetch_config.enable {
+            return;
+        }
+        let size = factory_config.cache.prefetch_config.merging_size as u64;
```
jiangliu, line 439: need to validate size here:
1. size is not zero, otherwise it may cause crash by invalid input from attackers.
2. size is power of 2.

hsiangkao (approved):

jiangliu (approved):

Issue #403, opened 2022-04-26
nydus doesn't show ".." sometimes


liubogithub:
With a nydus image build with `-v 6`, sometimes it doesn't show ".." in the `ls` output.

```
root@foobar:/# ls -al usr
total 56
drwxr-xr-x 14 root root  229 Apr 21 13:02 .
drwxr-xr-x  2 root root 5381 Apr 21 13:05 bin
drwxr-xr-x  2 root root   27 Apr 18 10:28 games
drwxr-xr-x  2 root root   27 Apr 18 10:28 include
drwxr-xr-x 15 root root  290 Apr 21 13:02 lib
drwxr-xr-x  2 root root   27 Apr 21 13:02 lib32
drwxr-xr-x  2 root root   59 Apr 21 13:03 lib64
drwxr-xr-x  4 root root   64 Apr 21 13:02 libexec
drwxr-xr-x  2 root root   27 Apr 21 13:02 libx32
drwxr-xr-x 10 root root  171 Apr 21 13:02 local
drwxr-xr-x  2 root root 2479 Apr 21 13:05 sbin
drwxr-xr-x 33 root root  615 Apr 21 13:02 share
drwxr-xr-x  2 root root   27 Apr 18 10:28 src
```

PR #405, opened 2022-04-26
Bug fix: the `..` disappear.

pigletDips:
Avoid the root inode use 0 as nid, which may cause the `..` directory disappear in root dir 's subdirs; What's more, change the dirname's length from in character to in byte.

### DIFF
--- a/src/bin/nydus-image/core/bootstrap.rs
+++ b/src/bin/nydus-image/core/bootstrap.rs
@@ -560,7 +560,13 @@ impl Bootstrap {
             (0, 0)
         };
 
-        let orig_meta_addr = bootstrap_ctx.nodes[0].offset;
+        // Make the superblock's meta_blkaddr one block ahead of the inode table,
+        // to avoid using 0 as root nid.
+        // inode offset = meta_blkaddr * block_size + 32 * nid
+        // When using nid 0 as root nid,
+        // the root directory will not be shown by glibc's getdents/readdir.
+        // Because in some OS, ino == 0 represents corresponding file is deleted.
+        let orig_meta_addr = bootstrap_ctx.nodes[0].offset - EROFS_BLOCK_SIZE;
         let meta_addr = if blob_table_size > 0 {
             align_offset(
                 blob_table_offset + blob_table_size + prefetch_table_size as u64,
--- a/src/bin/nydus-image/core/node.rs
+++ b/src/bin/nydus-image/core/node.rs
@@ -600,8 +600,9 @@ impl Node {
                         entry.set_name_offset(nameoff as u16);
                         dir_data.extend(entry.as_ref());
                         entry_names.push(*name);
-
-                        nameoff += name.len() as u64;
+                        // Use length in byte, instead of length in character.
+                        // Because some characters could occupy more than one byte.
+                        nameoff += name.as_bytes().len() as u64;
                     }
 
                     for name in entry_names.iter() {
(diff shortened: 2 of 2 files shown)


jiangliu (approved):

jiangliu:
Fixes: #403

liubogithub (commented):
Can you please make the commit log even more verbose given the problem is a bit tricky?

As it fixes a problem in nydus-image binary, a proper title can be `"nydus-image: fix xxx problem"`.

In the description part, please explain why the file of "inode id == 0" doesn't get shown by glibc's `getdents`/`readdir` wrapper (a link for reference would be appreciated).

src/bin/nydus-image/core/node.rs
```diff
@@ -600,8 +600,8 @@ impl Node {
                         entry.set_name_offset(nameoff as u16);
                         dir_data.extend(entry.as_ref());
                         entry_names.push(*name);
-
-                        nameoff += name.len() as u64;
+                        // Use length in byte, instead of length in character.
```
liubogithub, line 603: Please make a separate patch for this part.  They look irrelevant to the title.

src/bin/nydus-image/core/node.rs
```diff
@@ -1073,13 +1073,15 @@ impl Node {
...
+        // Use length in byte, instead of length in character.
+        let mut d_size: u64 = (".".as_bytes().len()
+            + size_of::<RafsV6Dirent>()
+            + "..".as_bytes().len()
+            + size_of::<RafsV6Dirent>()) as u64;
 
         // Safe as we have check tree above.
         for child in tree.children.iter() {
-            let len = child.node.name().len() + size_of::<RafsV6Dirent>();
+            let len = child.node.name().as_bytes().len() + size_of::<RafsV6Dirent>();
```
liubogithub, line 1084: Ditto.

src/bin/nydus-image/core/bootstrap.rs
```diff
@@ -560,7 +560,8 @@ impl Bootstrap {
             (0, 0)
         };
 
-        let orig_meta_addr = bootstrap_ctx.nodes[0].offset;
+        // To avoid root inode use nid 0
```
liubogithub, line 563: Please explain the 'why to avoid' if possible.

liubogithub (changes requested):

src/bin/nydus-image/core/bootstrap.rs
```diff
@@ -560,7 +560,8 @@ impl Bootstrap {
...
+        let orig_meta_addr = bootstrap_ctx.nodes[0].offset - EROFS_INODE_SLOT_SIZE as u64;
```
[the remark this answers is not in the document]
pigletDips:
```rust
let orig_meta_addr = bootstrap_ctx.nodes[0].offset;
let root_nid = calculate_nid(
            bootstrap_ctx.nodes[0].offset - orig_meta_addr + meta_addr,
            meta_addr,
        );

pub fn calculate_nid(offset: u64, meta_size: u64) -> u64 {
    (offset - meta_size) >> EROFS_INODE_SLOT_BITS
}
```
If we just set meta_addr to meta_addr - EROFS_BLOCK_SIZE, I think root_nid will still be 0.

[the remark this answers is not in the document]
hsiangkao: Take an example for explaination...
If the root inode starts at the block 1, if we set on-disk meta_blkaddr = 1, so the nid would be 0;
But if we set on-disk meta_blkaddr = 0, so the nid would be = 4096/32 = 128 instead.

src/bin/nydus-image/core/bootstrap.rs
```diff
@@ -560,7 +560,10 @@ impl Bootstrap {
             (0, 0)
         };
 
-        let orig_meta_addr = bootstrap_ctx.nodes[0].offset;
+        // Leave one block between the superblock and the inodes
+        // to avoid root inode use 0 as nid,
+        // which may cause the `..` disappear in some subdirs of root directory.
```
bergwolf, line 565: The "WHY" is not really clear to me. Why would it cause `..` to disappear in some subdirs of the root directory?

Also when the comments says `Leave one block between the superblock and the inodes`, the meta address is moved one block backward. Wouldn't it cause bootstrap to overlap with the inode table?

bergwolf, line 565: The "WHY" is not really clear to me. Why would it cause `..` to disappear in some subdirs of the root directory?

Also when the comments says `Leave one block between the superblock and the inodes`, the meta address is moved one block backward. Wouldn't it cause bootstrap to overlap with the inode table?

hsiangkao: yeah, actually I understand what it really means. but the commit might need do some refinement anyway.

liubogithub (commented):
For all contributors's reference as well,  the format of a typical bugfix format is
```
$(subsystem): $(what is this patch for)

$(explain why the following changes are needed UNLESS they are way too simple to understand)

Signed-off-by: 
```

The purpose of doing above is to help people better follow what happened here.

As for this patch, please explain why `as_bytes()` is preferred or what corner case would be hit without the patch.

liubogithub (approved):

hsiangkao (approved):

PR #412, opened 2022-04-28
nydus-image: Feature, add a block allocator for reducing bootstrap size by reusing used blocks

pigletDips:
Firstly, we add a `dirents_offset` field in `struct Node`, which is used by directory files.
Previously, the dirents have to be placed the next block where its dir inode was placed, by
introducing this field, we allow the dirents to be seperated from its dir inode.
Secondly, we add an `available_blocks` field in `struct BootstrapCtx`,
to record blocks that are used but still contain unused space.

feature #348

### DIFF
--- a/src/bin/nydus-image/core/node.rs
+++ b/src/bin/nydus-image/core/node.rs
@@ -23,7 +23,7 @@ use sha2::digest::Digest;
 use nydus_utils::{
     compress,
     digest::{DigestHasher, RafsDigest},
-    div_round_up, try_round_up_4k, ByteSize,
+    div_round_up, round_down_4k, round_up, try_round_up_4k, ByteSize,
 };
 use rafs::metadata::cached_v5::{CachedChunkInfoV5, CachedInodeV5};
 use rafs::metadata::direct_v5::{DirectChunkInfoV5, OndiskInodeWrapper};
@@ -234,6 +234,8 @@ pub struct Node {
     pub offset: u64,
     /// Used by Rafsv6 to build a Dir inode
     pub dirents: Vec<(u64, OsString, u32)>,
+    /// Used by v6 directory file, indicates its dirents offset
+    pub dirents_offset: u64,
 }
 
 impl Display for Node {
@@ -294,6 +296,7 @@ impl Node {
             v6_datalayout: EROFS_INODE_FLAT_PLAIN,
             v6_force_extended_inode,
             v6_compact_inode: false,
+            dirents_offset: 0,
         };
 
         node.build_inode(chunk_size)
@@ -560,12 +563,10 @@ impl Node {
 
         // update all the inodes's offset according to the new 'meta_addr'.
         self.offset = self.offset - orig_meta_addr + meta_addr;
+        self.dirents_offset = self.dirents_offset + meta_addr - orig_meta_addr;
         if self.is_dir() {
             // the 1st 4k block after dir inode.
-            let mut dirent_off = align_offset(
-                self.offset + self.size_with_xattr() as u64,
-                EROFS_BLOCK_SIZE,
-            );
+            let mut dirent_off = self.dirents_offset;
 
             inode.set_u((dirent_off / EROFS_BLOCK_SIZE) as u32);
             // Dump inode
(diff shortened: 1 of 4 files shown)


src/bin/nydus-image/core/node.rs
```diff
@@ -1150,31 +1152,61 @@ impl Node {
...
+        // We use a simple inline strategy here:
+        // If the inode size with xattr + tail data size <= EROFS_BLOCK_SIZE,
+        // we choose to inline it.
+        // Firstly, if it's bigger that EROFS_BLOCK_SIZE,
+        // in most cases, we can assume that the tail data size is close to EROFS_BLOCK_SIZE,
+        // in this condition, even if we don't inline the tail data, there won't be much waste.
+        // Secondly, the `available_blocks` that we maintain in the `BootstrapCtx`,
+        // since it contain only single blocks with some unsed space, the available space can only be smaller that EROFS_BLOCK_SIZE,
+        // therefore we can't use our used blocks to store the inode plus the tail data bigger than EROFS_BLOCK_SIZE.
+        let should_inline = tail != 0 && (inode_size + tail) <= EROFS_BLOCK_SIZE;
```
hsiangkao, line 1166: how about inode_size + inline xattr + tail > EROFS_BLOCK_SIZE but inline xattr % 4k + tail < EROFS_BLOCK_SIZE?

hsiangkao: I suggest that you need to add some test cases for these, also you need to consider if nydus will use EROFS compressed inodes as well, so generic logic is preferred.

src/bin/nydus-image/core/node.rs
```diff
@@ -1150,31 +1152,61 @@ impl Node {
...
                 bootstrap_ctx.align_offset(EROFS_BLOCK_SIZE);
-                EROFS_INODE_FLAT_PLAIN
             }
+            self.dirents_offset = bootstrap_ctx.offset;
+            bootstrap_ctx.offset += round_down_4k(d_size);
+
+            EROFS_INODE_FLAT_INLINE
+        // Otherwise, we first try to allocate space for the inode from used blocks.
+        // If no available used block exits, we allocate space sequentially.
+        // Then we allocate space for all data.
```
liubogithub, line 1196: Can you move the comments below `} else {`?

src/bin/nydus-image/core/node.rs
```diff
@@ -1150,31 +1152,61 @@ impl Node {
...
+                bootstrap_ctx.append_available_block(bootstrap_ctx.offset);
                 bootstrap_ctx.align_offset(EROFS_BLOCK_SIZE);
-                EROFS_INODE_FLAT_PLAIN
             }
+            self.dirents_offset = bootstrap_ctx.offset;
+            bootstrap_ctx.offset += round_down_4k(d_size);
+
+            EROFS_INODE_FLAT_INLINE
+        // Otherwise, we first try to allocate space for the inode from used blocks.
+        // If no available used block exits, we allocate space sequentially.
```
liubogithub, line 1195: s/exits/exists/g

src/bin/nydus-image/core/node.rs
liubogithub: `compressed inodes`?
does it refer to `compact inodes`?

@hsiangkao

liubogithub (changes requested):

src/bin/nydus-image/core/node.rs
hsiangkao: no, I was thinking we would add compressed inodes maybe later, and use flat inodes not only for directory inodes.
The current approach is not quite flexible on my side.

pigletDips:
image.boot size comparison before this patch and after this patch
This image is a standard ubuntu:latest image with linux-5.17.3 src and a self compiled fssum tool.

pigletDips:
fssum result
`./fssum -MACUG ./linux-5.17.3`
This `fssum` has been slightly modified to bypass the device id check.

pigletDips:
bootstrap
Generated bootstrap size comparison between`mkfs.erofs` and `nydus-image` after patch(without chunk info array).
The `nydus-image` generated bootstrap is slightly smaller.

pigletDips:
fsck
The `erofs.fsck` result of our bootstrap file.

pigletDips:
@hsiangkao OK, I am working on it.

liubogithub (commented):
With the comments addressed, the patch looks good to me.

hsiangkao:
> without chunk info array

Yeah, so the chunk info array from v5 really sucks, we should get rid of it and use another v6 version as soon as possible.

pigletDips:
截屏2022-04-29 16 19 46
The fssum result of linux-5.17.3(with xattr) between nydus v6 image and OCI image.

hsiangkao (approved):

pigletDips:
截屏2022-04-29 15 31 36
The bootstrap size of linux-5.17.3(with xattr) between nydus-image and mkfs.erofs.

liubogithub (approved):

PR #413, opened 2022-04-29
builder: fix lost hardlink

imeoer:
The merge operation in builder (nydus-image) use `RafsSuper::walk_inodes`
method to walkthrough the whole filesystem inodes by DFS order, but it
does not call the callback for the hardlink pair, which causes some hardlink
files to be lost in the final bootstrap after merge.

The root cause is that the `RafsSuper::get_inode` method always gets only
the RafsInode of a particular file name in the hardlink pair because the ino
number of the hardlink pair are the same.

Fix by passing `RafsInode` trait object in `RafsSuper::walk_inodes` directly.

### DIFF
--- a/rafs/src/metadata/mod.rs
+++ b/rafs/src/metadata/mod.rs
@@ -17,6 +17,8 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
+use anyhow::bail;
+
 use fuse_backend_rs::abi::fuse_abi::Attr;
 use fuse_backend_rs::api::filesystem::{Entry, ROOT_ID};
 use nydus_utils::compress;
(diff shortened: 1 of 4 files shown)


liubogithub (commented):
It seems to be an important fix, any test case to keep it from happening in the future?

liubogithub (commented):
Correct me if I am wrong.

`walk_inodes` is only used to getting chunk info from an inode, but hardlink inode pair should share the same chunk infos, so how are lost hardlink files related to `walk_inodes`?

imeoer:
> It seems to be an important fix, any test case to keep it from happening in the future?

@liubogithub Will add a test case to the [converter](https://github.com/containerd/nydus-snapshotter/tree/main/pkg/converter) package since `walk_inodes` is currently used mainly for tar build.

rafs/src/metadata/mod.rs
```diff
@@ -744,27 +744,37 @@ impl RafsSuper {
         Ok(())
     }
 
+    /// Walkthrough inode tree by DFS order, please ensure ino is not a hardlink,
+    /// otherwise will lose some hardlink pairs.
     pub fn walk_inodes(
```
bergwolf, line 749: How do we ensure `ino` is not a hardlink? I don't see anywhere we put such enforcement.

imeoer: There is no such enforcement, we usually call this method for the directory, and it is currently only used for root ino.

bergwolf: IMO,
1. If we only use it for root ino (no plan for other usage), let's expose a `walk_root_tree`-like method so that it will not be called in other ways.
2. If we plan to use it for other inodes other than root, we need to add a check in `walk_inodes` to make sure the passed in inode is not a hardlink.

imeoer: Agree with point 2, but I think it makes no sense to walk a non-directory for now, how about checking that ino is a directory first? Also change the func name to `walk_dir`. This is easier than checking if an ino has a hardlink pair.

imeoer: Fixed

bergwolf (approved):

PR #415, opened 2022-04-29
builder: implement tar like build

imeoer:
Enable `inline-bootstrap` option to merge the blob and bootstrap into one
file. We need some header to index the location of the blob and bootstrap,
this patch uses a tar header structure that arranges the data as follows:

`blob_data | blob_tar_header | bootstrap_data | boostrap_tar_header`

This is a tar-like structure, except that we put the tar header after the
data. The advantage is that we do not need to determine the size of the data
first, so that we can write the blob data by stream without seek to improve
the performance of the blob dump by using fifo, if we need to read the bootstrap
data quickly, first need to read the 512 bytes tar header from the end of blob
file first, and then seek offset to read bootstrap data.

Related PR: https://github.com/containerd/nydus-snapshotter/pull/44

### DIFF
--- a/src/bin/nydus-image/builder/diff.rs
+++ b/src/bin/nydus-image/builder/diff.rs
@@ -305,7 +305,8 @@ fn dump_blob(
     blob_nodes: &mut Vec<Node>,
     chunk_dict: Arc<dyn ChunkDict>,
 ) -> Result<(Option<BlobContext>, ChunkMap)> {
-    let mut blob_ctx = BlobContext::new(blob_id, blob_storage, ctx.blob_offset)?;
+    let mut blob_ctx =
+        BlobContext::new(blob_id, blob_storage, ctx.blob_offset, ctx.inline_bootstrap)?;
     blob_ctx.set_chunk_dict(chunk_dict);
     blob_ctx.set_chunk_size(ctx.chunk_size);
     blob_ctx.set_meta_info_enabled(ctx.fs_version == RafsVersion::V6);
@@ -324,7 +325,10 @@ fn dump_blob(
         blob_nodes,
         &mut chunk_cache,
     )? {
-        blob_ctx.finalize()?;
+        let blob_id = blob_ctx.blob_id();
+        if let Some(writer) = &mut blob_ctx.writer {
+            writer.finalize(blob_id)?;
+        }
         Some(blob_ctx)
     } else {
         None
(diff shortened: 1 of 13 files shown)


src/bin/nydus-image/builder/diff.rs
```diff
@@ -792,7 +797,7 @@ impl DiffBuilder {
         for (snapshot_idx, worker) in workers.into_iter().enumerate() {
             // FIXME: the behavior of build with diff is not working.
```
bergwolf, line 798: Is this still the case?

src/bin/nydus-image/builder/directory.rs
```diff
@@ -134,10 +138,12 @@ impl Builder for DirectoryBuilder {
         )?;
 
         // Dump blob file
+        let fifo = ctx.inline_bootstrap;
         let mut blob_ctx = BlobContext::new(
             ctx.blob_id.clone(),
             ctx.blob_storage.clone(),
             ctx.blob_offset,
+            fifo,
```
bergwolf, line 146: Just use `ctx.inline_bootstrap`.

src/bin/nydus-image/builder/stargz.rs
```diff
@@ -577,7 +577,9 @@ impl StargzBuilder {
         let mut decompressed_blob_size = 0u64;
         let mut compressed_blob_size = 0u64;
         let blob_index = blob_mgr.alloc_index()?;
-        let mut blob_ctx = BlobContext::new(ctx.blob_id.clone(), ctx.blob_storage.clone(), 0)?;
+        let fifo = ctx.inline_bootstrap;
+        let mut blob_ctx =
+            BlobContext::new(ctx.blob_id.clone(), ctx.blob_storage.clone(), 0, fifo)?;
```
bergwolf, line 582: can drop the `fifo` variable.

src/bin/nydus-image/builder/diff.rs
imeoer: The diff build includes two methods，`build_with_hint` (for overlayfs snapshot) and `build_with_diff` (for non-overlayfs snapshot, for example devmapper snapshot), currently we only use `build_with_hint` in acceld + containerd + overlayfs snapshotter, the support of non-overlayfs snapshotter has not been tested in detail, so a FIXME is still needed here.

src/bin/nydus-image/builder/directory.rs
imeoer: Fixed.

src/bin/nydus-image/builder/stargz.rs
imeoer: Fixed.

bergwolf (approved):

PR #418, opened 2022-05-07
cat heredoc will not work with sudo

wanderxjtu:
`sudo cat > SOMEPATH <<EOF` won't work when user has no permission to write to SOMEPATH.
Using `sudo tee SOMEPATH` instead.


bergwolf (approved):
Thanks!

bergwolf:
Thanks @wanderxjtu!

It is a good practice to add a subsystem prefix to the commit message title and a few words to better describe it. So something like this would look better:

```
docs: fix cat script example

sudo cat > SOMEPATH <<EOF won't work when user has no permission to write to SOMEPATH.
Using sudo tee SOMEPATH instead.

Signed-off-by wanderxjtu <wanderhuang@gmail.com>
```

wanderxjtu:
@bergwolf commit message updated, thx.

bergwolf:
@wanderxjtu Thanks a lot!

PR #421, opened 2022-05-12
rafs: delete RafsSuperBlobs trait

liubin:
method get_blobs of RafsSuperBlobs has the same signature as
get_blob_infos of RafsSuperBlock,
and RafsSuperBlock is inheritance of RafsSuperBlobs.

But the only one use of the get_blobs is based on type of
RafsSuperBlock, so RafsSuperBlobs is no meanings, it can
be deleted and replaced by get_blob_infos method.

### DIFF
--- a/rafs/src/fs.rs
+++ b/rafs/src/fs.rs
@@ -488,7 +488,7 @@ impl Rafs {
         // Without too much layout concern, just prefetch a certain range from backend.
         let prefetches = sb
             .superblock
-            .get_blobs()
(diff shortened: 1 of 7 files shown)


bergwolf:
Thanks @liubin ! Looks reasonable to me. @jiangliu please confirm if there is future plan on the trait, as it is redundant right now.

imeoer (approved):

jiangliu (approved):

PR #429, opened 2022-05-13
action: release macos bianries as well

bergwolf:
Right now, only nydusctl and nydusd compile on macos. So just include
them in the release tarball. Other binaries can be included later.

Example run can be found here: https://github.com/bergwolf/image-service/runs/6420841588

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  build-nydus-rs:
+  nydus-linux:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -80,46 +120,90 @@ jobs:
...
-  upload-artifacts:
+
+  prepare-tarball-linux:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         arch: [amd64, arm64]
-    needs: [build-nydus-rs, build-contrib]
+        os: [linux]
+    needs: [nydus-linux, nydus-macos, contrib-linux]
```
imeoer, line 137: `nydus-macos` is unnecessary.

.github/workflows/release.yml
```diff
@@ -80,46 +120,90 @@ jobs:
...
+        path: ${{ env.tarball }}
+
+  # use a seperate job for darwin because github action if: condition cannot handle && properly.
+  prepare-tarball-darwin:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: [amd64]
+        os: [darwin]
+    needs: [nydus-linux, nydus-macos, contrib-linux]
```
imeoer, line 164: `nydus-linux` is unnecessary.

.github/workflows/release.yml
```diff
@@ -80,46 +120,90 @@ jobs:
...
+    - name: push release
+      if: github.event_name == 'push'
       uses: softprops/action-gh-release@v1
       with:
         name: "Nydus Image Service ${{ env.tag }}"
         generate_release_notes: true
-        files:
-          ${{ env.tarball }}
+        files: |
+          ${{ env.tarball1 }}
```
imeoer, line 207: Can we use an `env.tarballs` value which is split by `\n` here to simplify this?

bergwolf: The newline will be eaten by $GITHUB_ENV file and we can only retrive a single line there.

bergwolf: fixed

imeoer (approved):
LGTM!

PR #431, opened 2022-05-13
rafs v6: bug fix for the incorrect dirents process

### LINKED ISSUE #428: Nydusd-fuse panic when running in v6 format
image

pigletDips:
The dir file's dirents may occupy multiple blocks, and only the last
block may not be fully used. Previously, the last dirent of each block
is treated as the last dirent of the last block, which causes the
nydusd-fusedev to panic when using rafs v6 format.

To fix this bug, we process the last dirent of the last block different
from the last dirent of the other blocks.

fix #428 

### DIFF
--- a/rafs/src/metadata/direct_v6.rs
+++ b/rafs/src/metadata/direct_v6.rs
@@ -463,16 +463,23 @@ impl OndiskInodeWrapper {
                 let s = (de.e_nameoff - head_de.e_nameoff) as u64
                     + (size_of::<RafsV6Dirent>() * max_entries) as u64;
 
-                let e = slice::from_raw_parts(
-                    block_mapping.add(de.e_nameoff as usize),
-                    (self.size() % EROFS_BLOCK_SIZE - s) as usize,
-                );
+                // The possible maximum len of the last dirent's file name should be calculated
+                // differently depends on whether the dirent is at the last block of the dir file.
+                // Because the other blocks should be fully used, while the last may not.
+                let len = if div_round_up(self.size(), EROFS_BLOCK_SIZE) as usize == block_index + 1
+                {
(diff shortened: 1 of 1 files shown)


rafs/src/metadata/direct_v6.rs
```diff
@@ -463,16 +463,20 @@ impl OndiskInodeWrapper {
                 let s = (de.e_nameoff - head_de.e_nameoff) as u64
                     + (size_of::<RafsV6Dirent>() * max_entries) as u64;
 
-                let e = slice::from_raw_parts(
-                    block_mapping.add(de.e_nameoff as usize),
-                    (self.size() % EROFS_BLOCK_SIZE - s) as usize,
-                );
+                let len = if div_round_up(self.size(), EROFS_BLOCK_SIZE) as usize == block_index + 1
```
liubogithub, line 466: A comment is appreciated.

liubogithub (approved):
Looks good.

Can you please add a test case for it?

rafs/src/metadata/direct_v6.rs
pigletDips: OK

pigletDips:
> Looks good.
> 
> Can you please add a test case for it?

The test case may be a little hard to implement. This mod lacks tests, and the dir inode and its dirents are pretty hard to construct.

hsiangkao:
> > Looks good.
> > Can you please add a test case for it?
> 
> The test case may be a little hard to implement. This mod lacks tests, and the dir inode and its dirents are pretty hard to construct.

since it's already broken(because fuse use cases are not what we're mainly addressing) let's resolve it as another PR.

hsiangkao (approved):

Issue #424, opened 2022-05-12
Support rafs v6 in smoke test


liubogithub:
Although we have rafs v6 fuse support in code, the smoke test still lacks support for rafs v6 format.

It's supposed to work for rafs v6 as well as for rafs v5.

liubogithub:
#438

bergwolf:
fixed by #438

PR #438, opened 2022-05-18
smoke: add basic testing for rafs v6

imeoer:
Let's make smoke testing to build rafs v6 layer, then mount using nydusd,
and verify filesystem consistency basically.

Since nydusd does not support cached mode and dummycache for rafs v6, the
part of test cases are not enabled.

Also no load is implemented for rafs v6 parent bootstrap, the layered build
test case is temporarily disabled.

Signed-off-by: Yan Song <yansong.ys@antfin.com>

### DIFF
--- a/tests/builder.rs
+++ b/tests/builder.rs
@@ -189,19 +189,26 @@ impl<'a> Builder<'a> {
         );
     }
 
-    pub fn build_lower(&mut self, compressor: &str) {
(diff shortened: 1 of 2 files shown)


liubogithub (approved):
#424 
Thanks so much for making it work.
I'll check if v6 load can be fixed.

PR #439, opened 2022-05-18
refact: nydusd's state machine

ccx1024cc:
### QUESTION
`UPGRADING` and `INTERRUPTED` are designed for hot-upgrading, which is a use case instead of daemon's state. As a result, they are weird in lack of upgrading function.

### DESIGN
Merge `UPGRADING` and `INTERRUPTED` to `READY`. `READY` means service is well-configured, but waiting for trigger. For fuse-based daemon, it means fuse device is mounted, while fuse mountpoint may be not attached. The daemon lives in `INIT` -> `READY` -> `RUNNING` -> `READY` -> `DIE`.

### CHANGE
1. Daemon context with new status.
2. Async `TerminateFuseService`, `Umount` actions.
3. `Start` endpoint, which is usually used after `takeover` endpoint.
4. Async exit after fuse device unmounted abnormally.

### DIFF
--- a/api/src/http_endpoint_v1.rs
+++ b/api/src/http_endpoint_v1.rs
@@ -245,6 +245,23 @@ impl EndpointHandler for MountHandler {
     }
 }
 
+pub struct StartHandler {}
+impl EndpointHandler for StartHandler {
+    fn handle_request(
+        &self,
+        req: &Request,
+        kicker: &dyn Fn(ApiRequest) -> ApiResponse,
+    ) -> HttpResult {
+        match (req.method(), req.body.as_ref()) {
+            (Method::Put, None) => {
+                let r = kicker(ApiRequest::Start);
+                Ok(convert_to_response(r, HttpError::Upgrade))
+            }
+            _ => Err(HttpError::BadRequest),
+        }
+    }
+}
+
 pub struct MetricsInflightHandler {}
 impl EndpointHandler for MetricsInflightHandler {
     fn handle_request(
--- a/src/bin/nydusd/daemon.rs
+++ b/src/bin/nydusd/daemon.rs
@@ -35,10 +35,9 @@ use crate::upgrade::UpgradeMgrError;
 pub enum DaemonState {
     INIT = 1,
     RUNNING = 2,
-    UPGRADING = 3,
-    INTERRUPTED = 4,
-    STOPPED = 5,
-    UNKNOWN = 6,
+    READY = 3,
+    STOPPED = 4,
+    UNKNOWN = 5,
 }
 
 impl Display for DaemonState {
@@ -52,9 +51,8 @@ impl From<i32> for DaemonState {
         match i {
             1 => DaemonState::INIT,
             2 => DaemonState::RUNNING,
-            3 => DaemonState::UPGRADING,
-            4 => DaemonState::INTERRUPTED,
-            5 => DaemonState::STOPPED,
+            3 => DaemonState::READY,
+            4 => DaemonState::STOPPED,
             _ => DaemonState::UNKNOWN,
         }
     }
@@ -196,23 +194,49 @@ pub trait NydusDaemon: DaemonStateMachineSubscriber + Send + Sync {
     fn interrupt(&self) {}
     fn stop(&self) -> DaemonResult<()> {
         let s = self.get_state();
-        if s != DaemonState::INTERRUPTED && s != DaemonState::STOPPED {
-            return self.on_event(DaemonStateMachineInput::Stop);
+
+        if s == DaemonState::STOPPED {
+            return Ok(());
         }
-        Ok(())
+
+        if s == DaemonState::RUNNING {
+            self.on_event(DaemonStateMachineInput::Stop)?;
+        }
+
+        self.on_event(DaemonStateMachineInput::Stop)
     }
     fn wait(&self) -> DaemonResult<()>;
+    fn wait_service(&self) -> DaemonResult<()> {
+        Ok(())
+    }
+    fn wait_state_machine(&self) -> DaemonResult<()> {
+        Ok(())
+    }
     fn trigger_exit(&self) -> DaemonResult<()> {
+        let s = self.get_state();
+
+        if s == DaemonState::STOPPED {
+            return Ok(());
+        }
+
+        if s == DaemonState::INIT {
+            return self.on_event(DaemonStateMachineInput::Stop);
+        }
+
+        if s == DaemonState::RUNNING {
+            self.on_event(DaemonStateMachineInput::Stop)?;
+        }
+
         self.on_event(DaemonStateMachineInput::Exit)
     }
-
     fn supervisor(&self) -> Option<String>;
     fn save(&self) -> DaemonResult<()>;
     fn restore(&self) -> DaemonResult<()>;
     fn trigger_takeover(&self) -> DaemonResult<()> {
-        self.on_event(DaemonStateMachineInput::Takeover)?;
-        self.on_event(DaemonStateMachineInput::Successful)?;
-        Ok(())
+        self.on_event(DaemonStateMachineInput::Takeover)
+    }
+    fn trigger_start(&self) -> DaemonResult<()> {
+        self.on_event(DaemonStateMachineInput::Start)
     }
 
     // For backward compatibility.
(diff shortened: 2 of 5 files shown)


api/src/http_endpoint_v1.rs
```diff
@@ -257,6 +257,23 @@ impl EndpointHandler for TakeoverHandler {
...
 
+pub struct StartHandler {}
+impl EndpointHandler for StartHandler {
+    fn handle_request(
+        &self,
+        req: &Request,
+        kicker: &dyn Fn(ApiRequest) -> ApiResponse,
+    ) -> HttpResult {
+        match (req.method(), req.body.as_ref()) {
+            (Method::Put, None) => {
```
imeoer, line 268: `PUT` method in RESTful API design means that the operation can be executed multiple times and always have an idempotent state, is this what the API expects?

ccx1024cc: It's expected that daemon in `READY` is trigger to `RUNNING`. Just like takeover endpoint with PUT method, it's required for daemon in certain status.

ccx1024cc: > `PUT` method in RESTful API design means that the operation can be executed multiple times and always have an idempotent state, is this what the API expects?

It's expected that daemon in READY is trigger to RUNNING. Just like takeover endpoint with PUT method, it's required for daemon in certain status.

src/bin/nydusd/daemon.rs
```diff
@@ -248,12 +258,27 @@ pub trait NydusDaemon: DaemonStateMachineSubscriber {
...
+        if s == DaemonState::STOPPED {
+            return Ok(());
+        }
+
+        if s == DaemonState::INIT {
+            return self.on_event(DaemonStateMachineInput::Stop);
+        }
+
+        if s == DaemonState::RUNNING {
+            self.on_event(DaemonStateMachineInput::Stop)?;
```
changweige, line 272: Why not merge the two events handling if blocks.
They both react on `Stop` event

src/bin/nydusd/fusedev.rs
```diff
@@ -175,14 +176,20 @@ impl FusedevDaemon {
...
             .name("fuse_server".to_string())
             .spawn(move || {
-                let _ = s.svc_loop(&inflight_op);
-                // quit the daemon if any fuse server thread exits
-                exit_daemon();
+                if let Err(err) = s.svc_loop(&inflight_op) {
+                    warn!("fuse server exit with err: {:?}, exit daemon", err);
+                    if let Err(e) = thread::Builder::new()
+                        .name("exit_thread".to_string())
+                        .spawn(exit_daemon)
```
changweige, line 183: Why `exit_daemon` has to be invoked in another thread? A possible reason is state machine now has blocking operation inside?

src/bin/nydusd/daemon.rs
ccx1024cc: The `INIT` status can transfer to `DIE` directly, while the `RUNNING` is transferred to `DIE` by `READY`. So After `stop` on `RUNNING`, there should be another `exit` event.

ccx1024cc: > Why not merge the two events handling if blocks. They both react on `Stop` event

The `INIT` can transfer to `DIE` by `stop` directly. While the `RUNNING` is transferred to `DIE` via `READY`. For `INIT`, there is only one `stop` event, instead of two `stop` events for `RUNNING`.

src/bin/nydusd/fusedev.rs
ccx1024cc: > Why `exit_daemon` has to be invoked in another thread? A possible reason is state machine now has blocking operation inside?

`A possible reason is state machine now has blocking operation inside` That is the point.  As a result, thread exiting is decoupling with process exiting by new thread.

changweige:
Please check the patch's changelog, it is somewhat a mess of duplicated messages.  Can you squash the SOB lines?

ccx1024cc:
> Please check the patch's changelog, it is somewhat a mess of duplicated messages. Can you squash the SOB lines? I suggest adding this PR's description into the commit message.

OK~

ccx1024cc:
> Please check the patch's changelog, it is somewhat a mess of duplicated messages. Can you squash the SOB lines? I suggest adding this PR's description into the commit message.

Good idea! 3q

src/bin/nydusd/fusedev.rs
```diff
@@ -230,8 +237,32 @@ impl NydusDaemon for FusedevDaemon {
...
+        let mut guard = self.state_machine_thread.lock().unwrap();
+        if guard.is_some() {
+            guard
+                .take()
+                .unwrap()
+                .join()
+                .map_err(|e| {
+                    DaemonError::WaitDaemon(
+                        *e.downcast::<Error>()
+                            .unwrap_or_else(|e| Box::new(eother!(e))),
```
changweige, line 254: Um. Hard error conversion path to `io::Error`, maybe let `WaitDaemon` wrap another type of error message is an easier way. But has nothing to do with this PR.

src/bin/nydusd/fusedev.rs
changweige, line 183: I still don't catch the idea why the `exit_daemon` has to be invoked asynchronously.
Looks that nydusd's main thread does nothing eventually but wait for state machine and fuse service loop termination. `exit_daemon` only pushes `Stop` event. Can we just wait for statemachine thread here.  And we stop state machine by a carefully predefined event  `StopStateMachine`, which must not involve blocking operations. Then we don't need to spawn more threads which looks a little weird.

src/bin/nydusd/fusedev.rs
```diff
@@ -477,12 +508,13 @@ pub fn create_nydus_daemon(
         inflight_ops: Mutex::new(Vec::new()),
         result_receiver: Mutex::new(result_receiver),
         trigger: Arc::new(Mutex::new(trigger)),
-        threads: Mutex::new(Vec::new()),
+        state_machine_thread: Mutex::new(None),
+        fuse_service_threads: Mutex::new(Vec::new()),
```
changweige, line 512: Why is reaping fuse threads important?
Only wait for state machine termination?

src/bin/nydusd/fusedev.rs
ccx1024cc: > I still don't catch the idea why the `exit_daemon` has to be invoked asynchronously. Looks that nydusd's main thread does nothing eventually but wait for state machine and fuse service loop termination. `exit_daemon` only pushes `Stop` event. Can we just wait for statemachine thread here. And we stop state machine by a carefully predefined event `StopStateMachine`, which must not involve blocking operations. Then we don't need to spawn more threads which looks a little weird.

The worker thread is responsibility to yell 'F**k by unexpected error', however without responsibility to guarantee   process exiting. If state machine or processs crews up, the worker threads still exits. The worker takes no care about whether the state machine or process exits or not.
But the way to throw out signal is absolutely weird by spawning new thread. Maybe change other async way in other PR, such as channel in state machine thread, SYSTERM/SYSINT.

src/bin/nydusd/fusedev.rs
ccx1024cc: > Why is reaping fuse threads important? Only wait for state machine termination?

Yep, seeking for waiting state machine and fuse service thread termination separately. Because daemon status is only updated after successful action. For example, `interrupt` action is executed, then wait fuse service workers done, finally the daemon status is changed from `RUNNING` to `READY`.

src/bin/nydusd/daemon.rs
```diff
@@ -429,35 +454,28 @@ fn fs_backend_factory(cmd: &FsBackendMountCmd) -> DaemonResult<BackFileSystem> {
...
     derive(Debug, Clone)
     pub DaemonStateMachine(Init)
 
-    // FIXME: It's possible that failover does not succeed or resource is not capable to
-    // be passed. To handle event `Stop` when being `Init`.
     Init => {
-        Mount => Running [StartService],
-        Takeover => Upgrading [Restore],
-        Exit => Die[StopStateMachine],
+        Mount => Ready,
```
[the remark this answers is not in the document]
ccx1024cc: > Seems we no longer need `Mount` event since nothing is done when migrating state to Ready

In fact, we shall do `Mount` action. However, it's a bit risky now. Because the Mount content are different in different cases, such as upgrade startup, normal startup, endpoint startup. In total, You're right. There will be abstraction for `Mount` action in future.

bergwolf (approved):
LGTM!

PR #441, opened 2022-05-20
nydus-image: fix invalid meta_addr in rafs v6

liubogithub:
When building a rafs v6 image against a directory with only empty
files, it turns out that meta_addr is 0 which overlaps with the
superblock area.

This fixes the problem by setting meta_addr to EROFS_BLOCK_SIZE in the
above case.

This also makes sure that the root inode starts with RAFS_ROOT_INODE
to keep consistent with rafs v5.

Signed-off-by: Liu Bo <bo.liu@linux.alibaba.com>

### DIFF
--- a/src/bin/nydus-image/core/bootstrap.rs
+++ b/src/bin/nydus-image/core/bootstrap.rs
@@ -558,26 +558,24 @@ impl Bootstrap {
             (0, 0)
         };
 
-        // Make the superblock's meta_blkaddr one block ahead of the inode table,
-        // to avoid using 0 as root nid.
-        // inode offset = meta_blkaddr * block_size + 32 * nid
-        // When using nid 0 as root nid,
-        // the root directory will not be shown by glibc's getdents/readdir.
-        // Because in some OS, ino == 0 represents corresponding file is deleted.
-        let orig_meta_addr = bootstrap_ctx.nodes[0].offset - EROFS_BLOCK_SIZE;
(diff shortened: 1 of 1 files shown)


src/bin/nydus-image/core/bootstrap.rs
```diff
@@ -558,26 +558,24 @@ impl Bootstrap {
...
         } else {
-            orig_meta_addr
+            align_offset(orig_meta_addr, EROFS_BLOCK_SIZE as u64)
         };
 
         let root_nid = calculate_nid(
             bootstrap_ctx.nodes[0].offset - orig_meta_addr + meta_addr,
             meta_addr,
         );
+        debug_assert!(root_nid == RAFS_ROOT_INODE);
```
hsiangkao, line 578: I still don't quite get the point why we should use 1 for root inode since it's not a rule for rafs v6.
RAFS_ROOT_INODE should be renamed as RAFS_V5_ROOT_INODE

imeoer, line 578: Some users have encountered assert panic issue when using binary with debug build, but the release does work fine, should we still distinguish this inconsistent behavior?

liubogithub:
Turns out that it's okay to have meta_addr be zero. Am going to close the pr.

hsiangkao:
> Turns out that it's okay to have meta_addr be zero. Am going to close the pr.

Thanks for your understanding.

PR #442, opened 2022-05-25
chore: update tutorial.md

fatelei:
`containerd-nydus-grpc` has been removed from 2.x.x, so we should update tutorial.md for getting `containerd-nydus-grpc` from [nydus-snapshotter release](https://github.com/containerd/nydus-snapshotter/releases)


bergwolf (approved):
Thanks for spotting it!

PR #458, opened 2022-05-30
rafs/v6: fix wrongly calculated prefetch table size

changweige:
When specifying v6 prefetching files by --prefetch-polich=fs,
"/" indicates that all files should be prefetched in this nydus image.

The prefetch table size is defined as u32, but the count of prefetching
entries is wrongly calculated since the divisor is the size of u64 resulting
in no files being prefetch.

Moreover,  we should consider picking this patch into stable branches.


bergwolf (approved):
Nice catch!

PR #461, opened 2022-05-31
Tiny typo fix and improve nydusd help message

changweige:
Frankly speaking, `--shared-dir` is useless when nydusd works on FUSE mode. But we can still use this option to demonstrate that nydusd can work as FUSE passthroughfs and test the corresponding logics belong to `fuse-backend-rs`.

Add a more detailed help message telling why it is reserved.
In addition, fix a type name.

### DIFF
--- a/src/bin/nydusd/main.rs
+++ b/src/bin/nydusd/main.rs
@@ -210,7 +210,7 @@ fn append_fs_options(app: App<'static, 'static>) -> App<'static, 'static> {
         Arg::with_name("shared-dir")
             .long("shared-dir")
             .short("s")
-            .help("Directory shared by host and guest for passthroughfs, which also enables pathroughfs mode")
+            .help("Directory shared by host and guest for passthroughfs or a passthroughfs source for FUSE(for test and demonstration purpose), which also enables passthroughfs mode")
             .takes_value(true)
             .conflicts_with("bootstrap"),
     )
--- a/src/bin/nydusd/main.rs
+++ b/src/bin/nydusd/main.rs
@@ -195,6 +195,14 @@ extern "C" fn sig_exit(_sig: std::os::raw::c_int) {
     DAEMON_CONTROLLER.shutdown();
 }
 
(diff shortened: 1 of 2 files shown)


src/bin/nydusd/main.rs
```diff
@@ -210,7 +210,7 @@ fn append_fs_options(app: App<'static, 'static>) -> App<'static, 'static> {
         Arg::with_name("shared-dir")
             .long("shared-dir")
             .short("s")
-            .help("Directory shared by host and guest for passthroughfs, which also enables pathroughfs mode")
+            .help("Directory shared by host and guest for passthroughfs or a passthroughfs source for FUSE(for test and demonstration purpose), which also enables passthroughfs mode")
```
hsiangkao, line 213: should we a space and  (for the purpose of testing and demonstration)

changweige: Sure, leaving a white space makes it neater.

changweige: Then the help message looks like  below:
image

src/bin/nydusd/main.rs
```diff
@@ -210,7 +210,9 @@ fn append_fs_options(app: App<'static, 'static>) -> App<'static, 'static> {
         Arg::with_name("shared-dir")
             .long("shared-dir")
             .short("s")
-            .help("Directory shared by host and guest for passthroughfs, which also enables pathroughfs mode")
+            .help("Directory shared by host and guest for passthroughfs or a \
```
imeoer, line 213: In fact there are now two types of binary for nydusd, is it possible to distinguish here depending on whether the compilation feature is `fusedev` or `virtiofs`?

changweige: It sounds reasonable and possible. But I suppose there is a concrete plan to merge `virtiofs` and `fusedev` mode into a single nydusd,  thus making it easier to be deployed and configured,  isn't it?

PR #466, opened 2022-06-03
Improvements to the blob cache manager

jiangliu:
Several improvements to the nydusd blob cache manager.

### DIFF
--- a/src/bin/nydusd/blob_cache.rs
+++ b/src/bin/nydusd/blob_cache.rs
@@ -14,7 +14,15 @@ use storage::cache::FsCacheConfig;
 use storage::device::BlobInfo;
 use storage::factory::{BackendConfig, CacheConfig, FactoryConfig};
 
-const ID_SPLITTER: &str = "-";
+const ID_SPLITTER: &str = "/";
+
+fn generate_blob_key(domain_id: &str, blob_id: &str) -> String {
+    if domain_id.is_empty() {
+        blob_id.to_string()
+    } else {
+        format!("{}{}{}", domain_id, ID_SPLITTER, blob_id)
+    }
+}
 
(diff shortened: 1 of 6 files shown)


src/bin/nydusd/blob_cache.rs
```diff
@@ -277,28 +336,76 @@ impl BlobCacheMgr {
...
+            );
+            let key = data_blob.get_key().to_string();
+
+            if let Err(e) = state.try_add(data_blob) {
+                // Rollback added bootstrap/data blobs.
+                let id = BlobCacheObjectId {
+                    domain_id: domain_id.to_string(),
+                    blob_id: id.to_string(),
+                };
+                let _ = state.remove(&id);
```
imeoer, line 386: Do we need to rollback all added bootstrap/blob here?

src/bin/nydusd/blob_cache.rs
```diff
@@ -277,28 +336,76 @@ impl BlobCacheMgr {
...
+                return Err(e);
+            }
+
+            // Associate the data blob with the bootstrap blob.
+            if let Some(BlobCacheObjectConfig::DataBlob(entry)) = state.id_to_config_map.get(&key) {
```
imeoer, line 391: We can do `data_blob.clone()` before this to simplify the code.

imeoer:
Others LGTM!

hsiangkao:
@nan1994 please help test this with the latest fscache kernel, thanks!

src/bin/nydusd/blob_cache.rs
jiangliu: Thanks, fixed.

nan1994:
I have tested this with the latest fscache kernel and it works well.

Issue #425, opened 2022-05-12
add support for rafsv6 in nydusify


liubogithub:
At this moment, nydusify can only do conversion from OCIv1 image to rafsv5 image, conversion from OCIv1 image to rafsv6 image needs to be supported as well.

[acceld](https://github.com/goharbor/acceleration-service) provides the required functionality, so maybe we can share the same golang package with nydusify?

imeoer:
Supported in https://github.com/dragonflyoss/image-service/pull/473, thanks @zyfjeff  @yawqi

PR #473, opened 2022-06-07
nydusify & nydus-image: v6 image conversion support

yawqi:
We have adapted the nydusify and nydus-image to support using nydusify directly to build v6 image, to accomplish this purpose, we have introduce several features and fix several bugs.

Features:
1. Previously, the load_parent_bootstrap is unusable for v6 image, we have adapted it for both V5 and V6.
2. We have implemented all methods for v6's OndiskInodeWrapper.

Bug fixes:
1. The dirents_offset will only be used by dir and symlink files, for other files, calculating it may leads to overflow.
2. Correct chunk info implementation for v6.
3. Previous, we assume the . and .. are the first and second dirents, which is wrong.
4. Under some circumstances, the xattr size is wrong.

FYI, we have tested converting all top images in the image_list.txt and checked their bootstraps with fsck.erofs.
#425

### DIFF
--- a/contrib/nydusify/cmd/nydusify.go
+++ b/contrib/nydusify/cmd/nydusify.go
@@ -210,7 +210,7 @@ func main() {
 				// chosen to make it compatible with the 127 max in graph driver of
 				// docker so that we can pull cache image using docker.
 				&cli.UintFlag{Name: "build-cache-max-records", Value: maxCacheMaxRecords, Usage: "Maximum cache records in cache image", EnvVars: []string{"BUILD_CACHE_MAX_RECORDS"}},
-				&cli.StringFlag{Name: "image-version", Required: false, Usage: "Image format version", EnvVars: []string{"IMAGE_VERSION"}, Value: "5", DefaultText: "V5 format"},
+				&cli.StringFlag{Name: "fs-version", Required: false, Usage: "Image format version", EnvVars: []string{"FS_VERSION"}, Value: "5", DefaultText: "V5 format"},
 			},
 			Action: func(c *cli.Context) error {
 				setupLogLevel(c)
@@ -312,7 +312,7 @@ func main() {
 
 					NydusifyVersion: version,
 					Source:          c.String("source"),
-					ImageVersion:    c.String("image-version"),
+					FsVersion:       c.String("fs-version"),
 
 					ChunkDict: converter.ChunkDictOpt{
 						Args:     c.String("chunk-dict"),
--- a/contrib/nydusify/cmd/nydusify.go
+++ b/contrib/nydusify/cmd/nydusify.go
@@ -210,6 +210,7 @@ func main() {
 				// chosen to make it compatible with the 127 max in graph driver of
 				// docker so that we can pull cache image using docker.
 				&cli.UintFlag{Name: "build-cache-max-records", Value: maxCacheMaxRecords, Usage: "Maximum cache records in cache image", EnvVars: []string{"BUILD_CACHE_MAX_RECORDS"}},
+				&cli.StringFlag{Name: "image-version", Required: false, Usage: "Image format version", EnvVars: []string{"IMAGE_VERSION"}, Value: "5", DefaultText: "V5 format"},
 			},
 			Action: func(c *cli.Context) error {
 				setupLogLevel(c)
@@ -311,6 +312,7 @@ func main() {
 
 					NydusifyVersion: version,
 					Source:          c.String("source"),
+					ImageVersion:    c.String("image-version"),
 
 					ChunkDict: converter.ChunkDictOpt{
 						Args:     c.String("chunk-dict"),
--- a/contrib/nydusify/cmd/nydusify.go
+++ b/contrib/nydusify/cmd/nydusify.go
@@ -210,7 +210,7 @@ func main() {
 				// chosen to make it compatible with the 127 max in graph driver of
 				// docker so that we can pull cache image using docker.
 				&cli.UintFlag{Name: "build-cache-max-records", Value: maxCacheMaxRecords, Usage: "Maximum cache records in cache image", EnvVars: []string{"BUILD_CACHE_MAX_RECORDS"}},
-				&cli.StringFlag{Name: "fs-version", Required: false, Usage: "Image format version", EnvVars: []string{"FS_VERSION"}, Value: "5", DefaultText: "V5 format"},
+				&cli.StringFlag{Name: "fs-version", Required: false, Usage: "Version number of nydus image format, possible values: 5, 6", EnvVars: []string{"FS_VERSION"}, Value: "5", DefaultText: "V5 format"},
 			},
 			Action: func(c *cli.Context) error {
 				setupLogLevel(c)
--- a/contrib/nydusify/tests/image_test.go
+++ b/contrib/nydusify/tests/image_test.go
@@ -34,11 +34,11 @@ func transfer(t *testing.T, ref string) {
 	run(t, fmt.Sprintf("docker push %s/%s", host, ref), true)
 }
 
-func convert(t *testing.T, ref string, imageVersion string) {
+func convert(t *testing.T, ref string, fsVersion string) {
 	registry := NewRegistry(t)
 	defer registry.Destroy(t)
 	transfer(t, ref)
-	nydusify := NewNydusify(registry, ref, fmt.Sprintf("%s-nydus", ref), "", "", imageVersion)
+	nydusify := NewNydusify(registry, ref, fmt.Sprintf("%s-nydus", ref), "", "", fsVersion)
 	nydusify.Convert(t)
 	nydusify.Check(t)
 }
--- a/contrib/nydusify/tests/image_test.go
+++ b/contrib/nydusify/tests/image_test.go
@@ -34,17 +34,18 @@ func transfer(t *testing.T, ref string) {
 	run(t, fmt.Sprintf("docker push %s/%s", host, ref), true)
 }
 
-func convert(t *testing.T, ref string) {
+func convert(t *testing.T, ref string, imageVersion string) {
 	registry := NewRegistry(t)
 	defer registry.Destroy(t)
 	transfer(t, ref)
-	nydusify := NewNydusify(registry, ref, fmt.Sprintf("%s-nydus", ref), "", "")
+	nydusify := NewNydusify(registry, ref, fmt.Sprintf("%s-nydus", ref), "", "", imageVersion)
 	nydusify.Convert(t)
 	nydusify.Check(t)
 }
 
 func TestDockerHubImage(t *testing.T) {
 	for _, ref := range list {
-		convert(t, ref)
+		convert(t, ref, "5")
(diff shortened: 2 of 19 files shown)


rafs/src/metadata/direct_v6.rs
```diff
@@ -179,9 +180,29 @@ impl DirectSuperBlockV6 {
...
+        nid: u64,
+        parent_inode: Inode,
+        name: OsString,
+    ) -> Result<Arc<OndiskInodeWrapper>> {
+        self.inode_wrapper(nid).map(|inode| {
+            let mut inode = inode;
+            // # Safety
+            // inode always valid
+            Arc::get_mut(&mut inode).unwrap().parent_inode = Cell::new(Some(parent_inode));
+            Arc::get_mut(&mut inode).unwrap().name = Cell::new(Some(name));
```
liubogithub, line 210: Since `Cell` is not thread safe,  I suppose that it's not okay to assign `Cell` to a Arc which is a thread safe.

rafs/src/metadata/direct_v6.rs
```diff
@@ -873,15 +1008,42 @@ impl RafsInode for OndiskInodeWrapper {
...
-
+    // RafsV5 flags, not used by v6, return 0
     fn flags(&self) -> u64 {
-        todo!()
+        0
     }
 
     fn get_digest(&self) -> RafsDigest {
-        todo!()
+        RafsDigest { data: [0u8; 32] }
```
liubogithub, line 1106: RafsDigest has implemented `Default` trait, maybe we can use it directly.

pigletDips: Done.

contrib/nydusify/cmd/nydusify.go
```diff
@@ -210,6 +210,7 @@ func main() {
 				// chosen to make it compatible with the 127 max in graph driver of
 				// docker so that we can pull cache image using docker.
 				&cli.UintFlag{Name: "build-cache-max-records", Value: maxCacheMaxRecords, Usage: "Maximum cache records in cache image", EnvVars: []string{"BUILD_CACHE_MAX_RECORDS"}},
+				&cli.StringFlag{Name: "image-version", Required: false, Usage: "Image format version", EnvVars: []string{"IMAGE_VERSION"}, Value: "5", DefaultText: "V5 format"},
```
imeoer, line 213: Better named `fs-version`, consistent with nydus-image.

pigletDips: Done.

rafs/src/metadata/direct_v6.rs
```diff
@@ -30,6 +32,7 @@ use std::slice;
 use std::sync::Arc;
 
 use arc_swap::{ArcSwap, Guard};
+use std::cell::RefCell;
```
jiangliu, line 35: This line could be merged with line 20 above.

rafs/src/metadata/direct_v6.rs
```diff
@@ -61,6 +64,12 @@ use storage::device::{
 };
 use storage::utils::readahead;
 
+// Use to store chunk info pre inode, Oour build is actually single-threaded,
```
jiangliu, line 67: Use to store chunk info per inode, Our build is actually single-threaded

rafs/src/metadata/direct_v6.rs
```diff
@@ -179,9 +189,29 @@ impl DirectSuperBlockV6 {
...
+        &self,
+        nid: u64,
+        parent_inode: Inode,
+        name: OsString,
+    ) -> Result<Arc<OndiskInodeWrapper>> {
+        self.inode_wrapper(nid).map(|inode| {
+            let mut inode = inode;
+            // # Safety
+            // inode always valid
+            Arc::get_mut(&mut inode).unwrap().parent_inode = Cell::new(Some(parent_inode));
```
jiangliu, line 209: It's unsafe to use Arc::get_mut().unwrap() here, it may cause panic if the strong count is bigger than 1. Need more reliable mechanism here.

rafs/src/metadata/direct_v6.rs
zyfjeff: fixed

rafs/src/metadata/direct_v6.rs
zyfjeff: fixed

rafs/src/metadata/direct_v6.rs
zyfjeff: fixed

zyfjeff:
@liubogithub @jiangliu @imeoer PTAL

contrib/nydusify/tests/image_test.go
```diff
@@ -34,17 +34,18 @@ func transfer(t *testing.T, ref string) {
...
+	nydusify := NewNydusify(registry, ref, fmt.Sprintf("%s-nydus", ref), "", "", fsVersion)
 	nydusify.Convert(t)
 	nydusify.Check(t)
 }
 
 func TestDockerHubImage(t *testing.T) {
 	for _, ref := range list {
-		convert(t, ref)
+		convert(t, ref, "5")
+		convert(t, ref, "6")
```
imeoer, line 49: Can we make them execute concurrently to improve the speed of CI execution?

contrib/nydusify/cmd/nydusify.go
```diff
@@ -210,6 +210,7 @@ func main() {
 				// chosen to make it compatible with the 127 max in graph driver of
 				// docker so that we can pull cache image using docker.
 				&cli.UintFlag{Name: "build-cache-max-records", Value: maxCacheMaxRecords, Usage: "Maximum cache records in cache image", EnvVars: []string{"BUILD_CACHE_MAX_RECORDS"}},
+				&cli.StringFlag{Name: "fs-version", Required: false, Usage: "Image format version", EnvVars: []string{"FS_VERSION"}, Value: "5", DefaultText: "V5 format"},
```
imeoer, line 213: `Usage: "Version number of nydus image format, possible values: `5`, `6`"`

zyfjeff: fixed

src/bin/nydus-image/builder/directory.rs
```diff
@@ -124,13 +124,17 @@ impl Builder for DirectoryBuilder {
...
         let mut bootstrap = Bootstrap::new()?;
         if bootstrap_ctx.layered {
             // Merge with lower layer if there's one, do not prepare `prefetch` list during merging.
             ctx.prefetch.disable();
             bootstrap.build(ctx, &mut bootstrap_ctx, &mut tree)?;
             tree = bootstrap.apply(ctx, &mut bootstrap_ctx, bootstrap_mgr, blob_mgr, None)?;
         }
+        // If layered, the bootstrap_ctx.offset will be set in first build, so we need restore it here
+        bootstrap_ctx.offset = origin_bootstarp_offset;
+        bootstrap_ctx.layered = false;
```
liubogithub, line 137: This also affects https://github.com/dragonflyoss/image-service/blob/fdee5992103916910dea755a2df3d5d669ea6efb/src/bin/nydus-image/core/bootstrap.rs#L201

where `layered` is used to check whiteout, have you checked if setting `layered = false` would cause any difference?

liubogithub, line 137: This also affects https://github.com/dragonflyoss/image-service/blob/fdee5992103916910dea755a2df3d5d669ea6efb/src/bin/nydus-image/core/bootstrap.rs#L201

where `layered` is used to check whiteout, have you checked if setting `layered = false` would cause any difference?

contrib/nydusify/cmd/nydusify.go
liubogithub: looks like the changes are not updated, @yawqi

tests/smoke.rs
```diff
@@ -109,7 +109,7 @@ fn test(
             cache_compressed,
             rafs_mode.parse().unwrap(),
             "api.sock".into(),
-            true,
+            !rafsv6,
```
liubogithub, line 112: Can you please leave a comment either here or in the patch log for why digest validate for v6 is disabled.

src/bin/nydus-image/inspect.rs
```diff
@@ -141,20 +181,19 @@ impl RafsInspector {
 
     fn load_meta(f: &mut RafsIoReader) -> Result<(RafsMeta, RafsLayout)> {
         let layout_profile = RafsLayout::rafsv5_layout();
-        match Self::super_block_v5(f, &layout_profile) {
-            Ok(sb) => {
+        let mut res = Self::super_block_v5(f, &layout_profile).map(|sb| {
```
liubogithub, line 184: `super_block_v5` does `RafsV5SuperBlock::load()` which doesn't check fs magic or fs version, so it won't return errors if it's a v6 superblock.

zyfjeff: I modified the super_block_v5 and added a checker for rafs versions.

```rust
      if !sb.detect() {
            return Err(anyhow!("invalid rafs v5 version"));
        }
```

tests/smoke.rs
zyfjeff: Actually already has some comments. exists only where it was first used.
```
            // FIXME: Currently no digest validation is implemented for rafs v6.
            !rafsv6,
```

src/bin/nydus-image/builder/directory.rs
zyfjeff: In fact, the processing of whiteout is already done at the time of the first bootstarp.build, and here is actually the second time to do bootstarp.build

rafs/src/metadata/direct_v6.rs
```diff
@@ -821,25 +819,27 @@ impl RafsInode for OndiskInodeWrapper {
...
-            let nid = de.e_nid;
-            return Ok(self.mapping.inode_wrapper_with_info(
-                nid,
-                self.ino(),
-                OsString::from(d_name),
-            )? as Arc<dyn RafsInode>);
+            for j in 0..entries_count {
+                let de = self
+                    .get_entry(i as usize, j as usize)
+                    .map_err(err_invalidate_data)?;
```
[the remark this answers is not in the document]
zyfjeff: Yup

zyfjeff:
@liubogithub PTAL

rafs/src/metadata/direct_v6.rs
```diff
@@ -873,15 +1069,41 @@ impl RafsInode for OndiskInodeWrapper {
     /// # Safety
     /// It depends on Self::validate() to ensure valid memory layout.
     fn name(&self) -> OsString {
-        todo!()
+        if let Some(name) = self.name.take() {
```
jiangliu, line 1072: This could be simplified?:
```
match self.name.clone() {
    Some(v) => v,
    None => {
          xxxxx
    }
}
```

zyfjeff: fixed

imeoer (approved):

PR #489, opened 2022-06-13
nydus-image: supports multiple versions of merge

zyfjeff:
Let the merge subcommand support multiple versions of image merging

### DIFF
--- a/src/bin/nydus-image/core/context.rs
+++ b/src/bin/nydus-image/core/context.rs
@@ -24,6 +24,7 @@ use nydus_utils::{compress, digest, div_round_up, round_down_4k};
 use rafs::metadata::layout::v5::RafsV5BlobTable;
 use rafs::metadata::layout::v6::{RafsV6BlobTable, EROFS_BLOCK_SIZE, EROFS_INODE_SLOT_SIZE};
 use rafs::metadata::layout::RafsBlobTable;
+use rafs::metadata::layout::{RAFS_SUPER_VERSION_V5, RAFS_SUPER_VERSION_V6};
 use rafs::metadata::RafsSuperFlags;
 use rafs::metadata::{Inode, RAFS_DEFAULT_CHUNK_SIZE, RAFS_MAX_CHUNK_SIZE};
 use rafs::{RafsIoReader, RafsIoWrite};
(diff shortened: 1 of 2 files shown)


src/bin/nydus-image/merge.rs
```diff
@@ -166,6 +171,10 @@ impl Merger {
             }
         }
 
+        assert!(fs_versions.len() == 1);
```
imeoer, line 174: How about recording the first fs version in the `for` loop? so that we can throw a error like this:

`bail!("inconsistent fs version between layers, expected version {}", expected);`

zyfjeff: fixed

src/bin/nydus-image/core/context.rs
```diff
@@ -24,6 +24,7 @@ use nydus_utils::{compress, digest, div_round_up, round_down_4k};
 use rafs::metadata::layout::v5::RafsV5BlobTable;
 use rafs::metadata::layout::v6::{RafsV6BlobTable, EROFS_BLOCK_SIZE, EROFS_INODE_SLOT_SIZE};
 use rafs::metadata::layout::RafsBlobTable;
+use rafs::metadata::layout::{RAFS_SUPER_VERSION_V5, RAFS_SUPER_VERSION_V6};
```
jiangliu, line 27: This line could be merged into the line above.

src/bin/nydus-image/core/context.rs
```diff
@@ -51,6 +52,17 @@ impl Default for RafsVersion {
     }
 }
 
+impl From<u32> for RafsVersion {
```
jiangliu, line 55: Should we implement TryFrom instead of From so it could return errors.

imeoer (approved):
LGTM! Waiting to merge.

src/bin/nydus-image/core/context.rs
zyfjeff: fixed

src/bin/nydus-image/core/context.rs
zyfjeff: fixed

imeoer (approved):

PR #494, opened 2022-06-14
nydusify: bug fix, `check` subcommand uses direct mode

yawqi:
The cache mode is used to speed up. However, when checking, especially the RAFS v6 haven't support the cache mode,
we should directly use direct mode. 

What's more, the `--fs-version` hasn't been add to nydusify check's arg.

### DIFF
--- a/contrib/nydusify/pkg/checker/checker.go
+++ b/contrib/nydusify/pkg/checker/checker.go
@@ -32,7 +32,6 @@ type Opt struct {
 	BackendType    string
 	BackendConfig  string
 	ExpectedArch   string
-	FsVersion      string
 }
 
 // Checker validates Nydus image manifest, bootstrap and mounts filesystem
@@ -106,13 +105,8 @@ func (checker *Checker) Check(ctx context.Context) error {
 		return errors.Wrap(err, "output image information")
 	}
 
-	mode := "cached"
-	digestValidate := true
-	if checker.FsVersion == "6" {
-		mode = "direct"
-		digestValidate = false
-
-	}
+	mode := "direct"
+	digestValidate := false
 
 	rules := []rule.Rule{
 		&rule.ManifestRule{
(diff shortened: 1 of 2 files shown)


imeoer (approved):
LGTM

liubogithub (commented):
What bug is this patch trying fix?

yawqi:
> What bug is this patch trying fix?

The `nydusify check` subcommand doesn't offer a `--fs-version` option, so for both v5 and v6, the fs-version is the default 5, therefore, it will always use cache mode for both v5 and v6, which is not implemented for v6.
However the direct mode is implemented for both v5 and v6.

jiangliu:
> > What bug is this patch trying fix?
> 
> The `nydusify check` subcommand doesn't offer a `--fs-version` option, so for both v5 and v6, the fs-version is the default 5, therefore, it will always use cache mode for both v5 and v6, which is not implemented for v6. However the direct mode is implemented for both v5 and v6.

Could we enable "--fs-version" for `nydusfy check` command?

yawqi:
> > > What bug is this patch trying fix?
> > 
> > 
> > The `nydusify check` subcommand doesn't offer a `--fs-version` option, so for both v5 and v6, the fs-version is the default 5, therefore, it will always use cache mode for both v5 and v6, which is not implemented for v6. However the direct mode is implemented for both v5 and v6.
> 
> Could we enable "--fs-version" for `nydusfy check` command?

I am OK with that, but `nydusify check` actually is able to detect the fs-version I think. The fs-version is only used here actually, do we need to add it?

jiangliu:
> > > > What bug is this patch trying fix?
> > > 
> > > 
> > > The `nydusify check` subcommand doesn't offer a `--fs-version` option, so for both v5 and v6, the fs-version is the default 5, therefore, it will always use cache mode for both v5 and v6, which is not implemented for v6. However the direct mode is implemented for both v5 and v6.
> > 
> > 
> > Could we enable "--fs-version" for `nydusfy check` command?
> 
> I am OK with that, but `nydusify check` actually is able to detect the fs-version I think. The fs-version is only used here actually, do we need to add it?

@imeoer any suggestion?

imeoer:
> @imeoer any suggestion?

nydusify check is able to detect `fs-version` of the image (via bootstrap magic number or annotation in image manifest), so we don't need a flag.

jiangliu (approved):

PR #495, opened 2022-06-14
nydus-image: fixed a bug caused by empty blob and prefetched combined scenes

zyfjeff:
When building a layer independently, if the layer does not produce any blobs, but the prefetch is set,
the position of the prefetch table and the position of the inode table will overlap,
resulting in the prefetch table overwriting part of the inode table.


jiangliu (approved):

src/bin/nydus-image/core/blob.rs
```diff
@@ -36,16 +36,23 @@ impl Blob {
...
+                        has_chunk = true
+                    }
                     if idx < prefetch_entries {
                         blob_ctx.blob_readahead_size += size;
                     }
                 }
-                self.dump_meta_data(blob_ctx)?;
+                // Dump is only required if there is chunk in the blob
+                if has_chunk {
+                    self.dump_meta_data(blob_ctx)?;
```
imeoer, line 54: Better to check `blob_ctx.compressed_blob_size` or `blob_ctx.chunk_count` in `dump_meta_data`, so that we can reuse this logic for stargz build in future.

tests/builder.rs
```diff
@@ -281,6 +285,28 @@ impl<'a> Builder<'a> {
             )
             .as_str(),
             false,
+            b"",
+        ).unwrap();
+    }
+
+    pub fn build_empty_file_with_prefetch(&mut self) {
```
imeoer, line 292: Why not follow `build_lower` and `build_upper` (create new one named `builder_empty`)? so that we can reuse the `test` func in `tests/smoke.rs`.

zyfjeff: Bugs only occur when prefetched and empty files are combined, so there are no build_lower and build_upper used here

src/bin/nydus-image/core/blob.rs
zyfjeff: fixed

tests/builder.rs
zyfjeff: This case and other tests have now been combined.

src/bin/nydus-image/core/blob.rs
```diff
@@ -45,7 +45,10 @@ impl Blob {
                         blob_ctx.blob_readahead_size += size;
                     }
                 }
-                self.dump_meta_data(blob_ctx)?;
+                // Dump is only required if there is chunk in the blob
+                if blob_ctx.compressed_blob_size > 0 {
+                    self.dump_meta_data(blob_ctx)?;
```
imeoer, line 50: Move the `if blob_ctx.compressed_blob_size > 0` check into `dump_meta_data` should be better.

tests/smoke.rs
```diff
@@ -88,7 +106,6 @@ fn test(
             cache_compressed,
             rafs_mode.parse().unwrap(),
             "api.sock".into(),
-            // FIXME: Currently no digest validation is implemented for rafs v6.
```
imeoer, line 91: Why remove this comment?

zyfjeff: It was not deleted, just moved it to where it first appeared.

src/bin/nydus-image/core/blob.rs
zyfjeff: fixed

PR #497, opened 2022-06-15
nydusify: add fs-version annotation

### LINKED ISSUE #496: nydusify: append fs version annotation in image manifest
We need to add an annotation to the image manifest for detecting the bootstrap fs version without reading the magic number from bootstrap, example:

```
"containerd.io/snapshot/nydus-bootstrap": "true",
"containerd.io/snapshot/nydus-fs-version": "6",
...
```

yawqi:
Add an annotation to the image manifest for detecting the bootstrap fs
version without reading the magic number from bootstrap.

fix #496

### DIFF
--- a/contrib/nydusify/pkg/cache/cache.go
+++ b/contrib/nydusify/pkg/cache/cache.go
@@ -29,7 +29,16 @@ import (
 type Opt struct {
 	// Maximum records(bootstrap layer + blob layer) in cache image.
 	MaxRecords uint
-	Version    string
+	// Version of cache image, we need to discard cache layers when
+	// the required version (specified by `--build-cache-version`)
+	// is unmatched with the cache image version, for example nydus
+	// bootstrap format has a minor upgrade.
+	Version string
+	// Bootstrap's RAFS version of cache image, we need to discard cache
+	// layers when the required version (specified by `--fs-version`) is
+	// unmatched with the fs version recorded in cache image, for example
+	// we can't use rafs v5 cache layers for rafs v6 image.
+	FsVersion string
 	// Make cache image manifest compatible with the docker v2 media
 	// type defined in github.com/containerd/containerd/images.
 	DockerV2Format bool
@@ -129,6 +138,7 @@ func (cache *Cache) recordToLayer(record *Record) (*ocispec.Descriptor, *ocispec
 		Size:      record.NydusBootstrapDesc.Size,
 		Annotations: map[string]string{
 			utils.LayerAnnotationNydusBootstrap:     "true",
+			utils.LayerAnnotationNydusFsVersion:     cache.opt.FsVersion,
 			utils.LayerAnnotationNydusSourceChainID: record.SourceChainID.String(),
 			// Use the annotation to record bootstrap layer DiffID.
 			utils.LayerAnnotationUncompressed: record.NydusBootstrapDiffID.String(),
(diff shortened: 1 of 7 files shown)


contrib/nydusify/pkg/cache/cache.go
```diff
@@ -129,6 +129,7 @@ func (cache *Cache) recordToLayer(record *Record) (*ocispec.Descriptor, *ocispec
 		Size:      record.NydusBootstrapDesc.Size,
 		Annotations: map[string]string{
 			utils.LayerAnnotationNydusBootstrap:     "true",
+			utils.LayerAnnotationNydusFsVersion:     cache.opt.Version,
```
imeoer, line 132: Should be `cache.opt.FsVersion`, how this option be passed to the cache image?

yawqi: I checked, it is cache.opt.Version.

yawqi:
```
// Opt configures Nydus cache
type Opt struct {
	// Maximum records(bootstrap layer + blob layer) in cache image.
	MaxRecords uint
	Version    string
	// Make cache image manifest compatible with the docker v2 media
	// type defined in github.com/containerd/containerd/images.
	DockerV2Format bool
	// The blob layer record will not be written to cache image if
	// the backend be specified, because the blob layer will be uploaded
	// to backend.
	Backend backend.Backend
}
```

imeoer: @yawqi  This `Version` field is used to indicate the version of cache image (not fs version), nydusify will use cache records depending on the version. :)

imeoer: We can add some comments to distinguish these two fields:

``` go
// Opt configures Nydus cache
type Opt struct {
	...
	// For the version of cache image, we need to discard cache layers
	// when the required version (specified by `--build-cache-version`)
	// is unmatched with the cache image version, for example nydus
	// bootstrap format has a minor upgrade.
	Version    string
	// Bootstrap's RAFS version of cache image, we need to discard cache
	// layers when the required version (specified by `--fs-version`) is
	// unmatched with the fs version recorded in cache image, for example
	// we can't use rafs v5 cache layers for rafs v6 image.
	FsVersion string
	// Make cache image manifest compatible with the docker v2 media
	// type defined in github.com/containerd/containerd/images.
	...
}
```

yawqi: Done.

contrib/nydusify/pkg/cache/cache.go
```diff
@@ -443,14 +455,22 @@ func (cache *Cache) Import(ctx context.Context) error {
...
 		return fmt.Errorf(
 			"unmatched cache image version %s, required to be %s",
 			manifest.Annotations[utils.ManifestNydusCache], cache.opt.Version,
 		)
 	}
 
+	// Discard the cache if mismatched RAFS FsVersion
+	if manifest.Annotations[utils.LayerAnnotationNydusFsVersion] != cache.opt.FsVersion {
+		return fmt.Errorf(
+			"unmatched cache image version %s, required to be %s",
```
imeoer, line 469: Maybe `unmatched fs version %s of cache image, required to be %s`.

contrib/nydusify/pkg/cache/cache.go
```diff
@@ -443,14 +455,22 @@ func (cache *Cache) Import(ctx context.Context) error {
...
+	// Discard the cache if mismatched version
 	if manifest.Annotations[utils.ManifestNydusCache] != cache.opt.Version {
 		return fmt.Errorf(
 			"unmatched cache image version %s, required to be %s",
 			manifest.Annotations[utils.ManifestNydusCache], cache.opt.Version,
 		)
 	}
 
+	// Discard the cache if mismatched RAFS FsVersion
+	if manifest.Annotations[utils.LayerAnnotationNydusFsVersion] != cache.opt.FsVersion {
```
imeoer, line 467: For `manifest.Annotations[utils.LayerAnnotationNydusFsVersion] == ""` (e.g. old cache image), we should consider it as `5`.

imeoer (approved):
LGTM!

PR #499, opened 2022-06-16
doc: update expired dingtalk qr code

imeoer:
Signed-off-by: Yan Song <yansong.ys@antfin.com>


hsiangkao:
I think we just need to update the QR code, the validity period is one year.

hsiangkao (approved):

PR #501, opened 2022-06-17
nydusify: add command line arg fs-version check

yawqi:
Add fs-version value check, should be either "5" or "6".

### DIFF
--- a/contrib/nydusify/cmd/nydusify.go
+++ b/contrib/nydusify/cmd/nydusify.go
@@ -226,6 +226,12 @@ func main() {
 					return fmt.Errorf("--backend-type should be one of %v", possibleBackendTypes)
 				}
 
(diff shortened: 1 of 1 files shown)


contrib/nydusify/cmd/nydusify.go
```diff
@@ -226,6 +226,12 @@ func main() {
 					return fmt.Errorf("--backend-type should be one of %v", possibleBackendTypes)
 				}
 
+				fsVersion := c.String("fs-version")
```
imeoer, line 229: Perhaps we can reuse this variable below.

yawqi: Done.

imeoer (approved):
LGTM. Thanks!

PR #545, opened 2022-07-01
makefile: add target `install`

changweige:
We don't provide software package yet. In order to
let users easily install nydus tool chain onto machine.
We'd better provide `make install` to automatically
copy necessary binaris to run nydus-snapshotter, thus to
pull and run nydus image.

Moreover, nydus-snapshotter had already provide `make install`
helper.

Signed-off-by: Changewei Ge <gechangwei@bytedance.com>


yqleng1987:
@changweige , a new test job has been submitted. Please wait in patience.

jiangliu (approved):

PR #556, opened 2022-07-04
backend-proxy: prepare nydus backend proxy fd map reload

changweige:
Current code is hard to implement blobs fd map reload since blobs manager has to be mutable among different futures.
At present `nydus-backend-proxy` only loads blobs info when it starts by `readdir()`. So blob which is generated and stored in the `blobsdir` can't be found responding to the HTTP requests.

This PR refactors the proxy a bit and prepare to reload blobs, it will benefit nydus-test.


jiangliu:
> nydus-backend-proxy has no file header license declaration, should we add it? @bergwolf @liubin @jiangliu

Please add license claim for it:)

yqleng1987:
@changweige , your pull request has been updated. A new test job will be submitted. Please wait in patience.

yqleng1987:
@changweige , your test job has passed, and no need to test again.

changweige:
I already ran nydus-test using this nydus-banckend-proxy as storage backend without any data corruption failure.
So it is tested.

hsiangkao (approved):

contrib/nydus-backend-proxy/src/main.rs
```diff
@@ -1,250 +1,275 @@
...
+    *BLOB_BACKEND.lock().await = b;
+}
+
+#[derive(Debug)]
+struct BlobBackend {
+    root: PathBuf,
+    blobs: HashMap<String, fs::File>,
+}
+impl BlobBackend {
+    fn populate_blobs_map(&mut self) {
```
imeoer, line 58: Is it necessary to load the whole blob directory if the blob is not found on fetching?

changweige: My thought is to load all new files into fd map after re-scanning since the `nydusd-backend-proxy` 's  exports all the files  int `blobsdir` , which I think is its SLO

PR #573, opened 2022-07-07
nydusd: fixed getting the wrong chunk due to using the wrong blob index

zyfjeff:
The blob index on the disk where erofs exists starts at 1, but the blob index in BlobInfo starts at 0,
and the blob index on the disk is mistakenly used here instead of blobInfo's index
this will eventually result in the wrong file contents being obtained.

### DIFF
--- a/rafs/src/metadata/direct_v6.rs
+++ b/rafs/src/metadata/direct_v6.rs
@@ -1239,7 +1239,7 @@ impl RafsInode for OndiskInodeWrapper {
                 content_len = std::cmp::min(chunk_size, left);
                 let desc = self.make_chunk_io(c, 0, content_len, user_io);
 
-                if c.blob_index() as u32 != descs.bi_vec[0].blob.blob_index() {
+                if desc.blob.blob_index() as u32 != descs.bi_vec[0].blob.blob_index() {
                     vec.push(descs);
                     descs = BlobIoVec::new();
                 }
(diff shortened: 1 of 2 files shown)


yqleng1987:
@zyfjeff , a new test job has been submitted. Please wait in patience.

imeoer (approved):

jiangliu (approved):

hsiangkao:
could you leave some words on what it will impact?

zyfjeff:
> 

You're right, I'll change it.

zyfjeff:
> Why does this patch relate to nydus-image? It looks like a change of nydusd runtime

You're right, I'll change it.

yqleng1987:
@zyfjeff , your pull request has been updated. A new test job will be submitted. Please wait in patience.

yqleng1987:
@zyfjeff , your test job has passed, and no need to test again.

zyfjeff:
> could you leave some words on what it will impact? it impact chunk info?

This will cause the contents of the file to be confused, or the chunk index to go out of bounds

hsiangkao (approved):

changweige:
I don't quite understand why data is corrupted. I think it can only cause wrongly generating `descs` which is a set of backend chunk IO, which means backend IO merging is low-efficient.

zyfjeff:
> 



> I don't quite understand why data is corrupted. I think it can only cause wrongly generating `descs` which is a set of backend chunk IO, which means backend IO merging is low-efficient.

This will result in a BlobIoVec containing different Blobs, but where BlobIoVec is used, it is assumed that they all belong to the same Blob, but it is not, which will result in chunks being taken from the wrong blob

changweige (approved):

PR #606, opened 2022-07-22
make copen always return right file size

kevinXYin:


src/bin/nydusd/fs_cache.rs
```diff
@@ -433,22 +450,49 @@ impl FsCacheHandler {
...
+            assert!(guard.get_blobcache().is_none());
+            for _ in 0..5 {
+                match Self::create_data_blob_object(
+                    &guard.config,
+                    guard.file.clone(),
+                    guard.blobs_need,
+                ) {
+                    Err(e) => {
+                        warn!("fscache: create_data_blob_object failed {}", e);
+                        thread::sleep(time::Duration::from_millis(300));
```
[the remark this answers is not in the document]
hsiangkao: perhaps define a default delay interval?

hsiangkao:
Please resolve DCO checks @kevinXYin 
PTL @jiangliu @changweige @imeoer 

It's important to fscache backend.

src/bin/nydusd/fs_cache.rs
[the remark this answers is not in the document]
kevinXYin: fixed

hsiangkao (approved):

src/bin/nydusd/fs_cache.rs
```diff
@@ -642,19 +683,28 @@ impl FsCacheHandler {
...
+                                Err(e) => error!(
+                                    "{}",
+                                    format!("fscache: failed to read data from blob object: {}", e,)
+                                ),
+                            },
+                        }
+                    }
+                    _ => {
+                        warn!("fscache: blob object not ready")
+                        //TODO: maybe we should retry init blob object here
```
jiangliu, line 706: It would be better to add a flag to mark the init_blobcache() is still working or has failed out. If it's still working, we should wait for a while.

hsiangkao: Since Xin Yin is currently busy this week, could we merge this in advance and apply a follow-up fix later if possible?

storage/src/factory.rs
```diff
@@ -112,8 +112,9 @@ impl BlobFactory {
         let key = BlobCacheMgrKey {
             config: config.clone(),
         };
+        let mut guard = self.mgrs.lock().unwrap();
```
changweige, line 115: Looks not necessary to hold the mutex while creating each cache object where blob.meta is being downloaded.
I prefer the original style

kevinXYin: for the original style, if we call new_blob_cache() with same config simultaneously , we will create multiple BlobCacheMgr

jiangliu (approved):

PR #609, opened 2022-07-25
Reduce network IOs by amplifying prefetch requests

jiangliu:
Reduce backend network IOs by enabling amplification of v6 prefetch requests. For a java image, it could reduce 50% network requests by average.

截屏2022-07-25 下午7 51 12

### DIFF
--- a/storage/src/cache/cachedfile.rs
+++ b/storage/src/cache/cachedfile.rs
@@ -358,6 +358,15 @@ impl BlobObject for FileCacheEntry {
             }
         }
 
+        if range.blob_size < RAFS_DEFAULT_CHUNK_SIZE {
+            let max_size = RAFS_DEFAULT_CHUNK_SIZE - range.blob_size;
+            if let Some(meta) = self.meta.as_ref() {
+                if let Some(chunks) = meta.add_more_chunks(chunks, max_size) {
+                    return self.do_fetch_chunks(&chunks);
+                }
+            }
+        }
+
         self.do_fetch_chunks(chunks)
     }
 }
@@ -631,7 +640,8 @@ impl FileCacheEntry {
         }
 
         let blob_size = region.blob_len as usize;
-        debug!("total backend data {}KB", blob_size / 1024);
+        debug!("try to read {} bytes from backend", blob_size);
+
         let mut chunks = self.read_chunks(region.blob_address, blob_size, &region.chunks)?;
         assert_eq!(region.chunks.len(), chunks.len());
 
--- a/storage/src/cache/worker.rs
+++ b/storage/src/cache/worker.rs
@@ -17,6 +17,7 @@ use governor::state::{InMemoryState, NotKeyed};
 use governor::{Quota, RateLimiter};
 use nydus_utils::metrics::{BlobcacheMetrics, Metric};
 use tokio::runtime::Runtime;
+use tokio::sync::Semaphore;
 
 use nydus_api::http::BlobPrefetchConfig;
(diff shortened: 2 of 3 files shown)


storage/src/cache/worker.rs
```diff
@@ -235,28 +238,38 @@ impl AsyncWorkerMgr {
...
+                    let _ = mgr2.prefetch_sema.acquire().await;
                     if state.load(Ordering::Acquire) > 0 {
-                        let _ = rt.spawn(Self::handle_blob_prefetch_request(
-                            mgr.clone(),
-                            blob_cache,
-                            offset,
-                            size,
-                            state.clone(),
-                        ));
+                        rt.spawn(async move {
```
changweige, line 252: Why delegate the fetching data Future into the same tokio runtime rather than directly run to the end of future `handle_prefetch_requests` . Moreover, it seems the runtime `rt` only has a thread, so re-generate a future is efficent ?

storage/src/cache/cachedfile.rs
```diff
@@ -352,10 +352,21 @@ impl BlobObject for FileCacheEntry {
             return Ok(0);
         }
 
+        let mut size = chunks[0].compress_size();
         for idx in 1..chunks.len() {
             if chunks[idx - 1].id() + 1 != chunks[idx].id() {
                 return Err(einval!("chunks for fetch_chunks() must be continuous"));
             }
+            size += chunks[idx].compress_size();
```
changweige, line 360: Using the insane check is fine. Can we just use `BlobIoRange::blob_size` in case the insane check is removed in the future

storage/src/meta/mod.rs
```diff
@@ -563,6 +563,57 @@ impl BlobMetaInfo {
...
+        let mut vec = chunks.to_vec();
+        while index + 1 < infos.len() {
+            index += 1;
+            let entry = &infos[index];
+            if self.validate_chunk(entry).is_err() || entry.compressed_offset() != last_end {
+                break;
+            }
+
+            // Avoid read amplify if next chunk is too big.
+            if last_end >= end && entry.compressed_end() > batch_end {
```
changweige, line 601: `last_end >= end` is always true?

storage/src/meta/mod.rs
```diff
@@ -563,6 +563,57 @@ impl BlobMetaInfo {
...
+        let batch_end = std::cmp::min(
+            end.checked_add(max_size).unwrap_or(end),
+            self.state.compressed_size,
+        );
+        if batch_end <= end {
+            return None;
+        }
+
+        let mut last_end = end;
+        let mut vec = chunks.to_vec();
```
changweige, line 592: Better we can avoid the memory copy here?

storage/src/cache/worker.rs
jiangliu: The current future acts as a dispatcher, which receives requests from the mpmc queue and dispatch them.
So there may be multiple inflight requests.
Currently two inlfight-requests are allowed.

storage/src/cache/cachedfile.rs
jiangliu: Good point，fixed

storage/src/meta/mod.rs
jiangliu: That needs to change too much interfaces, so simplify it by clone().

storage/src/meta/mod.rs
jiangliu: Fixed

changweige (approved):
LGTM

PR #610, opened 2022-07-26
storage: fetch blob.meta in asynchronous mode

jiangliu:
Fetch blob.meta in asynchronous mode, to reduce container startup time.

### DIFF
--- a/storage/src/cache/cachedfile.rs
+++ b/storage/src/cache/cachedfile.rs
@@ -14,8 +14,9 @@ use std::io::{ErrorKind, Result, Seek, SeekFrom};
 use std::mem::ManuallyDrop;
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 use std::slice;
-use std::sync::atomic::{AtomicU32, Ordering};
-use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
 
 use fuse_backend_rs::file_buf::FileVolatileSlice;
 use nix::sys::uio;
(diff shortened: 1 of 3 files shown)


imeoer (approved):

changweige:
Unfortunately, I observed a performance regression with the PR applied to main branch :-(

jiangliu:
> Unfortunately, I observed a performance regression with the PR applied to main branch :-(

That doesn't make sense:)
It should not cause so big regression, could you please help to retest?

I could always observe performance improvements in my env.

storage/src/cache/cachedfile.rs
```diff
@@ -36,11 +37,69 @@ use crate::meta::{BlobMetaChunk, BlobMetaInfo};
 use crate::utils::{alloc_buf, copyv, readv, MemSliceCursor};
 use crate::{StorageError, StorageResult, RAFS_DEFAULT_CHUNK_SIZE};
 
+const DOWNLOAD_META_RETRY_COUNT: u32 = 5;
+const DOWNLOAD_META_RETRY_DELAY: u64 = 500;
+
+#[derive(Default, Clone)]
+pub(crate) struct FileCacheMeta {
+    status: Arc<AtomicBool>,
```
liubogithub, line 45: Do you mind adding a comment for `status`? 
looks like `status` indicates the failure of downloading blob.meta

changweige:
> > Unfortunately, I observed a performance regression with the PR applied to main branch :-(
> 
> That doesn't make sense:) It should not cause so big regression, could you please help to retest?
> 
> I could always observe performance improvements in my env.

I must be too sleepy that night. 😹 Yes performance with this PR. Sorry for the noise I made

storage/src/cache/cachedfile.rs
```diff
@@ -137,17 +196,26 @@ impl BlobCache for FileCacheEntry {
...
-                    let cki = BlobMetaChunk::new(chunk_index as usize, &chunks_meta.state);
-                    b.chunkinfo = BlobIoChunk::Base(Arc::new(cki));
+            if let BlobIoChunk::Address(_blob_index, chunk_index) = b.chunkinfo {
+                if let Some(meta) = self.meta.as_ref() {
+                    if let Some(bm) = meta.get_blob_meta() {
+                        let cki = BlobMetaChunk::new(chunk_index as usize, &bm.state);
+                        b.chunkinfo = BlobIoChunk::Base(Arc::new(cki));
+                    } else {
+                        warn!("failed to get blob.meta for prefetch");
+                        fail = true;
```
changweige, line 208: Can we just return error from here>

storage/src/cache/cachedfile.rs
```diff
@@ -36,11 +37,69 @@ use crate::meta::{BlobMetaChunk, BlobMetaInfo};
 use crate::utils::{alloc_buf, copyv, readv, MemSliceCursor};
 use crate::{StorageError, StorageResult, RAFS_DEFAULT_CHUNK_SIZE};
 
+const DOWNLOAD_META_RETRY_COUNT: u32 = 5;
+const DOWNLOAD_META_RETRY_DELAY: u64 = 500;
```
changweige, line 41: totally waiting for 2.5 seconds in a terrible network environment could not be enough. Can we make it longer like 5 seconds or 10 seconds?

storage/src/cache/cachedfile.rs
jiangliu: renamed to `has_error`.

storage/src/cache/cachedfile.rs
jiangliu: We are within an "for_each()" loop and can't return error here.

storage/src/cache/cachedfile.rs
jiangliu: changed to 20s.

changweige: Looks it was not changed, forget to stage changes?

jiangliu: Sorry, changed to 10s instead of 20s:)

PR #611, opened 2022-07-27
Prefetch limit

changweige:
The buffer size could be much larger than the real user io request.
For example, fuse-backend-rs will allocate 1MB buffer for each IO
regardless of its read request size ending up with eat up all
prefetch bandwidth when starting the container.

### DIFF
--- a/storage/src/cache/dummycache.rs
+++ b/storage/src/cache/dummycache.rs
@@ -145,13 +145,21 @@ impl BlobCache for DummyCache {
                 let mut d = alloc_buf(bio.chunkinfo.uncompress_size() as usize);
                 self.read_raw_chunk(&bio.chunkinfo, d.as_mut_slice(), false, None)?;
                 buffer_holder.push(d);
+                // Even a merged IO can hardly reach u32::MAX. So this is safe
                 user_size += bio.size;
             }
         }
 
-        copyv(&buffer_holder, bufs, offset as usize, user_size, 0, 0)
-            .map(|(n, _)| n)
-            .map_err(|e| eother!(e))
+        copyv(
+            &buffer_holder,
+            bufs,
+            offset as usize,
+            user_size as usize,
+            0,
+            0,
+        )
+        .map(|(n, _)| n)
+        .map_err(|e| eother!(e))
     }
 }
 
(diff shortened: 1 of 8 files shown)


storage/src/cache/dummycache.rs
```diff
@@ -149,9 +149,17 @@ impl BlobCache for DummyCache {
...
         }
 
-        copyv(&buffer_holder, bufs, offset as usize, user_size, 0, 0)
-            .map(|(n, _)| n)
-            .map_err(|e| eother!(e))
+        copyv(
+            &buffer_holder,
+            bufs,
+            offset as usize,
+            // Even a merged IO can hardly reach u32::MAX. So this is safe
```
jiangliu, line 156: Seems this comment is for line 148? It's always safe to cast u32 to usize.

storage/src/cache/worker.rs
```diff
@@ -84,12 +83,12 @@ impl AsyncPrefetchMessage {
     }
 }
 
+/// An asynchronous tasks pool manager which is only for prefetch IO right now.
 pub(crate) struct AsyncWorkerMgr {
     metrics: Arc<BlobcacheMetrics>,
     ping_requests: AtomicU32,
     workers: AtomicU32,
     active: AtomicBool,
-
```
jiangliu, line 92: The blank line is a separator for field groups:)

storage/src/cache/worker.rs
```diff
@@ -84,12 +83,12 @@ impl AsyncPrefetchMessage {
     }
 }
 
+/// An asynchronous tasks pool manager which is only for prefetch IO right now.
```
jiangliu, line 86: An asynchronous task manager for data prefetching?

storage/src/cache/dummycache.rs
changweige: Yes. Let's move it to L148

storage/src/cache/worker.rs
changweige: Ok. let's keep the blank line

storage/src/cache/worker.rs
changweige: SGTM

jiangliu (approved):

liubogithub (approved):
Looks good.

PR #613, opened 2022-07-27
Dump nydusd panic info and backtrace into logger and show nydusd git commit version

changweige:
This RR helps debug and develop nydusd.
1. Nydusd is sometimes forked by nydus-snapshotter and has no stdout and stderr to print panic info. So dumping the panic info into logger is a good choice.
2. print nydusd git commit version by nydusctl CLI tool

This can address https://github.com/containerd/nydus-snapshotter/issues/104

### DIFF
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -180,6 +180,12 @@ pub fn setup_logging(log_file_path: Option<PathBuf>, level: LevelFilter) -> Resu
     }
 
     log::set_max_level(level);
+
+    // Dump panic info and backtrace to logger.
+    log_panics::Config::new()
+        .backtrace_mode(log_panics::BacktraceMode::Resolved)
+        .install_panic_hook();
+
     Ok(())
 }
 
--- a/src/bin/nydusd/main.rs
+++ b/src/bin/nydusd/main.rs
@@ -717,6 +717,7 @@ fn main() -> Result<()> {
     let apisock = args.value_of("apisock");
 
     setup_logging(logging_file, level)?;
+
     dump_program_info(crate_version!());
     handle_rlimit_nofile_option(&args, "rlimit-nofile")?;
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -932,6 +932,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "log-panics"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68f9dd8546191c1850ecf67d22f5ff00a935b890d0e84713159a55495cc2ac5f"
+dependencies = [
+ "backtrace",
+ "log",
+]
+
+[[package]]
 name = "lz4-sys"
 version = "1.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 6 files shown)


jiangliu:
The features is fantastic. Could we remove the dependency on "log-panics"?

liubogithub (approved):

liubogithub:
> The features is fantastic. Could we remove the dependency on "log-panics"?

I checked with log_panics code, it's surprisingly simple, so to avoid the crate deps, maybe we can have our own `panic::set_hook(Box::new(move |info| {` in apps/lib.rs

@changweige @jiangliu what do you think?

src/bin/nydusd/main.rs
```diff
@@ -717,6 +717,11 @@ fn main() -> Result<()> {
     let apisock = args.value_of("apisock");
 
     setup_logging(logging_file, level)?;
+
+    log_panics::Config::new()
```
imeoer, line 721: Maybe move it into `setup_logging`, so that enables the panic hook for `nydus-image`, `nydusd`, `nydus-cached`.

changweige: good point

changweige:
> The features is fantastic. Could we remove the dependency on "log-panics"?

It is a little difficult. Even if we can get rid of crate `log_panics`, will still have to depend on crate `backtrace`. I prefer including the feature ASAP so we can move to other urgent tasks. If we have move bandwidth in the future, let's try to get rid of it.

changweige:
We can get panic info in the log like
```
[2022-07-28 11:03:57.257415 +08:00] ERROR [/home/gechangwei/.cargo/registry/src/mirrors.sjtug.sjtu.edu.cn-4f7dbcce21e258a2/log-panics-2.1.0/src/lib.rs:130] thread 'fuse_server' panicked at 'assertion failed: self.offset + self.len == _offset': storage/src/device.rs:671
   0: backtrace::capture::Backtrace::create
   1: backtrace::capture::Backtrace::new

FAILED

```

imeoer (approved):

app/src/lib.rs
```diff
@@ -180,6 +180,12 @@ pub fn setup_logging(log_file_path: Option<PathBuf>, level: LevelFilter) -> Resu
     }
 
     log::set_max_level(level);
+
+    // Dump panic info and backtrace to logger.
+    log_panics::Config::new()
+        .backtrace_mode(log_panics::BacktraceMode::Resolved)
+        .install_panic_hook();
```
bergwolf, line 187: Looks like this is equivalent to simply calling
```
log_panics::init()
```

ref: https://docs.rs/log-panics/latest/log_panics/

jiangliu:
> > The features is fantastic. Could we remove the dependency on "log-panics"?
> 
> I checked with log_panics code, it's surprisingly simple, so to avoid the crate deps, maybe we can have our own `panic::set_hook(Box::new(move |info| { ... }` in apps/lib.rs
> 
> @changweige @jiangliu what do you think?

Let's move on first, and simplify it later.

app/src/lib.rs
changweige: I am afraid they are not the same. I tried `log_panics::init()` ending up with no panic backtrace and resolved function name?

jiangliu (approved):

PR #614, opened 2022-07-28
storage: fix range prefetch algorithm from panic and io hang

changweige:
The original algorithm uses end as the last data chunk index
caused panic since` end` can be equal to pending.len().

Moreover, `end` chunk's pending flag is not cleared after the chunk
is fetched from remote storage

### DIFF
--- a/storage/src/cache/cachedfile.rs
+++ b/storage/src/cache/cachedfile.rs
@@ -316,19 +316,25 @@ impl BlobCache for FileCacheEntry {
         let mut total_size = 0;
         let mut start = 0;
         while start < pending.len() {
-            let mut end = start + 1;
-            while end < pending.len() && pending[end].id() == pending[end - 1].id() + 1 {
+            let mut end = start;
+
+            // `end` should be inclusive within `pending` slice.
+            while end < pending.len() - 1 && pending[end + 1].id() == pending[end].id() + 1 {
                 end += 1;
             }
 
             // Find a range with continuous chunk id
             let blob_offset = pending[start].compressed_offset();
+
+            // NOTE: Please be aware that being fetched chunks include index=`end` chunk in below `pending` slice.
+            // Do't forget to clear its pending state whenever backend IO fails.
             let blob_end = pending[end].compressed_offset() + pending[end].compressed_size() as u64;
             let blob_size = (blob_end - blob_offset) as usize;
-            match self.read_chunks(blob_offset, blob_size, &pending[start..end], true) {
+
+            match self.read_chunks(blob_offset, blob_size, &pending[start..=end], true) {
                 Ok(v) => {
                     total_size += blob_size;
-                    for idx in start..end {
+                    for idx in start..=end {
                         let offset = if self.is_compressed {
                             pending[idx].compressed_offset()
                         } else {
@@ -343,13 +349,14 @@ impl BlobCache for FileCacheEntry {
                     }
                 }
                 Err(_e) => {
-                    for chunk in pending.iter().take(end).skip(start) {
+                    // NOTE: We must clear pending flag for `end` chunk.
+                    for chunk in &mut pending[start..=end] {
                         self.chunk_map.clear_pending(chunk);
                     }
                 }
             }
 
-            start = end;
+            start = end + 1;
         }
 
         Ok(total_size)


bergwolf:
Is it possible to add a UT or something to validate the change? It's really an core code path that is worth a test case (and easily be broken if we don't have one, like right now).

changweige:
> Is it possible to add a UT or something to validate the change? It's really a core code path that is worth a test case (and prone to be broken if we don't have one, like right now).

It is a little difficult since `prefetch_range` has too many lines and is calling a few methods of other structs. Mocking blob cache chunk map (in `FileCacheEntry`) and backend reader to cover each abnormal branch is not easy. In fact, most logic can be covered by e2e test. As no e2e test so far goes into this function, the flaw is not detected. For backend read failure, we can try to do some failure simulation in `banckend_proxy` 🥶

changweige:
Can we prioritize merging this PR? I think it is a critical problem. @bergwolf @jiangliu

storage/src/cache/cachedfile.rs
```diff
@@ -250,19 +250,21 @@ impl BlobCache for FileCacheEntry {
         let mut total_size = 0;
         let mut start = 0;
         while start < pending.len() {
-            let mut end = start + 1;
-            while end < pending.len() && pending[end].id() == pending[end - 1].id() + 1 {
+            let mut end = start;
+
+            while end < pending.len() - 1 && pending[end + 1].id() == pending[end].id() + 1 {
                 end += 1;
```
bergwolf, line 256: `end` is exclusive in the original code. The patch changes it to be inclusive. How does it fix a panic?

bergwolf (commented):
Given that there are no tests to validate the change and it is a critical code path, I need to better understand it before giving it a pass. Can you attach the panic trace that you are seeing? Better some explanation on why it panics and how the patch fixes it.

storage/src/cache/cachedfile.rs
changweige: > Given that there are no tests to validate the change and it is a critical code path, I need to better understand it before giving it a pass. Can you attach the panic trace that you are seeing? Better some explanation on why it panics and how the patch fixes it.

Please check out the panic backtrace. I will also append one into the commit log.

```
thread 'cache-flusher' panicked at 'index out of bounds: the len is 2 but the index is 2': storage/src/cache/cachedfile.rs:328
   0: <backtrace::capture::Backtrace as core::default::Default>::default
             at /home/gechangwei/.cargo/registry/src/mirrors.sjtug.sjtu.edu.cn-4f7dbcce21e258a2/backtrace-0.3.66/src/capture.rs:410:9
   1: log_panics::Config::install_panic_hook::{{closure}}
             at /home/gechangwei/.cargo/registry/src/mirrors.sjtug.sjtu.edu.cn-4f7dbcce21e258a2/log-panics-2.1.0/src/lib.rs:115:29
   2: std::panicking::rust_panic_with_hook
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:702:17
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:588:13
   4: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/sys_common/backtrace.rs:138:18
   5: rust_begin_unwind
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   6: core::panicking::panic_fmt
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
   7: core::panicking::panic_bounds_check
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:84:5
   8: <usize as core::slice::index::SliceIndex<[T]>>::index
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/slice/index.rs:241:10
   9: core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/slice/index.rs:18:9
  10: <alloc::vec::Vec<T,A> as core::ops::index::Index<I>>::index
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/alloc/src/vec/mod.rs:2533:9
  11: <nydus_storage::cache::cachedfile::FileCacheEntry as nydus_storage::cache::BlobCache>::prefetch_range
             at /data00/home/gechangwei/git_repo/nydus-rs/storage/src/cache/cachedfile.rs:328:28
  12: nydus_storage::cache::worker::AsyncWorkerMgr::handle_fs_prefetch_request
             at /data00/home/gechangwei/git_repo/nydus-rs/storage/src/cache/worker.rs:392:9
  13: nydus_storage::cache::worker::AsyncWorkerMgr::handle_prefetch_requests::{{closure}}::{{closure}}
             at /data00/home/gechangwei/git_repo/nydus-rs/storage/src/cache/worker.rs:268:37
  14: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
             at /home/gechangwei/.cargo/registry/src/mirrors.sjtug.sjtu.edu.cn-4f7dbcce21e258a2/tokio-1.20.0/src/runtime/blocking/task.rs:42:21
  15: tokio::runtime::task::core::CoreStage<T>::poll::{{closure}}
             at /home/gechangwei/.cargo/registry/src/mirrors.sjtug.sjtu.edu.cn-4f7dbcce21e258a2/tokio-1.20.0/src/runtime/task/core.rs:165:17
```

bergwolf: Please DO NOT edit my comment :) Your comments can be raised as usual.

storage/src/cache/cachedfile.rs
```diff
@@ -318,19 +318,21 @@ impl BlobCache for FileCacheEntry {
...
             }
 
             // Find a range with continuous chunk id
             let blob_offset = pending[start].compress_offset();
             let blob_end = pending[end].compress_offset() + pending[end].compress_size() as u64;
```
bergwolf, line 329: So it seems that the only change that we need is to fix this line here, to make sure `end` is exclusive:
```
let blob_end = pending[end - 1].compress_offset() + pending[end - 1].compress_size() as u64;
```

Looks like a typo of the original code where one forgets that `end` is exclusive. Worth a comment at line 321 to remind readers that `end` is exclusive while `start` is inclusive.

changweige: Yes. It might be another way to address this. But I prefer current PR since it's more straightforward to represent the 
 intention and no need to add a comment line to mitigate the gap between code author and readers. Besides for L348,` take(end)` where the pending is not that long is really burning the brain 😥

changweige: Yes. It might be another way to address this. But I prefer current PR since it's more straightforward to represent the
intention and no need to add a comment line to mitigate the gap between code author and readers. Besides for L348, take(end) where the pending is not that long is really burning the brain 😥

changweige: aaa

bergwolf: IMO, it doesn't make the code easier to understand by pivoting from one way (exclusive ending) to another (inclusive ending). OTOH, a comment stating that we have either `[start, end)` or `[start, end]` makes it clear how the code is intended to be written. W/o such a reminder, even @jiangliu mistook It when he wrote the code in the first place.

changweige: Ture. But I am inspired by the PR https://github.com/dragonflyoss/image-service/pull/626 merged this morning.  It seems that `BlobCache::prefetch_range` and `BlobObject::prefetch_chunks` have very similar logics. So I am trying to remove the duplicated one `prefetch_range`. There are several advantages:

1. Make code neat 
2. prefech_chunks has more sophisticated logic
3. Simplify choice of using what to do backend prefetch

I will raise a PR later and if the new one is not accepted, let's do what you are suggesting with this PR

bergwolf: I don't object to changing the exclusive ending to an inclusive end. But I don't agree that change would make the code any way cleaner/clearer. We need a comment to remind us how `end` is intended to be, either exclusive or inclusive. And that is how the code becomes clearer.

bergwolf: If we want to have similar logics in #626 (which I have to admit that I didn't take a closer look at), please state it clearly in the commit message. Nevertheless, I still think a comment to state that we have either `[start, end)` or `[start, end]` is necessary here.

changweige: Sure.

changweige:
@bergwolf I closed the PR https://github.com/dragonflyoss/image-service/pull/627 which is trying to remove `prefetch_range` and reuse `prefetch_chunks`. But it fails due to incompatible chunk maps used within the two methods. So let's fix the issue still in this PR. And I added comments as a reminder to others now.

jiangliu (approved):

jiangliu (approved):

PR #616, opened 2022-07-28
nydusify: skip tls verification

imeoer:
An error is thrown when converting image to https registry with
self-signed TLS cert:

```
failed to do request: x509: certificte signed by unknown authority
```

1. Fixed it by retrying with `InsecureSkipVerify` option, if users specify
`--source-insecure` or `--target-insecure` option, then skips verifying
HTTPS certs.

2. Allows falling back to plain HTTP for registry if https request failed.

Tested to enable `--source-insecure` and `--target-insecure`:

- Tested with http registry with domain name like `registry.com`;
- Tested with https registry with self-signed TLS cert;

Tested to disable `--source-insecure` and `--target-insecure`:

- Tested with https registry with CA-signed TLS cert;
- Tested with http registry with localhost/127.0.0.1 address;

### DIFF
--- a/contrib/nydusify/cmd/nydusify.go
+++ b/contrib/nydusify/cmd/nydusify.go
@@ -185,8 +185,8 @@ func main() {
 				&cli.StringFlag{Name: "source", Required: true, Usage: "Source image reference", EnvVars: []string{"SOURCE"}},
 				&cli.StringFlag{Name: "target", Required: false, Usage: "Target (Nydus) image reference", EnvVars: []string{"TARGET"}},
 				&cli.StringFlag{Name: "target-suffix", Required: false, Usage: "Add suffix to source image reference as target image reference, conflict with --target", EnvVars: []string{"TARGET_SUFFIX"}},
-				&cli.BoolFlag{Name: "source-insecure", Required: false, Usage: "Allow http/insecure source registry communication", EnvVars: []string{"SOURCE_INSECURE"}},
-				&cli.BoolFlag{Name: "target-insecure", Required: false, Usage: "Allow http/insecure target registry communication", EnvVars: []string{"TARGET_INSECURE"}},
+				&cli.BoolFlag{Name: "source-insecure", Required: false, Usage: "Skip verifying server certs for HTTPS source registry", EnvVars: []string{"SOURCE_INSECURE"}},
+				&cli.BoolFlag{Name: "target-insecure", Required: false, Usage: "Skip verifying server certs for HTTPS target registry", EnvVars: []string{"TARGET_INSECURE"}},
 				&cli.StringFlag{Name: "work-dir", Value: "./tmp", Usage: "Work directory path for image conversion", EnvVars: []string{"WORK_DIR"}},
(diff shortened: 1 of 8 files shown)


contrib/nydusify/cmd/nydusify.go
```diff
@@ -185,8 +185,8 @@ func main() {
 				&cli.StringFlag{Name: "source", Required: true, Usage: "Source image reference", EnvVars: []string{"SOURCE"}},
 				&cli.StringFlag{Name: "target", Required: false, Usage: "Target (Nydus) image reference", EnvVars: []string{"TARGET"}},
 				&cli.StringFlag{Name: "target-suffix", Required: false, Usage: "Add suffix to source image reference as target image reference, conflict with --target", EnvVars: []string{"TARGET_SUFFIX"}},
-				&cli.BoolFlag{Name: "source-insecure", Required: false, Usage: "Allow http/insecure source registry communication", EnvVars: []string{"SOURCE_INSECURE"}},
-				&cli.BoolFlag{Name: "target-insecure", Required: false, Usage: "Allow http/insecure target registry communication", EnvVars: []string{"TARGET_INSECURE"}},
+				&cli.BoolFlag{Name: "source-insecure", Required: false, Usage: "Skips verifying HTTPS certs, and allows falling back to plain HTTP for source registry", EnvVars: []string{"SOURCE_INSECURE"}},
```
changweige, line 188: Skips -> skip ?

changweige: Allows -> allow

contrib/nydusify/pkg/converter/provider/source.go
```diff
@@ -100,16 +101,27 @@ func (sl *defaultSourceLayer) Mount(ctx context.Context) ([]mount.Mount, func()
...
-		defer reader.Close()
 
-		// Decompress layer from source stream
-		if err := utils.UnpackTargz(ctx, sl.mountDir, reader); err != nil {
-			return errors.Wrap(err, fmt.Sprintf("Decompress source layer %s", digestStr))
+		if err := pull(false); err != nil {
+			if !errutil.IsErrHTTPResponseToHTTPSClient(err) && !errutil.IsErrConnectionRefused(err) {
+				return err
+			}
+			return pull(true)
```
bergwolf, line 124: So we always retry without tls. How does the `insecure-xxx` command line argument work?

liubogithub:
The smoke failure seems to have nothing to do with this patch.

contrib/nydusify/cmd/nydusify.go
imeoer: Fixed.

contrib/nydusify/pkg/converter/provider/source.go
imeoer: Updated, `--insecure xxx` means `Skip verifying server certs for HTTPS xxx registry`. We always fall back to HTTP if HTTPS fails (with a warning message).

changweige (approved):
LGTM

PR #619, opened 2022-07-29
nydusd: gc unused blob files after umounting

jiangliu:
After umounting a filesystem, try to gc unused blob files, otherwise
it may break the gc logic in nydus-snapshotter.


imeoer (approved):

bergwolf:
How do we know that a blob is not used by other `nydusd` servers on the same host?

jiangliu:
> How do we know that a blob is not used by other `nydusd` servers on the same host?

GC here just means "close the file", it doesn't really delete those files.

changweige (approved):
LGTM

PR #622, opened 2022-07-30
storage: remove unnecessary debug_assert

changweige:
`self.offset + self.len != _offset` is legal when merge more regions.
So the debug_assert is wrong
Upper layer should ensure its validation.

### DIFF
--- a/storage/src/device.rs
+++ b/storage/src/device.rs
@@ -653,6 +653,8 @@ impl BlobIoVec {
 }
 
 /// A segment representing a continuous range for a blob IO operation.
+/// It can span multiple chunks while the `offset` is where the user io starts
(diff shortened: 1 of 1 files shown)


storage/src/device.rs
```diff
@@ -669,10 +671,6 @@ impl BlobIoSegment {
 
     #[inline]
     pub fn append(&mut self, _offset: u32, len: u32) {
-        debug_assert!(self.offset + self.len == _offset);
-        debug_assert!(_offset.checked_add(len).is_some());
```
jiangliu, line 673: Should we still keep these two checks?

changweige: Let's keep it. I have to preference for it

imeoer (approved):

PR #626, opened 2022-08-01
Optimize performance by fix a bug in prefetch 

jiangliu:
Optimize performance by fix a bug in prefetch. With these patches applied, it performs as better as v5, but reduce half of network IO requests to backend storage system.
截屏2022-08-02 上午12 33 43

### DIFF
--- a/storage/src/cache/cachedfile.rs
+++ b/storage/src/cache/cachedfile.rs
@@ -419,16 +419,27 @@ impl BlobObject for FileCacheEntry {
         self.do_fetch_chunks(&chunks)
     }
 
-    fn fetch_chunks(&self, range: &BlobIoRange) -> Result<usize> {
+    fn prefetch_chunks(&self, range: &BlobIoRange) -> Result<usize> {
         let chunks = &range.chunks;
         if chunks.is_empty() {
             return Ok(0);
         }
 
+        let mut ready_or_pending =
+            matches!(self.chunk_map.is_ready_or_pending(&chunks[0]), Ok(true));
         for idx in 1..chunks.len() {
             if chunks[idx - 1].id() + 1 != chunks[idx].id() {
                 return Err(einval!("chunks for fetch_chunks() must be continuous"));
             }
+            if ready_or_pending
+                && !matches!(self.chunk_map.is_ready_or_pending(&chunks[idx]), Ok(true))
+            {
+                ready_or_pending = false;
+            }
+        }
+        // All chunks to be prefetched are already pending for downloading, no need to reissue.
+        if ready_or_pending {
+            return Ok(0);
         }
 
         if range.blob_size < RAFS_DEFAULT_CHUNK_SIZE {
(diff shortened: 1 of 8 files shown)


storage/src/cache/worker.rs
```diff
@@ -238,15 +238,17 @@ impl AsyncWorkerMgr {
 
     async fn handle_prefetch_requests(mgr: Arc<AsyncWorkerMgr>, rt: &Runtime) {
         // Max 2 active requests per thread.
-        mgr.prefetch_sema.add_permits(2);
+        mgr.prefetch_sema.add_permits(1);
```
changweige, line 241: Inconsistent with above comment

storage/src/cache/worker.rs
```diff
@@ -238,15 +238,17 @@ impl AsyncWorkerMgr {
...
 
         while let Ok(msg) = mgr.prefetch_channel.recv().await {
             mgr.handle_prefetch_rate_limit(&msg).await;
             let mgr2 = mgr.clone();
 
             match msg {
                 AsyncPrefetchMessage::BlobPrefetch(state, blob_cache, offset, size) => {
-                    let _ = mgr2.prefetch_sema.acquire().await;
+                    let token = Semaphore::acquire_owned(mgr2.prefetch_sema.clone())
```
changweige, line 249: Could you help explain what is the specific bug with semaphore?

storage/src/cache/cachedfile.rs
```diff
@@ -447,20 +447,20 @@ impl BlobObject for FileCacheEntry {
...
         }
 
-        self.do_fetch_chunks(chunks)
+        self.do_fetch_chunks(chunks, true)
     }
 }
 
 impl FileCacheEntry {
-    fn do_fetch_chunks(&self, chunks: &[BlobIoChunk]) -> Result<usize> {
+    fn do_fetch_chunks(&self, chunks: &[BlobIoChunk], prefetch: bool) -> Result<usize> {
```
changweige, line 463: The newly added argument `prefetch` is only for debugging purposes? Can we remove it?

storage/src/cache/worker.rs
liubogithub: > Could you help explain what is the specific bug with semaphore?

My 2 cents.
Instead of `add_permits(1)`,  use the existing 2 permit.

liubogithub (approved):
I'd suggest adding more details to the bugfix commit log.

Otherwise, it looks good.

storage/src/cache/worker.rs
jiangliu: Fixed

storage/src/cache/worker.rs
jiangliu: Dropping the object returned by `Semaphore::acquire()` will increase the semaphore, so there are two issues here:
1) drop the returned object too early, it should be dropped by the async task instead.
2) shouldn't call `add_permits()`.

storage/src/cache/cachedfile.rs
jiangliu: It may help to analyze online issues, so added it:)

changweige (approved):
LGTM

PR #628, opened 2022-08-02
storage: remove localfs based prefetch

jiangliu:
The special prefetch in localfs is not actually used, but causes
much complexity, so remove it.


changweige (approved):
LGTM

PR #629, opened 2022-08-02
storage: convert chunk info per as blob meta regardless of

changweige:
indexed_chunk_map

Rafs v6 chunks' info from filesystem only has blob_id and chunk_index.
So storage layer has to convert it. But current code only load blob
meta with disable_indexed_map=false, which only affets v5 image

It cause nydusd panic when do prefetch or fs io

---
[2022-08-02 14:52:45.103790 +08:00] INFO [api/src/http.rs:778] <--- Get Uri { string: "/api/v1/metrics/blobcache" }
   1: nydus_storage::cache::BlobIoMergeState<F>::merge_and_issue
   2: <nydus_storage::cache::cachedfile::FileCacheEntry as nydus_storage::cache::BlobCache[2022-08-02 14:52:45.103943 +08:00] INFO [api/src/http.rs:783] ---> Get Status Code: OK, Elapse: Ok(157.427µs), Body Size: 365
>::prefetch
   3: nydus_storage::device::BlobDevice::prefetch
   4: nydus_rafs::fs::Rafs::do_prefetch::{{closure}}
   5: nydus_rafs::metadata::RafsSuper::prefetch_inode
   6: nydus_rafs::metadata::RafsSuper::prefetch_data
   7: nydus_rafs::metadata::RafsSuper::prefetch_files
   8: nydus_rafs::fs::Rafs::do_prefetch
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

### DIFF
--- a/storage/src/cache/filecache/mod.rs
+++ b/storage/src/cache/filecache/mod.rs
@@ -255,7 +255,10 @@ impl FileCacheEntry {
         // The builder now records the number of chunks in the blob table, so we can
         // use IndexedChunkMap as a chunk map, but for the old Nydus bootstrap, we
         // need downgrade to use DigestedChunkMap as a compatible solution.
-        let chunk_map: Arc<dyn ChunkMap> = if mgr.disable_indexed_map
+
+        let is_v5 = !blob_info.meta_ci_is_valid();
+
+        let chunk_map: Arc<dyn ChunkMap> = if (is_v5 && mgr.disable_indexed_map)
             || blob_info.has_feature(BlobFeatures::V5_NO_EXT_BLOB_TABLE)
         {
             direct_chunkmap = false;


storage/src/cache/filecache/mod.rs
```diff
@@ -202,7 +202,7 @@ impl FileCacheEntry {
             is_direct_chunkmap,
             is_stargz
         );
-        let meta = if is_get_blob_object_supported && blob_info.meta_ci_is_valid() {
+        let meta = if blob_info.meta_ci_is_valid() {
```
jiangliu, line 205: It would be better to tweek value of `is_get_blob_object_supported`, such as
```
let is_get_blob_object_supported = is_v6 && !mgr.is_compressed;
```

jiangliu: Oh, seems it would be even better to enforce direct_chunkmap for v6?
It would be better to refine `create_chunk_map()` instead?
```
let (chunk_map, is_direct_chunkmap) =
            Self::create_chunk_map(mgr, &blob_info, &blob_file_path)?;
```

changweige: But v5 image can also have direct_chunk_map. BTW, how do we define the direct_chunk_map feature?

jiangliu: How about this? The `disable_indexed_map` only takes effect for v5 images.
```
diff --git a/storage/src/cache/filecache/mod.rs b/storage/src/cache/filecache/mod.rs
index f92881b2..d387c144 100644
--- a/storage/src/cache/filecache/mod.rs
+++ b/storage/src/cache/filecache/mod.rs
@@ -255,7 +255,8 @@ impl FileCacheEntry {
         // The builder now records the number of chunks in the blob table, so we can
         // use IndexedChunkMap as a chunk map, but for the old Nydus bootstrap, we
         // need downgrade to use DigestedChunkMap as a compatible solution.
-        let chunk_map: Arc<dyn ChunkMap> = if mgr.disable_indexed_map
+        let is_v5 = !blob_info.meta_ci_is_valid();
+        let chunk_map: Arc<dyn ChunkMap> = if (is_v5 && mgr.disable_indexed_map)
             || blob_info.has_feature(BlobFeatures::V5_NO_EXT_BLOB_TABLE)
         {
             direct_chunkmap = false;
```

changweige: I suppose we don't have to consider what image version (5 or 6) or flag `disable_indexed_map` when loading blob meta info into memory. As long as `blob_info.meta_ci_is_valid()` is True we should load its meta otherwise storage blob io can't work.

changweige: OK. Make sense.

jiangliu (approved):

PR #630, opened 2022-08-02
unpack: fix empty username and groupname

imeoer:
To make the unpacked tar consistent with the OCI-formatted tar before the pack,
we need to backfill the username and groupname in the tar header, which may
break the repeatable build when unpacking in different hosts, but actually has
little effect.


ccx1024cc (approved):
LGTM

PR #633, opened 2022-08-02
Improve prefetch performance further

jiangliu:
Improve prefetch performance further, now v5 image performs as better as 1.1.2.

### DIFF
--- a/rafs/src/fs.rs
+++ b/rafs/src/fs.rs
@@ -530,7 +530,7 @@ impl Rafs {
             }
         }
 
-        let ignore_prefetch_all = prefetch_files
+        let mut ignore_prefetch_all = prefetch_files
             .as_ref()
             .map(|f| f.len() == 1 && f[0].as_os_str() == "/")
             .unwrap_or(false);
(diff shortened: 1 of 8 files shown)


rafs/src/metadata/md_v5.rs
```diff
@@ -83,13 +83,18 @@ impl RafsSuper {
         Ok(meta_size as usize)
     }
 
-    pub(crate) fn prefetch_data_v5<F>(&self, r: &mut RafsIoReader, fetcher: F) -> RafsResult<usize>
+    pub(crate) fn prefetch_data_v5<F>(
+        &self,
+        r: &mut RafsIoReader,
+        root_nid: Inode,
```
changweige, line 89: This might be misnamed. V5 has no nid concept

rafs/src/metadata/md_v6.rs
```diff
@@ -67,14 +67,18 @@ impl RafsSuper {
         }
     }
 
-    pub(crate) fn prefetch_data_v6<F>(&self, r: &mut RafsIoReader, fetcher: F) -> RafsResult<usize>
+    pub(crate) fn prefetch_data_v6<F>(
+        &self,
+        r: &mut RafsIoReader,
+        root_nid: Inode,
```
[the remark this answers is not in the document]
changweige: IMO, nid is used to address EROFS disk structure

storage/src/device.rs
```diff
@@ -652,6 +654,38 @@ impl BlobIoVec {
     }
 }
 
+/// Helper structure to merge blob IOs to reduce IO requests.
+#[derive(Default)]
+pub struct BlobIoMerge {
+    map: HashMap<String, BlobIoVec>,
+    current: String,
+}
```
changweige, line 662: I don't understand how the prefetch merge is improved with this patch. Could you help explain further?
Besides, it never tries to merge backend blob io from Rafs layer. It only gathers chunks that might be on different blobs and issue them to the storage layer. Backend blob IO will be merged in the storage layer.  Moreover, seems the newly introduced structure `BlobIoMerge` brings many memory allocation/overhead. Afterwards the hash map key is never used when issue prefetch.

rafs/src/fs.rs
```diff
@@ -503,28 +504,45 @@ impl Rafs {
...
+            for blob in sb.superblock.get_blob_infos() {
+                let sz = blob.readahead_size();
+                if sz > 0 {
+                    let mut offset = 0;
+                    while offset < sz {
+                        let len = if sz - offset < RAFS_DEFAULT_CHUNK_SIZE {
+                            sz - offset
+                        } else {
+                            RAFS_DEFAULT_CHUNK_SIZE
+                        };
```
liubogithub, line 520: can we use cmp::min?

rafs/src/metadata/md_v5.rs
jiangliu: Fixed

rafs/src/metadata/md_v6.rs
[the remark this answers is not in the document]
jiangliu: fixed

rafs/src/fs.rs
jiangliu: fixed

storage/src/device.rs
jiangliu: The rafs layer prefetch will try to prefetch one or more files. We have observed that those merged requests are still small, due to adjacent files are stored on different blobs. So the patch tries to merge
```
file 1: blob 1, chunk 1
file 2: blob 2, chunk 1
file 3: blob 1, chunk 2
file 4: blob 2, chunk 2
```
into
```
request 1: blob 1, chunk 1-2
request 2: blob 2, chunk 1-2
```
With current implementation, four requests will be issued instead of two.

jiangliu: And compared with network latency, a few more memory allocation are acceptable tradeoff:)

changweige (approved):
LGTM

PR #636, opened 2022-08-03
builder: fix broken symlink in unpack

imeoer:
When a symlink file in nydus with an absolute path like `/usr/bin/touch -> /bin/touch`,
unpack will output the tar with symlink file like `/usr/bin/touch -> bin/touch`,
this behavior causes that the symlink is broken.


jiangliu:
I have contribute safe-path crate to kata, seems we may use it for nydus too:)

imeoer:
@jiangliu Good work! In this PR we just convert RAFS inode into tar header, so I think there is no security requirement yet, but maybe we can consider using `safe-path` to refactor for nydus, thanks!

ccx1024cc (approved):
LGTM

PR #637, opened 2022-08-03
ci: remove nydus-cached binary build

imeoer:
The nydus-cached functionality was moved to nydusd daemon mode,
so we don't need to compile this binary in CI anymore.


changweige (approved):
LGTM

PR #642, opened 2022-08-05
builder: fix rafs v5 dir size for reproducible build

imeoer:
Since different filesystems may calculate different directory sizes,
for RAFS v5, nydus needs to implement the directory size calculation
itself to achieve reproducible builds on different hosts or filesystems.

For RAFS v6, the directory size is calculated by actual entries.

### DIFF
--- a/src/bin/nydus-image/builder/directory.rs
+++ b/src/bin/nydus-image/builder/directory.rs
@@ -17,6 +17,12 @@ use crate::core::context::{
 };
 use crate::core::node::{Node, Overlay};
 use crate::core::tree::Tree;
+use nydus_utils::try_round_up_4k;
+
+// The RAFS v5 format does not have directory entries, so for ensuring
+// that the directory size is independent of the host filesystem for
+// reproducible builds, we define a virtual dummy value here.
+const RAFS_V5_VIRTUAL_ENTRY_SIZE: u8 = 8;
 
 const TAR_BLOB_NAME: &str = "image.blob";
 const TAR_BOOTSTRAP_NAME: &str = "image.boot";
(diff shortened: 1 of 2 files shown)


eryugey (approved):
LGTM

src/bin/nydus-image/builder/directory.rs
```diff
@@ -17,6 +17,12 @@ use crate::core::context::{
 };
 use crate::core::node::{Node, Overlay};
 use crate::core::tree::Tree;
+use nydus_utils::try_round_up_4k;
+
+// The RAFS v5 format does not have directory entries, so for ensuring
+// that the directory size is independent of the host filesystem for
+// reproducible builds, we define a virtual dummy value here.
+const RAFS_V5_VIRTUAL_ENTRY_SIZE: u8 = 8;
```
jiangliu, line 25: How about defining it as "u64"?

src/bin/nydus-image/builder/directory.rs
```diff
@@ -85,6 +92,26 @@ impl DirectoryBuilder {
         Self {}
     }
 
+    // Since different filesystems may calculate different directory sizes,
+    // for RAFS v5, nydus needs to implement the directory size calculation
+    // itself to achieve reproducible builds on different hosts or filesystems.
+    // For RAFS v6, the directory size is calculated by actual entries later.
+    fn set_v5_dir_size(fs_version: &RafsVersion, tree: &mut Tree) {
```
jiangliu, line 99: How about passing "RafsVersion" instead of "&RafsVersion"?

jiangliu:
Let's it first and I will submit a following on PR to fix the minor refinements.

jiangliu (approved):

PR #643, opened 2022-08-05
nydus-image: fix incorrect node offset in rafs v6 for hardlinks

yixiaojie0x:
Hardlinks in RAFS v6 have different inode numbers. To fix it, we need to set correct offset for hardlinks.

Signed-off-by: yixiaojie <yixiaojie@bytedance.com>

### DIFF
--- a/src/bin/nydus-image/core/node.rs
+++ b/src/bin/nydus-image/core/node.rs
@@ -1066,20 +1066,27 @@ impl Node {
     }
 
     /// Set node offset in bootstrap and return the next position.
-    pub fn v6_set_offset(&mut self, bootstrap_ctx: &mut BootstrapContext) {
+    pub fn v6_set_offset(
+        &mut self,
+        bootstrap_ctx: &mut BootstrapContext,
+        v6_hardlink_offset: Option<u64>,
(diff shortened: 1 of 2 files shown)


src/bin/nydus-image/core/node.rs
```diff
@@ -1079,20 +1079,27 @@ impl Node {
     }
 
     /// Set node offset in bootstrap and return the next position.
-    pub(crate) fn set_v6_offset(&mut self, bootstrap_ctx: &mut BootstrapContext) {
+    pub(crate) fn set_v6_offset(
+        &mut self,
+        bootstrap_ctx: &mut BootstrapContext,
+        hardlink_v6_offset: u64,
```
jiangliu, line 1085: How about using `Option<u64>` here? So we could use `None` for non-hardlink files.

changweige: Otherwise this comment, LGTM

yixiaojie0x: Done. How about this patch?

jiangliu (approved):

PR #648, opened 2022-08-08
nydus-image: ensure has_attr flag is set correctly

jiangliu:
There are some cases missing the has_attr flag, it should be set
when there are extended attributes.


imeoer (approved):

imeoer:
Thanks for this!

PR #650, opened 2022-08-08
storage: reflect prefetch state from each blob cache object

changweige:
The blob cache manager is associated with prefetch state, so we don't
have to clone it and pass it everywhere.
Thus simplifying code.

### DIFF
--- a/storage/src/cache/worker.rs
+++ b/storage/src/cache/worker.rs
@@ -52,9 +52,9 @@ impl From<BlobPrefetchConfig> for AsyncPrefetchConfig {
 /// Asynchronous service request message.
 pub(crate) enum AsyncPrefetchMessage {
     /// Asynchronous blob layer prefetch request with (offset, size) of blob on storage backend.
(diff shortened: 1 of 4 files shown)


storage/src/cache/worker.rs
```diff
@@ -289,15 +292,25 @@ impl AsyncWorkerMgr {
         // Allocate network bandwidth budget
         if let Some(limiter) = &self.prefetch_limiter {
             let size = match msg {
-                AsyncPrefetchMessage::BlobPrefetch(state, _blob_cache, _offset, size) => {
-                    if state.load(Ordering::Acquire) > 0 {
+                AsyncPrefetchMessage::BlobPrefetch(blob_cache, _offset, size) => {
+                    if blob_cache
```
jiangliu, line 296: We could simplify code further by adding a helper function for the repeated code snippet to get state.

jiangliu:
Otherwise LGTM.

storage/src/cache/worker.rs
changweige: Wrapped it into a helper function now

jiangliu (approved):

PR #656, opened 2022-08-09
Get rid of BlobIoChunk::V5

changweige:
Methods like file_offset() are not required in the storage layer. So we early fill the io vec with BlobChunkInfo directly to
get rid of the chunk conversion. User IO file offset and in-chunk offset mapping is provided in io vec control block. This is the first step in trying to reduce storage layer complications by simplify essential IO structures and remove everywhere `as_base`

### DIFF
--- a/storage/src/device.rs
+++ b/storage/src/device.rs
@@ -417,8 +417,6 @@ pub enum BlobIoChunk {
     // Try to avoid the value conversion? Maybe, we can just use trait object `BlobChunkInfo`
     // in the storage module rather than `BlobIoChunk`, it makes data conversion complicated.
     Base(Arc<dyn BlobChunkInfo>),
-    // For rafs v5 to pass chunk info to storage module.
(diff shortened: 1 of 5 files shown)


storage/src/device.rs
```diff
@@ -417,29 +417,17 @@ pub enum BlobIoChunk {
...
-    V5(Arc<dyn self::v5::BlobV5ChunkInfo>),
 }
 
 impl BlobIoChunk {
     /// Convert a [BlobIoChunk] to a reference to [BlobChunkInfo] trait object.
     pub fn as_base(&self) -> &(dyn BlobChunkInfo) {
         match self {
             BlobIoChunk::Base(v) => &**v,
-            BlobIoChunk::V5(v) => v.as_base(),
+            // BlobIoChunk::V5(v) => v.as_base(),
```
jiangliu, line 427: How about removing the above line?

jiangliu (approved):

PR #659, opened 2022-08-09
nydus-test: delay for some time before verifying prefetch metrics

changweige:
Somehow, prefetch is not started very fast ending up with zero
prefetch metrics retained.

**This is fixing broken nightly CI**


jiangliu:
The dbs-uhttp limits the maximum concurrent connections to MAX_CONNECTIONS(10). Maybe we should enlarge the limit.

jiangliu (approved):

Issue #662, opened 2022-08-09
Nydusify can't convert OCI image with a big file to nydus image


yixiaojie0x:
I tried to convert the OCI image ubuntu_fio:oci to nydus v6 image, which has a single 5G file 
fio_test_image in it. 

> root@xxx-xxx-xxx:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/4/fs/root# du * -sh
5.1G    fio_test_image

I used nydusify to convert it with

> time ./nydusify convert  --source hub.xxx.org/nydus_io_test/ubuntu_fio:oci   --target localhost:5000/ubuntu_fio:nydus_v6   --nydus-image /home/yixiaojie/image-service/target/debug/nydus-image   --work-dir ../nydusify_work_dir   --fs-version 6

And I got  a panic, here comes the info
> INFO[2022-08-09T10:41:11+08:00] Version: 846cc9f1f8010aad47fb49806fd5ece7ca656202.20220808.1020 
INFO[2022-08-09T10:41:11+08:00] Parsing image hub.byted.org/nydus_io_test/ubuntu_fio:oci 
INFO[2022-08-09T10:41:11+08:00] Converting to localhost:5000/ubuntu_fio:nydus_v6 
INFO[2022-08-09T10:41:12+08:00] [SOUR] Mount layer                            Digest="sha256:c5ec52c98b3193052e15d783aca2bef10d8d829fa0d58fedfede511920b8f997" Size="75 MB"
INFO[2022-08-09T10:41:12+08:00] [SOUR] Mount layer                            Digest="sha256:d933381eb13674378341a78f73f55a89432cde50779a839d532ea2546c39480c" Size="5.4 GB"
INFO[2022-08-09T10:41:13+08:00] [SOUR] Mount layer                            Digest="sha256:c5ec52c98b3193052e15d783aca2bef10d8d829fa0d58fedfede511920b8f997" Size="75 MB" Time=1.02757099s
INFO[2022-08-09T10:41:13+08:00] [DUMP] Build layer                            Digest="sha256:c5ec52c98b3193052e15d783aca2bef10d8d829fa0d58fedfede511920b8f997" Size="75 MB"
INFO[2022-08-09T10:41:18+08:00] [DUMP] Build layer                            Digest="sha256:c5ec52c98b3193052e15d783aca2bef10d8d829fa0d58fedfede511920b8f997" Size="75 MB" Time=5.391367132s
[...]

PR #663, opened 2022-08-09
builder: fix panic on building big file

imeoer:
Panic when the size of a regular file exceeds 4GB:

```
thread 'main' panicked at 'attempt to multiply with overflow': src/bin/nydus-image/core/node.rs:371
```

Fix it by changing `(chunk_size * i) as u64` to `chunk_size as u64 * i as u64`.

Fix https://github.com/dragonflyoss/image-service/issues/662

### DIFF
--- a/src/bin/nydus-image/core/node.rs
+++ b/src/bin/nydus-image/core/node.rs
@@ -368,7 +368,7 @@ impl Node {
             let file_offset = i as u64 * chunk_size as u64;
             let uncompressed_size = if i == self.inode.child_count() - 1 {
                 (self.inode.size() as u64)
-                    .checked_sub((chunk_size * i) as u64)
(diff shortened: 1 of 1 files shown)


src/bin/nydus-image/core/node.rs
```diff
@@ -368,7 +368,7 @@ impl Node {
             let file_offset = i as u64 * chunk_size as u64;
             let uncompressed_size = if i == self.inode.child_count() - 1 {
                 (self.inode.size() as u64)
-                    .checked_sub((chunk_size * i) as u64)
+                    .checked_sub(chunk_size as u64 * i as u64)
```
changweige, line 371: How about make `i` as u64 defined?

changweige (approved):
LGTM

PR #664, opened 2022-08-10
storage: use dyn BlobChunkInfo in storage layer

changweige:
Thus not converting chunk types everywhere. Only use BlobChunkInfo when processing backend IO. BlobIoChunk is
only used to pass backend IO requests from Rafs layer.

### DIFF
--- a/storage/src/device.rs
+++ b/storage/src/device.rs
@@ -411,11 +411,8 @@ pub trait BlobChunkInfo: Any + Sync + Send {
 #[derive(Clone)]
 pub enum BlobIoChunk {
     // For rafs v6 to pass chunk info to storage module.
-    // (blob_index, chunk_index)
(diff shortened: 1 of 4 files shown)


storage/src/device.rs
```diff
@@ -423,8 +423,15 @@ impl BlobIoChunk {
...
+            BlobIoChunk::Base(v) => v.as_ref(),
+            _ => panic!(),
+        }
+    }
+
+    pub fn inner(&self) -> Arc<dyn BlobChunkInfo> {
+        match self {
+            BlobIoChunk::Base(v) => v.clone(),
+            // TODO: Don't panic
             _ => panic!(),
```
imeoer, line 435: Better to give a message for panic?

changweige: Sure. Add a message

imeoer (approved):

PR #669, opened 2022-08-11
storage/backend: use rate log to prevent tons of logs

changweige:
Stop printinging a warning message for each fallback IO.

### DIFF
--- a/storage/src/backend/connection.rs
+++ b/storage/src/backend/connection.rs
@@ -3,19 +3,19 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! Help library to manage network connections.
+use std::cell::RefCell;
 use std::collections::HashMap;
-use std::io::Read;
-use std::io::Result;
+use std::io::{Read, Result};
 use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::thread;
-use std::time::{Duration, Instant};
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
-use reqwest::header::HeaderMap;
 use reqwest::{
     self,
     blocking::{Body, Client, Response},
+    header::HeaderMap,
     redirect::Policy,
     Method, StatusCode, Url,
 };
@@ -23,6 +23,12 @@ use reqwest::{
 use nydus_api::http::{OssConfig, ProxyConfig, RegistryConfig};
 
 const HEADER_AUTHORIZATION: &str = "Authorization";
+
+const RATE_LIMITED_LOG_TIME: u8 = 2;
+
+thread_local! {
+    pub static LAST_FALLBACK_AT: RefCell<SystemTime> = RefCell::new(UNIX_EPOCH);
+}
 
 /// Error codes related to network communication.
 #[derive(Debug)]
@@ -41,8 +47,8 @@ type ConnectionResult<T> = std::result::Result<T, ConnectionError>;
 pub(crate) struct ConnectionConfig {
     pub proxy: ProxyConfig,
     pub skip_verify: bool,
-    pub timeout: u64,
-    pub connect_timeout: u64,
+    pub timeout: u32,
+    pub connect_timeout: u32,
     pub retry_limit: u8,
 }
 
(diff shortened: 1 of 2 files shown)


storage/src/backend/connection.rs
```diff
@@ -215,12 +220,23 @@ impl Connection {
...
                         let _ = client
                             .get(ping_url.clone())
-                            .timeout(Duration::from_secs(connect_timeout))
+                            .timeout(Duration::from_secs(connect_timeout as u64))
                             .send()
                             .map(|resp| {
-                                proxy.health.set(is_success_status(resp.status()));
+                                let success = is_success_status(resp.status());
+                                if !success {
+                                    warn!(
```
imeoer, line 228: Is it still possible to print a lot of logs when the proxy server is consistently unhealthy? I think maybe only print the log when the health status changed.

storage/src/backend/connection.rs
```diff
@@ -290,7 +306,13 @@ impl Connection {
                 // fallback to origin server, the policy only applicable to non-upload operation
                 warn!("Request proxy server failed, fallback to original server");
             } else {
-                warn!("Proxy server is not healthy, fallback to original server");
+                LAST_LOGGING_TIME.with(|f| {
+                    let current = SystemTime::now();
+                    if current.duration_since(*f.borrow()).unwrap().as_secs() >= 2 {
+                        warn!("Proxy server is not healthy, fallback to original server");
```
imeoer, line 312: Is it possible to get all responses from the original server within 2 seconds, then without printing any warning?

storage/src/backend/connection.rs
```diff
@@ -24,6 +24,11 @@ use nydus_api::http::{OssConfig, ProxyConfig, RegistryConfig};
 
 const HEADER_AUTHORIZATION: &str = "Authorization";
 
+thread_local! {
+    pub static LAST_LOGGING_TIME: RefCell<SystemTime> = RefCell::new(SystemTime::now());
```
imeoer, line 28: Is it necessary to use a global variable, how about naming `LAST_FALLBACK_AT`?

storage/src/backend/connection.rs
changweige: good idea

storage/src/backend/connection.rs
changweige: The proxy is still unhealthy, so the warning can be printed after 2s. To ensure the first warning must be printed. I can make LAST_LOGGING_TIME start from EPOCH.

storage/src/backend/connection.rs
changweige: Changed.

storage/src/backend/connection.rs
```diff
@@ -210,17 +214,30 @@ impl Connection {
...
+                                    resp.status());
+                                } else if !last_success && success {
+                                    info!("Proxy recovered!")
+                                }
+                                last_success = success;
+                                proxy.health.set(success);
                             })
-                            .map_err(|_e| proxy.health.set(false));
+                            .map_err(|e| {
+                                warn!("Detected proxy unhealthy when ping proxy, {}", e);
```
imeoer, line 238: Ditto.

storage/src/backend/connection.rs
```diff
@@ -290,7 +307,13 @@ impl Connection {
                 // fallback to origin server, the policy only applicable to non-upload operation
                 warn!("Request proxy server failed, fallback to original server");
             } else {
-                warn!("Proxy server is not healthy, fallback to original server");
+                LAST_FALLBACK_AT.with(|f| {
+                    let current = SystemTime::now();
+                    if current.duration_since(*f.borrow()).unwrap().as_secs() >= 2 {
```
imeoer, line 312: Can we make it as constant?

changweige: fixed

storage/src/backend/connection.rs
changweige: fixed

storage/src/backend/connection.rs
```diff
@@ -210,17 +216,33 @@ impl Connection {
...
                             .send()
                             .map(|resp| {
-                                proxy.health.set(is_success_status(resp.status()));
+                                let success = is_success_status(resp.status());
+                                if last_success && !success {
+                                    warn!(
+                                    "Detected proxy unhealthy when pinging proxy, response status {}",
+                                    resp.status());
+                                } else if !last_success && success {
+                                    info!("Proxy recovered!")
```
imeoer, line 234: Maybe `Detected proxy recovered to health`.

imeoer (approved):

PR #678, opened 2022-08-16
builder: only reference the blob actually used in the chunk dict

imeoer:
During the build process, if a blob in the chunk dict is never used
for de-duplication, the blob is still referenced in the blob table
of final bootstrap.

This causes that the image will reference to some redundant blobs,
and these blobs will not be GC-ed properly.

This fixup is to add only the blobs actually used in the chunk dict
to blob table during the build process.


changweige (approved):
LGTM

Issue #667, opened 2022-08-10
Optimize entry search algorithm for rafs v6


changweige:
It would be better to implement binary search when searching an entry from directories. Now it is a linear search in the method `get_child_by_name()`

changweige:
In method `get_child_by_name`

changweige:
[yixiaojie0x](https://github.com/yixiaojie0x) is helping with this issue.

PR #683, opened 2022-08-17
rafs: optimize entry search in rafs v6

yixiaojie0x:
Implement binary search in rafs/src/metadata/direct_v6.rs
get_child_by_name() to accelerate entry search.

Signed-off-by: yixiaojie <yixiaojie@bytedance.com>

### DIFF
--- a/rafs/src/metadata/direct_v6.rs
+++ b/rafs/src/metadata/direct_v6.rs
@@ -18,6 +18,7 @@
 /// rule is to call validate() after creating any data structure from the on-disk bootstrap.
 use std::any::Any;
 use std::cell::{Cell, RefCell};
+use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::ffi::{OsStr, OsString};
 use std::fs::File;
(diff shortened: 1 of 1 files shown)


yixiaojie0x:
This PR is for #667 
It worked well functionally in my views, and it may need more tests/reviews. Any suggestions will be appreciated: )

rafs/src/metadata/direct_v6.rs
```diff
@@ -756,37 +757,63 @@ impl RafsInode for OndiskInodeWrapper {
         let blocks_count = div_round_up(inode.size(), EROFS_BLOCK_SIZE);
         let mut target: Option<u64> = None;
 
-        let nid = 'outer: for i in 0..blocks_count {
-            let head_entry = self.get_entry(i as usize, 0).map_err(err_invalidate_data)?;
-            // `e_nameoff` is offset from each single block?
+        // find target block
```
jiangliu, line 760: How about extracting this as a helper function?
And seems it may simplify by using usize instead of i32?

hsiangkao (approved):

rafs/src/metadata/direct_v6.rs
yixiaojie0x: Done.

jiangliu (approved):

PR #687, opened 2022-08-19
move to new fashion of rust-toolchain.toml

changweige:
Move to the new fashion pinning rust toolchain. It allows us to pin more than one target at the same time. Then we don't have to install rust components when building static nydus artifacts.


misc/musl-static/Dockerfile
```diff
@@ -2,9 +2,9 @@ FROM clux/muslrust:1.61.0
 
 ARG ARCH=x86_64
 
+RUN rustup component add clippy && rustup component add rustfmt && \
+  rustup target add $ARCH-unknown-linux-musl
```
imeoer, line 6: `$ARCH` is put in `RUN` instruction, does this work with `aarch64` or other archs?

rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+
+[toolchain]
+channel = "1.61.0"
+components = ["rustfmt", "clippy"]
+targets = ["x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl"]
```
imeoer, line 5: Are these targets must define here?

Issue #668, opened 2022-08-10
rotate nydusd log if it is configured to write to files


changweige:
By nydusd startup option `--log-file` , logs are written to a file. We'd better rotate it when it exceeds a certain size.

imeoer:
Typically nydusd is called by the control component (snapshotter/crio storage plugin), my suggestion is that nydusd should only support directing logs to stdout/stderr to allow the control component more flexibility in aggregating log timeline, filtering by subsystem, and rotating logs.

changweige:
I suppose it's a user's decision how to transport and persist nydusd logs. Stdout/stderr  and regurlar log files are not conflicted with each other. Moreover, nydusd already provides a CLI option to configure it, we can't just remove the option `--log-file`.  In practice, AFAIK, many user choose to store logs into regular files.

imeoer:
In scenarios where nydusd is used separately, it can use its own log file, but when integrated into snapshotter, we should let use the snapshotter logger directly.

changweige:
Close it as it is implemented and merged

PR #689, opened 2022-08-19
add log rotation

sctb512:
add log rotation to nydusd and nydus-image,
default set to 0 (disable log rotation).

This PR is addressing #668 

### DIFF
--- a/src/bin/nydus-image/main.rs
+++ b/src/bin/nydus-image/main.rs
@@ -545,7 +545,7 @@ fn init_log(matches: &ArgMatches) -> Result<()> {
     // Safe to unwrap because it has a default value and possible values are defined.
     let level = matches.value_of("log-level").unwrap().parse().unwrap();
 
-    setup_logging(log_file, level).context("failed to setup logging")
+    setup_logging(log_file, level, 0).context("failed to setup logging")
 }
 
 fn main() -> Result<()> {
--- a/src/bin/nydusd/main.rs
+++ b/src/bin/nydusd/main.rs
@@ -405,6 +405,15 @@ fn prepare_commandline_options() -> App<'static, 'static> {
                 .global(true),
         )
         .arg(
+            Arg::with_name("log-rotation-size")
+                .long("log-rotation-size")
+                .help("Specify log rotation size(MB), 0 to disable")
+                .default_value("0")
+                .takes_value(true)
+                .required(false)
+                .global(true),
+        )
+        .arg(
             Arg::with_name("rlimit-nofile")
                 .long("rlimit-nofile")
                 .short("R")
@@ -715,8 +724,13 @@ fn main() -> Result<()> {
     // Safe to unwrap because it has default value and possible values are defined
     let level = args.value_of("log-level").unwrap().parse().unwrap();
     let apisock = args.value_of("apisock");
+    let rotation_size = args
+        .value_of("log-rotation-size")
+        .unwrap()
+        .parse::<u64>()
+        .map_err(|e| einval!(format!("Invalid log rotation size: {}", e)))?;
 
-    setup_logging(logging_file, level)?;
+    setup_logging(logging_file, level, rotation_size)?;
 
     dump_program_info(crate_version!());
     handle_rlimit_nofile_option(&args, "rlimit-nofile")?;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -438,6 +438,7 @@ checksum = "33ab94b6ac8eb69f1496a6993f26f785b5fd6d99b7416023eb2a6175c0b242b1"
 dependencies = [
  "atty",
  "chrono",
+ "flate2",
  "glob",
  "lazy_static",
  "log",
--- a/app/Cargo.toml
+++ b/app/Cargo.toml
@@ -15,7 +15,7 @@ built = { version = "=0.4.3", features = ["chrono", "git2"] }
 [dependencies]
 # pin regex to fix RUSTSEC-2022-0013
 regex = "1.5.5"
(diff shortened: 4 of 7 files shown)


sctb512:
PTAL @changweige

src/bin/nydus-image/main.rs
```diff
@@ -532,6 +532,16 @@ fn prepare_cmd_args(bti_string: String) -> ArgMatches<'static> {
                 .required(false)
                 .global(true),
         )
+        .arg(
+            Arg::with_name("log-rotate-size")
+                .long("log-rotate-size")
```
changweige, line 537: There is no need to bring log rotation to `nyuds-image` since it is always one-time execution.

src/bin/nydusd/main.rs
```diff
@@ -404,6 +404,16 @@ fn prepare_commandline_options() -> App<'static, 'static> {
                 .required(false)
                 .global(true),
         )
+        .arg(
+            Arg::with_name("log-rotate-size")
+                .long("log-rotate-size")
+                .short("r")
```
changweige, line 410: As log rotation is not always needed, we'd better not give it a shore argument `-r`. It can reserved for other potential arguments.

src/bin/nydus-image/main.rs
```diff
@@ -544,8 +554,13 @@ fn init_log(matches: &ArgMatches) -> Result<()> {
...
     // Safe to unwrap because it has a default value and possible values are defined.
     let level = matches.value_of("log-level").unwrap().parse().unwrap();
+    let rotate_size = matches
+        .value_of("log-rotate-size")
+        .unwrap()
+        .parse::<u64>()
+        .unwrap();
 
-    setup_logging(log_file, level).context("failed to setup logging")
+    setup_logging(log_file, level, rotate_size).context("failed to setup logging")
```
changweige, line 563: rotate should be a verb while rotation is a noun. So better to rename it to rotation_size

src/bin/nydus-image/main.rs
```diff
@@ -544,8 +554,13 @@ fn init_log(matches: &ArgMatches) -> Result<()> {
 
     // Safe to unwrap because it has a default value and possible values are defined.
     let level = matches.value_of("log-level").unwrap().parse().unwrap();
+    let rotate_size = matches
+        .value_of("log-rotate-size")
```
changweige, line 558: Better to rename it to `log-rotation-size`

src/bin/nydus-image/main.rs
```diff
@@ -544,8 +554,13 @@ fn init_log(matches: &ArgMatches) -> Result<()> {
...
+        .unwrap()
+        .parse::<u64>()
+        .unwrap();
```
changweige, line 561: In Rust, we don't quickly unwrap. If the input CLI argument can't be parsed to integer from string. Returning error is better.

src/bin/nydusd/main.rs
```diff
@@ -715,8 +725,13 @@ fn main() -> Result<()> {
     // Safe to unwrap because it has default value and possible values are defined
     let level = args.value_of("log-level").unwrap().parse().unwrap();
     let apisock = args.value_of("apisock");
+    let rotate_size = args
+        .value_of("log-rotate-size")
```
changweige, line 729: log-rotation-size is better

src/bin/nydus-image/main.rs
sctb512: Ok, understand.

changweige:
Otherwise, looks good to me. Many thanks for contributing to this feature. It benefits infra maintenance

src/bin/nydus-image/main.rs
sctb512: Ok, i have removed it.

src/bin/nydus-image/main.rs
sctb512: You are right, i have renamed all "rotate".

src/bin/nydus-image/main.rs
changweige: I prefer returning an error from here and giving user an explicit message

sctb512: Got it.

sctb512: Done, please take a look.

src/bin/nydusd/main.rs
```diff
@@ -715,8 +724,17 @@ fn main() -> Result<()> {
...
     let level = args.value_of("log-level").unwrap().parse().unwrap();
     let apisock = args.value_of("apisock");
+    let rotation_size = match args.value_of("log-rotation-size").unwrap().parse::<u64>() {
+        Ok(size) => size,
+        Err(error) => {
+            return Err(Error::new(
+                ErrorKind::InvalidData,
+                format!("Invalid log rotation size: {}", error),
+            ));
+        }
```
changweige, line 734: match the result is fine. But a neater style will be:
```
    // Safe to unwrap since it has a default value.
    let rotation_size = args
        .value_of("log-rotation-size")
        .unwrap()
        .parse::<u64>()
        .map_err(|e| einval!(format!("Invalid log rotation size: {}", e)))?;
```

changweige:
Better to add a prefix "nydusd: addd log rotation" to respect the commit log style.
In addition, the commit log says it also adds log rotation to nydus-image. It should be removed.

src/bin/nydusd/main.rs
sctb512: Ok, I learned.

imeoer (approved):
LGTM

changweige (approved):

PR #692, opened 2022-08-23
nydus-test framework supports fscache mode

changweige:
1. Improve current framework to support start nydusd as fscache daemon
2. add test case for nydusd as fscache daemon


hsiangkao:
Excellent!

imeoer (approved):

Issue #693, opened 2022-08-23
Nydus-test found an error during executing fscache mode test


changweige:
The test image is exactly the same with fuse mode nydusd while fuse mode can pass the case

kernel info:
```
Linux n227-089-202 5.19.0-rc4 #2 SMP PREEMPT_DYNAMIC Thu Jun 30 17:33:40 CST 2022 x86_64 GNU/Linux

```

```
[2022-08-23 09:21:25,948] ERROR [workload_gen - 457:io_read] - Stress read failure, [Errno 5] Input/output error
Traceback (most recent call last):
  File "/data00/home/gechangwei/git_repo/nydus-rs/contrib/nydus-test/framework/workload_gen.py", line 454, in io_read
    cnt, size, duration = self.read_collected_files(io_duration)
  File "/data00/home/gechangwei/git_repo/nydus-rs/contrib/nydus-test/framework/workload_gen.py", line 417, in read_collected_files
    raise exc
  File "/data00/home/gechangwei/git_repo/nydus-rs/contrib/nydus-test/framework/workload_gen.py", line 411, in read_collected_files
    buf = f.read(io_size)
OSError: [Errno 5] Input/output error
[2022-08-23 09:21:26.115574 +08:00] ERROR [error/src/error.rs:21] Error:
	"entry not found index 215 infos.len 216"
	at storage/src/meta/mod.rs:492
	note: enable `RUST_BACKTRACE=1` env to display a backtrace
[2022-08-23 09:21:26.115631 +08:00] ERROR [src/bin/nydusd/fs_cache.rs:707] fscache: failed to read data from blob object: Invalid argument (os error 22)
[2022-08-23 09:21:26.117981 +08:00] ERROR [error/src/error.rs:21] Error:
	"entry not found index 215 infos.len 216"
	at storage/src/meta/mod.rs:492
	note: enable `RUST_BACKTRACE=1` env to display a backtrace
[2022-08-23 09:21:26.118024 +08:00] ERROR [src/bin/nydusd/fs_cache.rs:707] fscache: failed to read data from blob object: Invalid argument (os error 22)
[2022-08-23 09:21:26,120] ERROR [workload_gen - 415:read_collected_files] - file /data00/home/gechangwei/nydus-test-yard/rafs_mnt/DIR.0.1/DIR.1.1/DIR.2.4/DIR.3.2/DIR.4.0/DfxkrGkkjXoCCH.regular, offset 16621, io size 1048576
[...]

PR #703, opened 2022-08-30
storage: fix blob batch read error

kevinXYin:
this pr will fix https://github.com/dragonflyoss/image-service/issues/693

### DIFF
--- a/storage/src/meta/mod.rs
+++ b/storage/src/meta/mod.rs
@@ -489,6 +489,9 @@ impl BlobMetaInfo {
                 }
             }
 
+            if last_end > end {
+                return Ok(vec);
+            }
             Err(einval!(format!(
                 "entry not found index {} infos.len {}",
                 index,


hsiangkao (approved):

storage/src/meta/mod.rs
```diff
@@ -489,6 +489,9 @@ impl BlobMetaInfo {
                 }
             }
 
+            if last_end > end {
```
changweige, line 492: Should compare `last_end` with `batch_end` ? And `last_end` == `last_end` should be valid

hsiangkao: let mut last_end = entry.aligned_uncompressed_end();

hsiangkao: @kevinXYin

kevinXYin: in while loop already compare last_end with batch_end , for https://github.com/dragonflyoss/image-service/issues/693 issue last chunk's aligned_uncompressed_end not reach the batch_end which is state.uncompressed_size. And I am not sure is this valid (state.uncompressed_size > last chunk's aligned_uncompressed_end) ? But I think just checking the last_end to cover the end is enough.  And , yes last_end == end should be valid , I will send another pr to change to `if last_end >= end` , does this make sense ? @changweige @hsiangkao

changweige: The root cause of this problem is blob's uncompressed size is wrongly calculated. I have raised a PR to fix the image format. https://github.com/dragonflyoss/image-service/pull/712

Issue #704, opened 2022-08-30
Nydusd should clean up the API unix socket file in a normal shutdown


changweige:
We'd better not leave a unused and residual unix socket file in local file system

PR #706, opened 2022-08-30
nydusd: remove unix socket before exiting

sctb512:
remove unix socket before exiting,
it also works when getting SIGINT & SIGTERM signals.

This PR addressing #704 

### DIFF
--- a/src/bin/nydusd/api_server_glue.rs
+++ b/src/bin/nydusd/api_server_glue.rs
@@ -443,5 +443,8 @@ impl ApiServerController {
                 );
             }
         }
+        if let Some(apisock) = self.sock.as_ref() {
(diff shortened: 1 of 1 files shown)


src/bin/nydusd/api_server_glue.rs
```diff
@@ -443,5 +443,8 @@ impl ApiServerController {
                 );
             }
         }
+        if let Some(apisock) = self.sock.as_ref() {
+            std::fs::remove_file(apisock).unwrap_or_default();
```
imeoer, line 447: Can we give a warning if any error happens?

sctb512: ok, resolved.

PR #708, opened 2022-08-31
cargo: update fuse-backend-rs dependency

bergwolf:
Update it to 0.9.6 to get a fix for vfs no_opendir option handling.


changweige:
Is the being fixed bug only affecting nydusd as passthroughfs?

bergwolf:
> Is the being fixed bug only affecting nydusd as passthroughfs?

Yes.

changweige (approved):
LGTM

PR #709, opened 2022-08-31
api: rename blob cache entry config's fs_prefetch

changweige:
In fact, it is configuring how a single blob is being prefetched.

### DIFF
--- a/api/src/http.rs
+++ b/api/src/http.rs
@@ -281,9 +281,6 @@ pub struct BlobCacheEntry {
     /// Domain id for the blob, which is used to group cached blobs into management domains.
     #[serde(default)]
     pub domain_id: String,
-    /// Deprecated: data prefetch configuration: BlobPrefetchConfig.
-    #[serde(default)]
-    pub fs_prefetch: Option<BlobPrefetchConfig>,
 }
 
 /// Configuration information for a list of cached blob objects.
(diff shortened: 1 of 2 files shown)


api/src/http.rs
```diff
@@ -289,7 +289,7 @@ pub struct BlobCacheEntry {
     pub domain_id: String,
     /// Deprecated: data prefetch configuration: BlobPrefetchConfig.
     #[serde(default)]
-    pub fs_prefetch: Option<BlobPrefetchConfig>,
+    pub blob_prefetch: Option<BlobPrefetchConfig>,
```
imeoer, line 292: Seems the field has been deprecated?

changweige: I suppose the deprecated one is BlobPrefetchConfig sturct ?

imeoer: I think it says that we should use `blob_config.prefetch_config` instead of `fs_prefetch`.

changweige: `BlobCacheEntry` is newly introduced and v2.1 is not released. Why deprecate a never released config?

imeoer: cc @jiangliu PTAL.

jiangliu: I would like to remove this field and makes use of the `blob_config.prefetch_config`. When I was refactering the code, `fs_prefetch` is kept for backward compatibility. Now we are going to change its name, so breaking backward compatibility. Then how about just removing it?

changweige: Agree :-)

changweige: Um. I suppose it won't break backward compatibility since the struct `BlobCacheEntry` is added in API/v2. We can just safely remove `fs_prefetch` from `BlobCacheEntry`

changweige: @jiangliu It is removed cc @imeoer

imeoer (approved):
LGTM.

PR #710, opened 2022-09-01
nydusify: abort oss blob upload if conversion failed

imeoer:
We have used `PutObject` to upload oss before, so if there is any failure during
conversion process, it will cause the uploaded blob to be left on oss, and these
blobs are hard to be GC-ed, so we need always to use the multipart upload, and should
call the `AbortMultipartUpload` method to prevent blob residue as much as possible
once any error happens during conversion process.

### DIFF
--- a/contrib/nydusify/pkg/backend/oss.go
+++ b/contrib/nydusify/pkg/backend/oss.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/aliyun/aliyun-oss-go-sdk/oss"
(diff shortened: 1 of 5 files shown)


contrib/nydusify/pkg/backend/oss.go
```diff
@@ -23,16 +24,26 @@ import (
 )
 
 const (
-	splitPartsCount = 4
-	// Blob size bigger than 100MB, apply multiparts upload.
-	multipartsUploadThreshold = 100 * 1024 * 1024
+	// We always use multipart upload for OSS, and limit the
+	// multipart chunk size to 500MB.
+	multipartChunkSize = 500 * 1024 * 1024
```
bergwolf, line 29: Why is the multipart chunk size increased? 500MB seems pretty large for a single upload stream.

contrib/nydusify/pkg/backend/oss.go
```diff
@@ -134,102 +131,119 @@ func (b *OSSBackend) Upload(ctx context.Context, blobID, blobPath string, size i
...
+					return errors.Wrapf(err, "calculate crc64")
+				}
 
-			err = <-crc64ErrChan
-			if err != nil {
-				return nil, errors.Wrapf(err, "calculate crc64")
-			}
+				if uploadedCrc != ms.crc64 {
+					return errors.Errorf("crc64 mismatch, uploaded=%d, expected=%d", uploadedCrc, ms.crc64)
+				}
```
bergwolf, line 236: Maybe not related to this PR, but what happens if crc64 does not match? Do we clean up the OSS object?

contrib/nydusify/pkg/backend/oss.go
imeoer: Updated comment, for multipart uploads, OSS has a maximum number of 10000 chunks, so we can only upload blob sizes of about 1T (10000 * 100MB) if we use `100MB` chunk, maybe adjusting to `200MB`? Anyway, it's ok for me.

contrib/nydusify/pkg/backend/oss.go
imeoer: We can only get the crc64 property of the OSS object after the upload is complete, if crc64 doesn't match, we can't actually do anything (we can't delete the object), only can return an error and stop pushing the image manifest.

In fact, I think the crc64 check is no longer needed, we are long-term use in production has not shown crc64 mismatches, so keeping it would only consume resources. We can add a UT to ensure that there is no data corruption with concurrent multipart uploads.

imeoer:
@bergwolf ping~

bergwolf (approved):

Issue #714, opened 2022-09-07
fscache e2e startup performance


kevinXYin:
The fscache & cachefiles framework use dio for local cache files filling and reading. In my environment , I found this may affects the e2e startup performance.

I have tried to change to use buffer IO , and I can see a performance boost. But I think we need a more formal scheme so we can go upstream.

Here is my test resault：
`{
  "type": "bootstrap",
  "config": {
    "backend_type": "registry",
    "backend_config": {
      "scheme": "https",
      "timeout": 3,
      "connect_timeout": 3,
      "retry_limit": 2
    },
    "cache_type": "fscache"
  },
  "mode":"direct",
  "digest_validate": false,
  "iostat_files": false,
  "enable_xattr": true,
  "fs_prefetch": {
      "enable":true,
      "threads_count": 5,
      "merging_size": 1048576,
      "prefetch_all": true
  }
}`
![xSuyHVqfu5](https://user-images.githubusercontent.com/20159251/188835953-e04f13d1-670e-40d7-8031-c5311a9ec427.jpg)

PR #717, opened 2022-09-08
kernel-patches: introduce kernel patch for fscache ondemand IO path o…

kevinXYin:
This is kernel patch for https://github.com/dragonflyoss/image-service/issues/714

before we figure out a formal solution for upstream, we should fix this performance issue , that may affect the user experience first.

### DIFF
--- a/contrib/kernel-patches/0001-cachefiles-optimize-on-demand-IO-path-with-buffer-IO.patch
+++ b/contrib/kernel-patches/0001-cachefiles-optimize-on-demand-IO-path-with-buffer-IO.patch
@@ -1,0 +1,132 @@
+From 304939a8dca54edd9833b27f1ca48435ade2ed49 Mon Sep 17 00:00:00 2001
+From: Xin Yin <yinxin.x@bytedance.com>
+Date: Thu, 8 Sep 2022 10:52:08 +0800
+Subject: [PATCH] cachefiles: optimize on-demand IO path with buffer IO
+
+The cachefiles framework use dio for local cache files filling
+and reading, which may affects the performance for on-demand IO
+path.
+
+Change to use buffer IO for cache files filling, and first try
+to find data in the pagecache during cache files reading. After
+the pagecache for cache files is recycled, we will not suffer from
+double caching issue.
+
+Signed-off-by: Xin Yin <yinxin.x@bytedance.com>
+---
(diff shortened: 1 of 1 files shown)


hsiangkao:
Yeah, that is a practical approach.
The Linux community is generally reluctant to mix buffered and directed I/O, so we have to implement a direct I/O approach for upstream instead.

contrib/kernel-patches/0001-cachefiles-optimize-on-demand-IO-path-with-buffer-IO.patch
```diff
@@ -0,0 +1,132 @@
...
++			unsigned int offset = iocb->ki_pos & ~PAGE_MASK;
++			unsigned int bytes = min_t(loff_t, end - iocb->ki_pos,
++						   PAGE_SIZE - offset);
++			unsigned int copied;
++
++			if (page->index * PAGE_SIZE >= end)
++				break;
++
++			if (!PageUptodate(page)) {
++				error = -EFAULT;
```
hsiangkao, line 72: -EIO ?

kevinXYin: This will not cause EIO , if we can not find all data in pagecache , will fallback to dio for the rest iter

hsiangkao: okay, anyway, the error code is somewhat confusing to me... ;)

hsiangkao (approved):

contrib/kernel-patches/0001-cachefiles-optimize-on-demand-IO-path-with-buffer-IO.patch
kevinXYin: yeah , this is a bit confusing since we don't return result codes here , how about just set this like err = -1 ?

hsiangkao: let's leave as it for now...

Park-Jiyeonn:
Hi @kevinXYin, sorry for digging up this old PR. Just curious, was this patch ever used in ByteDance production? If so, could you share roughly at what scale? Thx!

PR #718, opened 2022-09-08
storage: fix compile error

wllenyj:
Compiling storage alone will fail.
```
error[E0432]: unresolved import `tokio::time`
  --> storage/src/cache/worker.rs:12:12
   |
12 | use tokio::time::interval;
   |            ^^^^ could not find `time` in `tokio`

error[E0432]: unresolved import `tokio::time`
  --> storage/src/factory.rs:24:5
   |
24 |     time,
   |     ^^^^ no `time` in the root

```

Signed-off-by: wanglei01 <wllenyj@linux.alibaba.com>


bergwolf (approved):
Thanks!

PR #722, opened 2022-09-16
nydusd: rename daemon subcommand to singleton

bergwolf:
So that it is clear that the subcommand means to run nydusd in a global single instance mode.

### DIFF
--- a/src/bin/nydusd/main.rs
+++ b/src/bin/nydusd/main.rs
@@ -324,8 +324,8 @@ fn append_virtiofs_subcmd_options(app: App<'static, 'static>) -> App<'static, 's
 }
 
 fn append_services_subcmd_options(app: App<'static, 'static>) -> App<'static, 'static> {
-    let subcmd = SubCommand::with_name("daemon")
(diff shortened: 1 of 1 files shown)


src/bin/nydusd/main.rs
```diff
@@ -324,8 +324,8 @@ fn append_virtiofs_subcmd_options(app: App<'static, 'static>) -> App<'static, 's
 }
 
 fn append_services_subcmd_options(app: App<'static, 'static>) -> App<'static, 'static> {
-    let subcmd = SubCommand::with_name("daemon")
-        .about("Run as a global daemon hosting multiple blobcache/fscache/virtiofs services.")
+    let subcmd = SubCommand::with_name("singleton")
```
imeoer, line 327: Maybe should update together with snapshotter: https://github.com/containerd/nydus-snapshotter/blob/0f88fb6a4f190c01bab42aba2bd47aca784546c2/pkg/process/manager.go#L318

imeoer (approved):

PR #728, opened 2022-09-18
rafs: introduce iterator to walk inode of a rafs filesystem

jiangliu:
Introduce iterator to walk inode of a rafs filesystem.


rafs/src/lib.rs
```diff
@@ -224,9 +227,69 @@ impl dyn RafsIoRead {
     }
 }
 
+/// A trait to unpack a Rafs filesystem to a tarball or directory.
+pub trait RafsUnpack {
```
imeoer, line 231: Is any place using it?

jiangliu: Sorry, I'm refactor `nydus-image pack` and this trait should be introduced later.

imeoer (approved):

PR #732, opened 2022-09-19
nydusd: fix a bug in handling singleton subcommand

jiangliu:
Let the singleton daemon exits when received SIGINT.

### DIFF
--- a/src/bin/nydusd/main.rs
+++ b/src/bin/nydusd/main.rs
@@ -142,6 +142,7 @@ impl DaemonController {
     fn shutdown(&self) {
         // Marking exiting state.
         self.active.store(false, Ordering::Release);
+        DAEMON_CONTROLLER.set_singleton_mode(false);
         // Signal the `run_loop()` working thread to exit.
         let _ = self.waker.wake();
 


hsiangkao (commented):
```
impl DaemonController {
    fn new() -> Self {
        let poller = Poll::new().expect("Failed to create `ServiceController` instance");
        let waker = Waker::new(poller.registry(), Token(1))
            .expect("Failed to create waker for ServiceController");
        Self {
            active: AtomicBool::new(true),
            singleton_mode: AtomicBool::new(true),  // is there a bug as well?
```

jiangliu:
> ```
> impl DaemonController {
>     fn new() -> Self {
>         let poller = Poll::new().expect("Failed to create `ServiceController` instance");
>         let waker = Waker::new(poller.registry(), Token(1))
>             .expect("Failed to create waker for ServiceController");
>         Self {
>             active: AtomicBool::new(true),
>             singleton_mode: AtomicBool::new(true),  // is there a bug as well?
> ```

Eventually we should set it to false by default. But the `singleton` is not fully functional yet, so keep it as is temporarily.

hsiangkao (approved):

PR #734, opened 2022-09-20
action: enable fuse + rafsv5 integration test

bergwolf:
We need to validate fuse + rafsv5 as well.


bergwolf:
ref: https://github.com/bergwolf/image-service/actions/runs/3087280748

jiangliu (approved):

Issue #716, opened 2022-09-08
Nydusd accesses Dragonfly in a mirror registry fashion


changweige:
At present, nydusd can only access dragonfly as an HTTP proxy. It has some cons:
1. For HTTPS registry, must provide a copy of CA and private key to dragonfly dfdaemon.
2. Hardly support multiple mirrors
3. An extra HTTP connect is needed (a more round trip)

My proposal is to support accessing dragonfly as a pull-through cache.
To achieve it:
1. add more items to nydusd configuration file:
2. support filling HTTP header according to a configuration JSON file.

```
{
  "device": {
    "backend": {
      "type": "registry",
      "config": {
        "scheme": "http",
        "repo": "nydus",
        "host": "127.0.0.1:8000",
        "mirrors": [{"scheme":"https","host": "dragonfrly.dfdaemon.org:65001", "username": "itsme", "password": "PASS", "header":"X-dragonfaly-header"}]
      }
    },

```

Furthermore, we can try to configure multiple registries to nydusd when it works globally in snapshotter shared mode

imeoer:
As per the offline discussion with @changweige, nydus can integrate dragonfly in two ways, proxy mode and mirror mode.

1. proxy mode integration documentation: https://d7y.io/zh/docs/setup/runtime/containerd/proxy/

For example, for https image registry, we need to configure the https decryption certificate for d7y so that it can cache the blob data, but there is a workaround without configuring the certificate:

- For nydusd, set to `http` scheme for the source image registry.
- For d7y dfdaemon, enable `useHTTPS` option, when d7y dfdaemon intercepts the http request, it will automatically replace the http with https to request the source image registry center, thus eliminating the need to decrypt https.

This workaround needs to take into account fallback, when d7y has down, nydusd needs to fallback to the source image registry of https, this feature still needs to be implemented.

2. mirror mode integration documentation: https://d7y.io/zh/docs/setup/runtime/containerd/mirror

Currently nydus can only support single mirror (d7y fixed configuration mirror url), but not support dynamic multi-mirrors, a feasible design to support multi-mirrors is as follows:

For example, the mirror requirement is:

- `ctr pull A.com` will be mirrored to B.com
- `ctr pull C.com` will be mirrored to D.com

Then when `ctr pull A.com`, the nydus snapshotter should generate the `config.json` file for nydusd like this:

``` json
{
  "device": {
    "backend": {
      "type": "registry",
      "config": {
        "scheme": "https",
[...]

PR #736, opened 2022-09-20
storage: add mirror support

sctb512:
Add mirror support for nydusd, 
corresponding configurations: device.backend.config.mirrors
```json
[ {
    "host": "http://dragonfly.io",
    "headers": {
       "X-Dragonfly-Registry": "https://docker.io",
    }
  },
]
```

this PR is addressing #716

### DIFF
--- a/api/src/http.rs
+++ b/api/src/http.rs
@@ -131,6 +131,8 @@ pub struct LocalFsConfig {
 pub struct OssConfig {
     /// Enable HTTP proxy for the read request.
     pub proxy: ProxyConfig,
+    /// Enable mirrors for the read request.
+    pub mirrors: Vec<MirrorConfig>,
     /// Skip SSL certificate validation for HTTPS scheme.
     pub skip_verify: bool,
     /// Drop the read request once http request timeout, in seconds.
@@ -181,6 +183,8 @@ pub struct RegistryConfig {
     pub scheme: String,
     /// Registry url host
     pub host: String,
+    /// Mirrors' configuration
+    pub mirrors: Vec<MirrorConfig>,
     /// Registry image name, like 'library/ubuntu'
     pub repo: String,
     /// Base64_encoded(username:password), the field should be
@@ -411,6 +415,15 @@ impl Default for ProxyConfig {
             check_interval: 5,
         }
     }
+}
+
+/// Configuration for mirror.
+#[derive(Clone, Deserialize, Serialize, Debug)]
+pub struct MirrorConfig {
+    /// Mirror url host.
+    pub host: String,
+    /// Request headers for mirror server.
+    pub headers: HashMap<String, String>,
 }
 
 #[derive(Debug)]
--- a/storage/src/backend/connection.rs
+++ b/storage/src/backend/connection.rs
@@ -7,13 +7,15 @@ use std::cell::RefCell;
 use std::collections::HashMap;
 use std::io::{Read, Result};
 use std::str::FromStr;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU8, Ordering};
 use std::sync::Arc;
 use std::thread;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
+use arc_swap::ArcSwapOption;
 use log::{max_level, Level};
 
+use reqwest::header::{HeaderName, HeaderValue};
 use reqwest::{
     self,
     blocking::{Body, Client, Response},
@@ -22,7 +24,8 @@ use reqwest::{
     Method, StatusCode, Url,
 };
 
-use nydus_api::http::{OssConfig, ProxyConfig, RegistryConfig};
+use nydus_api::http::{MirrorConfig, OssConfig, ProxyConfig, RegistryConfig};
+use url::ParseError;
 
 const HEADER_AUTHORIZATION: &str = "Authorization";
 
@@ -39,6 +42,10 @@ pub enum ConnectionError {
     ErrorWithMsg(String),
     Common(reqwest::Error),
     Format(reqwest::Error),
+    Url(ParseError),
+    Scheme,
+    Host,
+    Port,
 }
 
 /// Specialized `Result` for network communication.
@@ -48,6 +55,7 @@ type ConnectionResult<T> = std::result::Result<T, ConnectionError>;
 #[derive(Debug, Clone)]
 pub(crate) struct ConnectionConfig {
     pub proxy: ProxyConfig,
+    pub mirrors: Vec<MirrorConfig>,
     pub skip_verify: bool,
     pub timeout: u32,
     pub connect_timeout: u32,
@@ -58,6 +66,7 @@ impl Default for ConnectionConfig {
     fn default() -> Self {
         Self {
             proxy: ProxyConfig::default(),
+            mirrors: Vec::<MirrorConfig>::new(),
             skip_verify: false,
             timeout: 5,
             connect_timeout: 5,
@@ -70,6 +79,7 @@ impl From<OssConfig> for ConnectionConfig {
     fn from(c: OssConfig) -> ConnectionConfig {
         ConnectionConfig {
             proxy: c.proxy,
+            mirrors: c.mirrors,
             skip_verify: c.skip_verify,
             timeout: c.timeout,
             connect_timeout: c.connect_timeout,
@@ -82,6 +92,7 @@ impl From<RegistryConfig> for ConnectionConfig {
     fn from(c: RegistryConfig) -> ConnectionConfig {
         ConnectionConfig {
             proxy: c.proxy,
+            mirrors: c.mirrors,
             skip_verify: c.skip_verify,
             timeout: c.timeout,
             connect_timeout: c.connect_timeout,
@@ -181,7 +192,28 @@ pub(crate) fn respond(resp: Response, catch_status: bool) -> ConnectionResult<Re
 pub(crate) struct Connection {
     client: Client,
     proxy: Option<Proxy>,
+    mirror_state: MirrorState,
     shutdown: AtomicBool,
+}
+
+#[derive(Debug)]
+pub(crate) struct MirrorState {
+    mirrors: Vec<Arc<Mirror>>,
+    /// Current mirror, None if there is no mirror available.
+    current: ArcSwapOption<Mirror>,
+}
+
+#[derive(Debug)]
+pub(crate) struct Mirror {
+    client: Client,
+    /// Information for mirror from configuration file.
+    config: MirrorConfig,
+    /// Mirror status, it will be set to false by atomic operation when mirror is not work.
+    status: AtomicBool,
+    /// Falied times for mirror, the status will be marked as false when fail_times = fail_limit.
+    fail_times: AtomicU8,
+    /// Failed limit for mirror.
+    fail_limit: u8,
 }
 
 impl Connection {
@@ -189,6 +221,7 @@ impl Connection {
     pub fn new(config: &ConnectionConfig) -> Result<Arc<Connection>> {
         info!("backend config: {:?}", config);
         let client = Self::build_connection("", config)?;
+
         let proxy = if !config.proxy.url.is_empty() {
             let ping_url = if !config.proxy.ping_url.is_empty() {
                 Some(Url::from_str(&config.proxy.ping_url).map_err(|e| einval!(e))?)
@@ -203,9 +236,32 @@ impl Connection {
         } else {
             None
         };
+
+        let mut mirrors = Vec::new();
+        for mirror_config in config.mirrors.iter() {
+            if !mirror_config.host.is_empty() {
+                mirrors.push(Arc::new(Mirror {
+                    // TODO: build mirror connection from MirrorConfig, now is from registry config
+                    client: Self::build_connection("", config)?,
+                    config: mirror_config.clone(),
+                    status: AtomicBool::from(true),
+                    // Maybe read from configuration file
+                    fail_limit: 5,
+                    fail_times: AtomicU8::from(0),
+                }));
+            }
+        }
+
+        let current = if let Some(first_mirror) = mirrors.first() {
+            ArcSwapOption::from(Some(first_mirror.clone()))
+        } else {
+            ArcSwapOption::from(None)
+        };
+
         let connection = Arc::new(Connection {
             client,
             proxy,
+            mirror_state: MirrorState { mirrors, current },
             shutdown: AtomicBool::new(false),
         });
 
@@ -258,6 +314,7 @@ impl Connection {
                 });
             }
         }
+        // TODO: check mirrors' health
 
         Ok(connection)
     }
@@ -324,6 +381,86 @@ impl Connection {
                         f.replace(current);
                     }
                 })
+            }
+        }
+
+        let current_mirror = self.mirror_state.current.load();
+
+        if let Some(mirror) = current_mirror.as_ref() {
+            let data_cloned: Option<ReqBody<R>> = match data.as_ref() {
+                Some(ReqBody::Form(form)) => Some(ReqBody::Form(form.clone())),
+                Some(ReqBody::Buf(buf)) => Some(ReqBody::Buf(buf.clone())),
+                _ => None,
+            };
+
+            let mirror_host = Url::parse(&mirror.config.host).map_err(ConnectionError::Url)?;
+
+            let mut current_url = Url::parse(url).map_err(ConnectionError::Url)?;
+
+            current_url
+                .set_scheme(mirror_host.scheme())
+                .map_err(|_| ConnectionError::Scheme)?;
+            current_url
+                .set_host(mirror_host.host_str())
+                .map_err(|_| ConnectionError::Host)?;
+            current_url
+                .set_port(mirror_host.port())
+                .map_err(|_| ConnectionError::Port)?;
+
+            for (key, value) in mirror.config.headers.iter() {
+                headers.insert(
+                    HeaderName::from_str(key).unwrap(),
+                    HeaderValue::from_str(value).unwrap(),
+                );
+            }
+
+            let result = self.call_inner(
+                &mirror.client,
+                method.clone(),
+                current_url.to_string().as_str(),
+                &query,
+                data_cloned,
+                headers,
+                catch_status,
+                false,
+            );
+
+            match result {
+                Ok(resp) => {
+                    if resp.status() < StatusCode::INTERNAL_SERVER_ERROR {
+                        return Ok(resp);
+                    }
+                }
+                Err(err) => {
+                    warn!(
+                        "request mirror failed, mirror: {:?},  error: {:?}",
+                        mirror, err
+                    );
+                    mirror.fail_times.fetch_add(1, Ordering::Relaxed);
+
+                    if mirror.fail_times.load(Ordering::Relaxed) >= mirror.fail_limit {
+                        error!(
+                            "reach to fail limit {}, disable mirror: {:?}",
+                            mirror.fail_limit, mirror
+                        );
+                        mirror.status.store(false, Ordering::Relaxed);
+
+                        let mut idx = 0;
+                        loop {
+                            if idx == self.mirror_state.mirrors.len() {
+                                break None;
+                            }
+                            let m = &self.mirror_state.mirrors[idx];
+                            if m.status.load(Ordering::Relaxed) {
+                                break Some(m);
+                            }
+
+                            idx += 1;
+                        }
+                        .map(|m| self.mirror_state.current.store(Some(m.clone())))
+                        .unwrap_or_else(|| self.mirror_state.current.store(None));
+                    }
+                }
             }
         }
 
@@ -487,5 +624,6 @@ mod tests {
         assert!(config.proxy.fallback);
         assert_eq!(config.proxy.ping_url, "");
         assert_eq!(config.proxy.url, "");
+        assert!(config.mirrors.is_empty());
     }
 }
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -545,6 +545,13 @@ impl Registry {
         let id = id.ok_or_else(|| einval!("Registry backend requires blob_id"))?;
         let config: RegistryConfig = serde_json::from_value(config).map_err(|e| einval!(e))?;
         let con_config: ConnectionConfig = config.clone().into();
+
+        if !config.proxy.url.is_empty() && !config.mirrors.is_empty() {
+            return Err(einval!(
+                "connection: proxy and mirrors cannot be configured at the same time."
+            ));
+        }
+
         let retry_limit = con_config.retry_limit;
         let connection = Connection::new(&con_config)?;
         let auth = trim(config.auth);


api/src/http.rs
```diff
@@ -416,6 +418,27 @@ impl Default for ProxyConfig {
...
+    /// Dragonfaly url host.
+    pub host: String,
+    /// Request header for Dragonfaly server. Currently, only "X-Dragonfly-Registry" needed.
+    pub header: DragonfalyHeader,
+}
+
+/// Configuration information for Dragonfaly header.
+#[derive(Debug, Clone, Default, Deserialize, Serialize)]
+#[serde(default)]
+pub struct DragonfalyHeader {
```
changweige, line 436: Headers should be transparent to nydusd/backend. They are just extra KV information that nydus has to put together with each GET request. So it's better that we don't define any `Dragonfly` specific struct here.

api/src/http.rs
```diff
@@ -416,6 +418,27 @@ impl Default for ProxyConfig {
...
+/// Configuration information for mirror.
+#[derive(Debug, Clone, Default, Deserialize, Serialize)]
+#[serde(default)]
+pub struct MirrorConfig {
+    /// Dragonfaly http scheme, either 'http' or 'https'.
+    pub scheme: String,
+    /// Dragonfaly url host.
+    pub host: String,
+    /// Request header for Dragonfaly server. Currently, only "X-Dragonfly-Registry" needed.
+    pub header: DragonfalyHeader,
```
changweige, line 430: This should be headers  which can accommodate multiple KV lines

storage/src/backend/registry.rs
```diff
@@ -151,6 +165,14 @@ struct RegistryState {
     cached_redirect: HashCache,
 }
 
+#[derive(Debug, Clone)]
+struct MirrorConfigState {
+    // HTTP scheme like: https, http
+    scheme: String,
+    host: String,
+    registry_header: String,
```
changweige, line 173: This should be a HashMap<String, String> ?

storage/src/backend/registry.rs
```diff
@@ -394,18 +468,30 @@ impl RegistryReader {
...
+        let url = state.url(url.as_str(), &[]).map_err(RegistryError::Url)?;
         let mut headers = HeaderMap::new();
         let end_at = offset + buf.len() as u64 - 1;
         let range = format!("bytes={}-{}", offset, end_at);
         headers.insert("Range", range.parse().unwrap());
 
+        if !self.is_registry() {
+            let mirror_index = self.mirrors.index.load(Ordering::Relaxed) as usize;
+            let mirror = self.mirrors.states[mirror_index].clone();
+            headers.insert(
```
changweige, line 482: Header should be gathered from backend json configuration and directly transfer to HTTP requests header lines. Nydusd does not care about the heard Key name.

storage/src/backend/registry.rs
```diff
@@ -394,18 +468,30 @@ impl RegistryReader {
...
+                "X-Dragonfly-Registry",
+                HeaderValue::from_str(
+                    format!("https://{}", mirror.config.registry_header).as_str(),
+                )
+                .unwrap(),
+            );
+
+            // let data = self.get_state().auth;
```
changweige, line 490: Please remove dead code

storage/src/backend/registry.rs
```diff
@@ -282,10 +304,62 @@ struct RegistryReader {
     blob_id: String,
     connection: Arc<Connection>,
     state: Arc<RegistryState>,
+    mirrors: Arc<Mirror>,
     metrics: Arc<BackendMetrics>,
 }
 
+struct Mirror {
+    states: Vec<MirrorState>,
+    index: Arc<AtomicI64>,
```
changweige, line 313: Atomic value does not need Arc wrapper. Please remove the Arc wrapper

storage/src/backend/registry.rs
```diff
@@ -282,10 +304,62 @@ struct RegistryReader {
     blob_id: String,
     connection: Arc<Connection>,
     state: Arc<RegistryState>,
+    mirrors: Arc<Mirror>,
```
changweige, line 307: The field name is `mirrors`, but the type `Arc<Mirror>` does not indicate its multitude

storage/src/backend/registry.rs
sctb512: Yes, fixed.

api/src/http.rs
sctb512: fixed

storage/src/backend/registry.rs
```diff
@@ -282,10 +298,62 @@ struct RegistryReader {
...
 }
 
+struct MirrorState {
+    mirrors: Vec<Mirror>,
+    index: ArcSwap<Option<usize>>,
+}
+
+#[derive(Debug)]
+struct Mirror {
+    state: RegistryState,
```
imeoer, line 312: Why not use `state: Arc<RegistryState>`.

storage/src/backend/registry.rs
```diff
@@ -394,18 +462,35 @@ impl RegistryReader {
...
         headers.insert("Range", range.parse().unwrap());
 
+        if !self.is_registry() {
+            let mirror_index = self.mirror_state.index.load().unwrap();
+            let mirror = &self.mirror_state.mirrors[mirror_index];
+
+            for (key, value) in &mirror.config.headers {
+                let value =
+                    if key == "X-Dragonfly-Registry" && !value.starts_with(&self.state.scheme) {
+                        format!("{}://{}", self.state.scheme, value)
```
imeoer, line 480: Why replace the target registry scheme, I think user can specify a different one with `state.scheme`.

storage/src/backend/registry.rs
```diff
@@ -519,7 +602,26 @@ impl BlobReader for RegistryReader {
...
+            .map_err(|e| {
+                if option_index.is_some() {
+                    let index = option_index.unwrap();
+                    let state = &self.mirror_state.mirrors[index].state;
+                    let config = &self.mirror_state.mirrors[index].config;
+                    error!(
+                        "read filed, current state: {:?}, mirror config: {:?},  error: {:?}",
+                        state, config, e
+                    );
+                    self.set_mirror_status(index, false);
```
imeoer, line 616: If this mirror request failed, how can we retry again later?

storage/src/backend/registry.rs
imeoer, line 312: How about use `Option<RegistryState>`, only set to `Some` if current mirror works fine.

api/src/http.rs
```diff
@@ -413,6 +415,17 @@ impl Default for ProxyConfig {
     }
 }
 
+/// Configuration information for mirror.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct MirrorConfig {
+    /// Dragonfaly http scheme, either 'http' or 'https'.
```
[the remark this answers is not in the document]
sctb512: fixed.

storage/src/backend/registry.rs
```diff
@@ -570,11 +675,39 @@ impl Registry {
             blob_url_scheme: config.blob_url_scheme,
             blob_redirected_host: config.blob_redirected_host,
             cached_redirect: HashCache::new(),
-        });
+        };
+
+        let mut mirrors = Vec::<Mirror>::new();
+
+        for config in config.mirrors.into_iter() {
```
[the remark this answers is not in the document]
sctb512: Yes, fixed. It will return an error when configuring mirror and proxy at the same time.

storage/src/backend/registry.rs
```diff
@@ -570,11 +675,39 @@ impl Registry {
...
+            let mut mirror_state = registry_state.clone();
```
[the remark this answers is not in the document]
sctb512: fixed, renamed to `state`.

storage/src/backend/registry.rs
```diff
@@ -282,10 +298,62 @@ struct RegistryReader {
...
+struct MirrorState {
+    mirrors: Vec<Mirror>,
+    index: ArcSwap<Option<usize>>,
+}
+
+#[derive(Debug)]
+struct Mirror {
+    state: RegistryState,
+    config: MirrorConfig,
+    status: AtomicBool,
```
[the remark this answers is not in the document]
sctb512: Yes, added.

storage/src/backend/registry.rs
```diff
@@ -570,11 +675,39 @@ impl Registry {
...
+            mirrors.push(Mirror {
+                state: mirror_state,
+                config,
+                status: AtomicBool::from(true),
+            });
+        }
+
+        for (index, mirror) in mirrors.iter().enumerate() {
+            let state = &mirror.state;
+            info!(
```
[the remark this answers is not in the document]
sctb512: Already changed to `info!("mirrors: {:?}", mirrors);`

storage/src/backend/registry.rs
sctb512: This code is to resolve the case of `X-Dragonfly-Registry` without scheme in the configuration file. Do you mean the scheme has to be filled in here? (as containerd does)

sctb512: Add `state.scheme` as scheme only when there is no scheme?

api/src/http.rs
```diff
@@ -413,6 +415,17 @@ impl Default for ProxyConfig {
...
+    pub scheme: String,
```
changweige, line 422: Can use mirror registry or pull-through cache rather than Dragonfly? In addition, I think docker registry can also work together with nydusd with the mirror configuration

storage/src/backend/registry.rs
sctb512: Maybe it must be wrapped in ArcSwap `ArcSwap<Option<RegistryState>>` ?

api/src/http.rs
sctb512: Fixed.

storage/src/backend/registry.rs
sctb512: I think it may not be necessary to modify the value of `Mirror.state` because it contains state information. If mirror is restored (not implemented yet) the state needs to be regenerated.

storage/src/backend/registry.rs
sctb512: It will retry `state.retry_limit` times in read() method:
```rust
fn read(&self, buf: &mut [u8], offset: u64) -> BackendResult<usize> {
    let mut retry_count = self.retry_limit();
    let begin_time = self.metrics().begin();

    loop {
        match self.try_read(buf, offset) {
            Ok(size) => {
                self.metrics().end(&begin_time, buf.len(), false);
                return Ok(size);
            }
            Err(err) => {
                if retry_count > 0 {
                    warn!(
                        "Read from backend failed: {:?}, retry count {}",
                        err, retry_count
                    );
                    retry_count -= 1;
                } else {
                    self.metrics().end(&begin_time, buf.len(), true);
                    ERROR_HOLDER
                        .lock()
                        .unwrap()
                        .push(&format!("{:?}", err))
                        .unwrap_or_else(|_| error!("Failed when try to hold error"));
                    return Err(err);
                }
            }
        }
    }
}
```

storage/src/backend/registry.rs
```diff
@@ -282,10 +298,62 @@ struct RegistryReader {
     blob_id: String,
     connection: Arc<Connection>,
     state: Arc<RegistryState>,
+    mirror_state: Arc<MirrorState>,
     metrics: Arc<BackendMetrics>,
 }
 
+struct MirrorState {
```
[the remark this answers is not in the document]
sctb512: We do not have OSS and are not considering implementation at this time.

storage/src/backend/registry.rs
```diff
@@ -282,10 +298,65 @@ struct RegistryReader {
...
+    state: RegistryState,
+    /// Information for mirror from configuration file.
+    config: MirrorConfig,
+    /// Mirror status, it will be set to false by atomic operation when mirror is not work.
+    status: AtomicBool,
+}
+
 impl RegistryReader {
+    fn get_state(&self) -> RegistryState {
+        let mirrors = &self.mirror_state.mirrors;
```
changweige, line 322: It's better to return `Arc<RegistryState>`  to avoid copying memory

storage/src/backend/registry.rs
```diff
@@ -282,10 +298,65 @@ struct RegistryReader {
...
+                let mut new_index = index;
+                for (index, current_mirror) in mirrors.iter().enumerate() {
+                    if current_mirror.status.load(Ordering::Relaxed) {
+                        self.mirror_state.index.store(Arc::new(Some(index)));
+                        new_index = index;
+                        break;
+                    }
+                }
+                if new_index == index {
+                    self.mirror_state.index.store(Arc::new(None));
```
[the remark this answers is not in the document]
changweige: Maybe, we should leave an issue to track the following task implementing how to check the mirror's health status. 
For instance, periodically get repo's tags list.

storage/src/backend/registry.rs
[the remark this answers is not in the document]
imeoer: Dragonfly supports oss, and nydus will support aws s3 in the future.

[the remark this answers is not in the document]
imeoer: Can we implement the mirror in `connection.rs` ? Thus we can add any HTTP-based backend in the future.

storage/src/backend/connection.rs
```diff
@@ -258,6 +307,7 @@ impl Connection {
                 });
             }
         }
+        // TODO: check mirrors' healthy
```
changweige, line 310: check mirror's health

storage/src/backend/registry.rs
```diff
@@ -394,6 +408,7 @@ impl RegistryReader {
         offset: u64,
         allow_retry: bool,
     ) -> RegistryResult<usize> {
+        // let state = self.state;
```
changweige, line 411: Please trim dead code

storage/src/backend/connection.rs
```diff
@@ -327,6 +377,93 @@ impl Connection {
...
+                    );
+
+                    let fail_limit = mirror.fail_limit.load(Ordering::Relaxed);
+                    if fail_limit > 0 {
+                        mirror.fail_limit.store(fail_limit - 1, Ordering::Relaxed);
+                    }
+
+                    if fail_limit == 1 && mirror.fail_limit.load(Ordering::Relaxed) <= 0 {
+                        error!("reach to fail limit, disable mirror: {:?}", mirror);
+                        mirror.status.store(false, Ordering::Relaxed);
```
changweige, line 447: There still is race between load and store

storage/src/backend/connection.rs
```diff
@@ -327,6 +377,93 @@ impl Connection {
             }
         }
 
+        let current_mirror = self.mirror_state.current.load();
+
+        if current_mirror.is_some() {
```
changweige, line 382: Try this ?
```
if let Some(m) = current_mirror {}
```

storage/src/backend/connection.rs
sctb512: Fixed.

storage/src/backend/registry.rs
sctb512: Done.

storage/src/backend/connection.rs
sctb512: Already replaced it with `fetch_add()`.

jiangliu:
@changweige @sctb512 could you please help to rebase onto the latest master branch? It should help to solve the anolis CI failure issue:)

sctb512:
> @changweige @sctb512 could you please help to rebase onto the latest master branch? It should help to solve the anolis CI failure issue:)

Finished.

sctb512:
> @changweige @sctb512 could you please help to rebase onto the latest master branch? It should help to solve the anolis CI failure issue:)

Failed, it seems that `nydus-snapshotter` is not ready?

```
======== start nydus-snapshotter and config containerd for nydus
start-nydus-snapshotter-config-containerd: pass

======== run a container with nydus image
fetching sha256:fa8ee4f8... application/vnd.oci.image.manifest.v1+json
fetching sha256:0fb1e874... application/vnd.oci.image.config.v1+json
ctr-remote: failed to prepare extraction snapshot "extract-225012612-LpUf sha256:667a64a38375ff83228d26bc144263c3afd3ea54e071956a80e7ec127fda75aa": connection error: desc = "transport: Error while dialing dial unix /run/containerd/containerd-nydus-grpc.sock: connect: connection refused": unavailable
fail to pull docker.io/hsiangkao/ubuntu:20.04-rafs-v6 with ctr-remote
run-container-with-nydus-image: fail
root     3787488 3757889  0 14:33 ?        00:00:00 containerd-nydus-grpc --config-path /tmp/tone/run/nydus-fscache-citest/nydus-erofs-config.json --daemon-mode shared --daemon-backend fscache --log-level info --root /var/lib/containerd/io.containerd.snapshotter.v1.nydus --cache-dir /var/lib/nydus/cache --address /run/containerd/containerd-nydus-grpc.sock --nydusd-path /usr/local/bin/nydusd --log-to-stdout
/tmp/tone/tests/nydus-fscache-citest/run.sh: line 241: 3787488 Killed                  containerd-nydus-grpc --config-path $TONE_BM_RUN_DIR/nydus-erofs-config.json --daemon-mode shared --daemon-backend fscache --log-level info --root /var/lib/containerd/io.containerd.snapshotter.v1.nydus --cache-dir /var/lib/nydus/cache --address /run/containerd/containerd-nydus-grpc.sock --nydusd-path /usr/local/bin/nydusd --log-to-stdout > $TONE_BM_RUN_DIR/nydus-snapshotter.log 2>&1
merge-target-branch: pass
build-docker-image: pass
compile-nydus: pass
compile-ctr-remote: pass
compile-nydus-snapshotter: pass
start-nydus-snapshotter-config-containerd: pass
run-container-with-nydus-image: fail
--------------------------------------------------------------------------------
2022-09-24 14:33:23
/tmp/tone/bin/testinfo.sh nydus-fscache-citest
--------------------------------------------------------------------------------
```

storage/src/backend/registry.rs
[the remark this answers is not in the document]
sctb512: Yes, done.

jiangliu (approved):

storage/src/backend/connection.rs
```diff
@@ -327,6 +384,86 @@ impl Connection {
...
+                Some(ReqBody::Form(form)) => Some(ReqBody::Form(form.clone())),
+                Some(ReqBody::Buf(buf)) => Some(ReqBody::Buf(buf.clone())),
+                _ => None,
+            };
+
+            let mirror_host = Url::parse(&mirror.config.host).map_err(ConnectionError::Url)?;
+
+            let mut current_url = Url::parse(url).map_err(ConnectionError::Url)?;
+
+            current_url
```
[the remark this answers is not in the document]
imeoer: And why ignore the error `.map_err(|_|` ?

storage/src/backend/connection.rs
```diff
@@ -327,6 +384,86 @@ impl Connection {
...
+                &query,
+                data_cloned,
+                headers,
+                catch_status,
+                false,
+            );
+
+            match result {
+                Ok(resp) => {
+                    if resp.status() < StatusCode::INTERNAL_SERVER_ERROR {
```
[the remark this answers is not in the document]
imeoer: Should give a warning like `Failed to request mirror server, fallback to original server` if fallback happens.

storage/src/backend/connection.rs
```diff
@@ -203,9 +236,32 @@ impl Connection {
         } else {
             None
         };
+
+        let mut mirrors = Vec::new();
+        for mirror_config in config.mirrors.iter() {
+            if !mirror_config.host.is_empty() {
+                mirrors.push(Arc::new(Mirror {
+                    // TODO: build mirror connection from MirrorConfig, now is from registry config
+                    client: Self::build_connection("", config)?,
```
imeoer, line 245: it is unnecessary to build the client on initializing.

storage/src/backend/connection.rs
```diff
@@ -327,6 +384,86 @@ impl Connection {
...
+                        mirror.status.store(false, Ordering::Relaxed);
+
+                        let mut idx = 0;
+                        loop {
+                            if idx == self.mirror_state.mirrors.len() {
+                                break None;
+                            }
+                            let m = &self.mirror_state.mirrors[idx];
+                            if m.status.load(Ordering::Relaxed) {
+                                break Some(m);
```
imeoer, line 455: Maybe give a warning to notify user that we have changed the mirror server?

storage/src/backend/connection.rs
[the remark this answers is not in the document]
sctb512: > And why ignore the error `.map_err(|_|` ?

Because the return value of `set_scheme()` `set_host` and `set_port` is `Err(())`.

storage/src/backend/connection.rs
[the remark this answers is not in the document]
sctb512: > If `500` status happens, why not fallback to the original server?

If `500` status happens, it will fallback to the original server.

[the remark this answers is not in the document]
sctb512: > warning like `Failed to request mirror server, fallback to original server` if fallback happens.

Yes, fixed.

storage/src/backend/connection.rs
sctb512: Yes, Fixed.

storage/src/backend/connection.rs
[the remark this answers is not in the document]
sctb512: > Why do we need to replace `scheme, host, port` for every request?

Because the request URL is the origin registry, we should modify them.

storage/src/backend/connection.rs
sctb512: If every request built client to mirror, does it affect performance? 
Or do you mean to reuse the `state.client`? just like:
```
self.call_inner(
            &self.client,
...
```

api/src/http.rs
```diff
@@ -181,6 +183,8 @@ pub struct RegistryConfig {
     pub scheme: String,
     /// Registry url host
     pub host: String,
+    /// Mirrors' configuration
```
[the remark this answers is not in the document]
sctb512: I think the configuration is the same, but it hasn't been implemented yet.

storage/src/backend/connection.rs
imeoer: Yes, should reuse the same client.

storage/src/backend/connection.rs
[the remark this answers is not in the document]
imeoer: Can we set these on initializing instead of every request?

storage/src/backend/connection.rs
sctb512: Got it.

storage/src/backend/connection.rs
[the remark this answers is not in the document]
sctb512: The URL is generated with `RegistryState.url(path, query)`, but we can't get the mirror information here.

api/src/http.rs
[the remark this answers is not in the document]
sctb512: > Keep consistent with `OssConfig`? `Enable mirrors for the read request.`

Fixed.

storage/src/backend/connection.rs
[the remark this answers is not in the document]
imeoer: > If 500 status happens, it will fallback to the original server.

Is it true?

Issue #686, opened 2022-08-18
Lazy pulling of OCI image for Nydus


hsiangkao:
For more details, see new project soci-snapshotter from AWS
https://github.com/awslabs/soci-snapshotter

Actually such trick above is a bit tricky (such each seekable point needs 32KiB sliding window data), but I'm not sure if most users would like to reuse ociv1 blobs and append some indexes, so...

PR #737, opened 2022-09-21
[WIP] Add zlib/gzip stream indexing and random access library

zyfjeff:
Add zlib/gzip stream indexing and random access library
 
* bindings for the soci indexer
* add magic number for blob, avoid crashes due to invalid blob
* add test cases for soci indexer

#686 

Signed-off-by: zyfjeff <zyfjeff@linux.alibaba-inc.com>


hsiangkao:
I don't think it's related to SOCI. It's just another OCI random accessor.
Maybe Nydus-OCI blob?

jiangliu:
We hope to use a pure rust version:)

PR #743, opened 2022-09-23
Unify fusedev and virtiofs frontend

jiangliu:
Add fuse and virtiofs to daemon subcommand.

Signed-off-by: uran0sH <huangwenyuu@outlook.com>

### DIFF
--- a/src/bin/nydusd/main.rs
+++ b/src/bin/nydusd/main.rs
@@ -19,7 +19,6 @@ extern crate nydus_error;
 extern crate nydus_rafs as rafs;
 extern crate nydus_storage as storage;
 
-#[cfg(feature = "fusedev")]
 use std::convert::TryInto;
 use std::io::{Error, ErrorKind, Result};
 use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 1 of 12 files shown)


src/bin/nydusd/main.rs
```diff
@@ -324,32 +313,50 @@ fn append_virtiofs_subcmd_options(app: App<'static, 'static>) -> App<'static, 's
...
+    /*
+    let subcmd = subcmd.arg(
+        Arg::with_name("fuse")
+            .long("fuse")
+            .short("f")
+            .help("Run as a shared FUSE server"),
+    );
+    let subcmd = append_fuse_options(subcmd);
+    let subcmd = append_fs_options(subcmd);
+    */
```
liubogithub, line 359: Why do we still need this?

liubogithub (approved):
Looks good.

src/bin/nydusd/main.rs
jiangliu: The `fuse` and `virtiofs` subcommands are designed for one image per instance mode. And the `singleton` subcommand is designed for one instance for all images mode. So we still need to add support for fuse/virtiofs for singleton.

PR #745, opened 2022-09-24
Enhancements to rafs v6 implemntation (for 2.2)

jiangliu:
Enhancements to rafs v6 implementation with:
1) correctly handle endian
2) better handle of xattr
3) strict validation of data structures


rafs/src/metadata/layout/v6.rs
```diff
@@ -177,7 +179,7 @@ impl RafsV6SuperBlock {
             return Err(einval!("invalid extended slots in Rafsv6 superblock"));
         }
 
-        if self.s_blocks == 0 {
+        if u32::from_le(self.s_blocks) == 0 {
```
hsiangkao, line 182: as a generic rule of linux, no need to convert endian when comparing with 0

jiangliu: Fixed several similar cases

hsiangkao (approved):
LGTM

PR #757, opened 2022-09-27
makefile: don't provide virtiofs to all cargo commands

changweige:
virtiofs feature is not existed anymore.


jiangliu:
Feature `virtiofs` still exists, and is used to disable virtiofs related code on macos.

jiangliu:
Seems like we are lacking of smoke test case to check availability of `virtiofs`:(

Issue #723, opened 2022-09-16
Support fallback to HTTP scheme for https registry when using proxy


imeoer:
See the point 2 at: https://github.com/dragonflyoss/image-service/issues/716#issuecomment-1248915349

changweige:
Let's close it since the feature has already been merged into mainline

PR #759, opened 2022-09-27
storage: support configure proxy use_http

changweige:
Dragonfly supports configuration `useHTTPS` while
work as a http proxy. In this mode, nydusd can access dfdaemon via http while dfdaemon is accessing registry using https. So CA is not needed to be passed to dfdaemon.

The limit is that registry scheme must be set to http ending up with reading the original registry failure when fallback.

With configuring proxy with use_http, it will replace the scheme when sending requests to dfdaemon.

Addressing https://github.com/dragonflyoss/image-service/issues/723

### DIFF
--- a/api/src/http.rs
+++ b/api/src/http.rs
@@ -404,6 +404,9 @@ pub struct ProxyConfig {
     pub fallback: bool,
     /// Interval of P2P proxy health checking, in seconds.
     pub check_interval: u64,
+    /// Replace URL to http to request source registry with proxy, and allow fallback to https if the proxy is unhealthy.
+    #[serde(default)]
+    pub use_http: bool,
 }
 
 impl Default for ProxyConfig {
@@ -413,6 +416,7 @@ impl Default for ProxyConfig {
             ping_url: String::new(),
             fallback: true,
             check_interval: 5,
+            use_http: false,
         }
     }
 }
--- a/storage/src/backend/connection.rs
+++ b/storage/src/backend/connection.rs
@@ -7,7 +7,7 @@ use std::cell::RefCell;
 use std::collections::HashMap;
 use std::io::{Read, Result};
 use std::str::FromStr;
-use std::sync::atomic::{AtomicBool, AtomicU8, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicI16, AtomicU8, Ordering};
 use std::sync::Arc;
 use std::thread;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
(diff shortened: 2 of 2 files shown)


changweige:
FYI @gaius-qi

api/src/http.rs
```diff
@@ -404,6 +404,7 @@ pub struct ProxyConfig {
     pub fallback: bool,
     /// Interval of P2P proxy health checking, in seconds.
     pub check_interval: u64,
+    pub use_http: bool,
 }
```
jiangliu, line 408: Please add doc for the new "use_http" field

api/src/http.rs
```diff
@@ -404,6 +404,7 @@ pub struct ProxyConfig {
     pub fallback: bool,
     /// Interval of P2P proxy health checking, in seconds.
     pub check_interval: u64,
```
jiangliu, line 406: I'm a little concerned about the backward compatibility here.
There's no default serde function for data fields here.
Seems it may fail to parse json string missings any fields. For example, existing client may send a json config string without the "use_http" field, what happens then ?
So suggest to define default value for fields by #[serde(default = xxx_default).

api/src/http.rs
changweige: Added

api/src/http.rs
changweige: Reasonable, added serde(default) macro

storage/src/backend/connection.rs
```diff
@@ -345,10 +354,37 @@ impl Connection {
                     Some(ReqBody::Buf(buf)) => Some(ReqBody::Buf(buf.clone())),
                     _ => None,
                 };
+
```
jiangliu, line 357: It would be great to abstract this as a helper function.

changweige: Good point, will add such a function ~

api/src/http.rs
```diff
@@ -404,6 +404,9 @@ pub struct ProxyConfig {
...
+    /// Replace url sent to proxy from https to http. So no CA is needed to be passed to the proxy.
```
imeoer, line 407: We can't assume the proxy server behavior, so I think this comment is better:

```
Replace URL to http to request source registry with proxy, and allow fallback to https if the proxy is unhealthy.
```

api/src/http.rs
```diff
@@ -413,6 +416,7 @@ impl Default for ProxyConfig {
             ping_url: String::new(),
             fallback: true,
             check_interval: 5,
+            use_http: true,
```
imeoer, line 419: Why use `true` as default here?

api/src/http.rs
changweige: Yes it's better

api/src/http.rs
changweige: changed to false

storage/src/backend/connection.rs
```diff
@@ -165,11 +165,39 @@ impl ProxyHealth {
     }
 }
 
+const UNSET: i16 = 0;
```
imeoer, line 168: Can we clarify these constants? for example `SCHEME_CACHE_UNSET`.

storage/src/backend/connection.rs
```diff
@@ -165,11 +165,39 @@ impl ProxyHealth {
...
+const REPLACE: i16 = 1;
+const RETAIN: i16 = 2;
+
 #[derive(Debug)]
 struct Proxy {
     client: Client,
     health: ProxyHealth,
     fallback: bool,
+    use_http: bool,
```
imeoer, line 177: Is necessary to use this variable? How about reusing `replace_scheme`?

changweige: We can't because `use_http` is a switch controlling whether to replace HTTP scheme while `replace_sheme` is only a cache indicating whether we should try replacing the scheme if the original registry URL is already HTTP scheme.

storage/src/backend/connection.rs
```diff
@@ -345,10 +375,21 @@ impl Connection {
...
                     _ => None,
                 };
+
+                let http_url: Option<String>;
+                let mut replaced_url = url;
+
+                if proxy.use_http {
+                    http_url = proxy.try_use_http(url);
+                    if let Some(ref r) = http_url {
+                        replaced_url = r.as_str();
```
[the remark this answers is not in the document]
changweige: sure

storage/src/backend/connection.rs
changweige: added

api/src/http.rs
changweige: replaced

api/src/http.rs
changweige: fixed

storage/src/backend/connection.rs
changweige: fixed

storage/src/backend/connection.rs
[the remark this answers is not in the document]
changweige: added

imeoer (approved):
LGTM

Issue #760, opened 2022-09-27
nydus-image inspect panics


changweige:
It panics when perform `prefetch` command 

```
-    235      "\xE2\xE3\xB7席\xB7\xFB\xB3\xB5\xE2\xEF\xE2\u{2cf4}\xE2\xE7\xB7\xFE\xB7\xF4\xE2\xFC\xE2\xF5\xE2\xF6\xB7\xF6ӺӮӼ\xE2\xF8.regular"
Inspecting Rafs :> prefetch
Prefetched Files: 1
[2022-09-27 12:54:15.946612 +08:00] ERROR [/home/gechangwei/.cargo/registry/src/mirrors.sjtug.sjtu.edu.cn-4f7dbcce21e258a2/log-panics-2.1.0/src/lib.rs:130] thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: NotFound, error: "No such file or directory (os error 2): rafs/src/metadata/direct_v6.rs:820" }': rafs/src/metadata/direct_v6.rs:1163
   0: backtrace::capture::Backtrace::create
   1: backtrace::capture::Backtrace::new

Aborted (core dumped)
```

imeoer:
@YushuoEdge PTAL if have any chance :)

changweige:
The root cause is that the regular file inode does not point to its parent dir. To fix this, must have a way similar to lookup

PR #767, opened 2022-09-30
Enrich nydus-image print, fix prefetch table layout and other couple of fixes

changweige:
1. rafs v5 and v6 has different root inode numbers. So use the real root inode number when getting path from inode number.
2. rafs v6's inode number can't be decided until the end of dumping bootstrap. So don't record inode number in readadhead patterns anymore, use node index instead.
3. enrich print nydus-image inspect 

fixes https://github.com/dragonflyoss/image-service/issues/760

### DIFF
--- a/rafs/src/metadata/layout/v6.rs
+++ b/rafs/src/metadata/layout/v6.rs
@@ -219,6 +219,11 @@ impl RafsV6SuperBlock {
         self.s_inos = inos.to_le();
     }
 
+    /// Get total inodes count of this Rafs
+    pub fn inodes_count(&self) -> u64 {
+        self.s_inos
+    }
+
     /// Set number of logical blocks.
     pub fn set_blocks(&mut self, blocks: u32) {
         self.s_blocks = blocks.to_le();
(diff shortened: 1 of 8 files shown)


rafs/src/metadata/layout/v6.rs
```diff
@@ -219,6 +219,10 @@ impl RafsV6SuperBlock {
         self.s_inos = inos.to_le();
     }
 
+    pub fn inodes_count(&self) -> u64 {
```
jiangliu, line 222: Please also add doc for this function

src/bin/nydus-image/core/prefetch.rs
```diff
@@ -168,7 +169,7 @@ impl Prefetch {
         if self.policy == PrefetchPolicy::Fs {
             let mut prefetch_table = RafsV5PrefetchTable::new();
             for i in self.readahead_patterns.values().filter_map(|v| *v) {
-                prefetch_table.add_entry(i as u32);
+                prefetch_table.add_entry(i as u32 + 1);
```
jiangliu, line 172: why plus one here?

changweige: Yes. I made a mistake here. will fix it.

rafs/src/metadata/layout/v6.rs
changweige: added

jiangliu (approved):

PR #772, opened 2022-09-30
rafs: enforce stricter inode validation rules for v5

jiangliu:
Enforce stricter inode validation rules for v5 by: 1) ensure ino is not zero
2) inode is less than max_inodes
3) i_name is not empty
4) parent inode is not zero
5) n_link is not zero
6) parent inode is less than current inode for non-hard-link inodes 7) i_blocks is correct
8) child inode is bigger than parent
9) child count is corrent

### DIFF
--- a/rafs/src/metadata/cached_v5.rs
+++ b/rafs/src/metadata/cached_v5.rs
@@ -355,27 +355,36 @@ impl CachedInodeV5 {
 }
 
 impl RafsInode for CachedInodeV5 {
-    #[allow(clippy::collapsible_if)]
+    // Somehow we got invalid `inode_count` from superblock.
     fn validate(&self, _inode_count: u64, chunk_size: u64) -> Result<()> {
-        if self.i_ino == 0 || self.i_name.len() > RAFS_MAX_NAME || self.i_nlink == 0 {
+        if self.i_ino == 0
+            // || self.i_ino > inode_count
+            || self.i_nlink == 0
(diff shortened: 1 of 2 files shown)


rafs/src/metadata/cached_v5.rs
```diff
@@ -355,27 +355,36 @@ impl CachedInodeV5 {
...
+            return Err(einval!("invalid parent inode"));
+        }
         if self.is_reg() {
             let chunks = (self.i_size + chunk_size - 1) / chunk_size;
             if !self.has_hole() && chunks != self.i_data.len() as u64 {
                 return Err(einval!("invalid chunk count"));
             }
+            let blocks = (self.i_size + 511) / 512;
+            // Old stargz builder generates inode with 0 blocks
+            if blocks != self.i_blocks && self.i_blocks != 0 {
```
imeoer, line 379: It seems we calculate the blocks by:

```
self.inode.set_blocks(div_round_up(
  self.inode.size() + self.xattrs.aligned_size_v5() as u64,
  512,
));
```

bergwolf (approved):
lgtm!

rafs/src/metadata/cached_v5.rs
jiangliu: Seems that code path only works for directory builder, stargz builder has its own implementation.

jiangliu: We should fix stargz builder.

PR #774, opened 2022-10-02
Split `RafsInode` into `RafsInode` and `RafsInodeExt`

jiangliu:
The `RafsInode` is designed according to Rafs v5, which causes some problems when enabling Rafs v6.
So split it into `RafsInode` and `RafsInodeExt. Among those, `RafsInode` is used to provide methods which are generic to v5/v6 and runtime/builder, `RafsInodeExt` provides version specific methods or methods for builder.

Along the way also do:
1) clean up some unused code
2) rename RAFS_ROOT_INODE to RAFS_V5_ROOT_INODE
3) add more unit test cases
4) add some helpers function related to v5 to simplify implementation

### DIFF
--- a/rafs/src/metadata/direct_v6.rs
+++ b/rafs/src/metadata/direct_v6.rs
@@ -1379,12 +1379,6 @@ impl BlobChunkInfo for DirectChunkInfoV6 {
             .contains(BlobChunkFlags::COMPRESSED)
     }
 
-    fn is_hole(&self) -> bool {
-        self.chunk(self.state().deref())
-            .flags
-            .contains(BlobChunkFlags::HOLECHUNK)
-    }
-
     fn as_any(&self) -> &dyn Any {
         self
     }
--- a/rafs/src/metadata/direct_v6.rs
+++ b/rafs/src/metadata/direct_v6.rs
@@ -34,6 +34,12 @@ use std::slice;
 use std::sync::Arc;
 
 use arc_swap::{ArcSwap, Guard};
+use nydus_utils::{digest::RafsDigest, div_round_up, round_up};
+use storage::device::{
+    v5::BlobV5ChunkInfo, BlobChunkFlags, BlobChunkInfo, BlobInfo, BlobIoChunk, BlobIoDesc,
+    BlobIoVec,
+};
+use storage::utils::readahead;
 
 use crate::metadata::layout::MetaRange;
 use crate::metadata::{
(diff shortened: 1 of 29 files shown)


src/bin/nydusd/fusedev.rs
```diff
@@ -113,10 +113,11 @@ impl FuseServer {
...
 
         loop {
-            if let Some((reader, writer)) = self.ch.get_request().map_err(|e| {
-                warn!("get fuse request failed: {:?}", e);
-                Error::from_raw_os_error(libc::EINVAL)
-            })? {
+            if let Some((reader, writer)) = self
+                .ch
+                .get_request()
+                .map_err(|_e| Error::from_raw_os_error(libc::EINVAL))?
```
imeoer, line 119: Better not to ignore the error message here?

src/bin/nydusd/fusedev.rs
```diff
@@ -255,8 +256,7 @@ impl FusedevDaemon {
         let thread = thread::Builder::new()
             .name("fuse_server".to_string())
             .spawn(move || {
-                if let Err(err) = s.svc_loop(&inflight_op) {
-                    warn!("fuse server exits with err: {:?}, exiting daemon", err);
+                if let Err(_err) = s.svc_loop(&inflight_op) {
```
imeoer, line 259: Have any other errors need to be printed?

rafs/src/metadata/direct_v6.rs
```diff
@@ -1264,14 +1204,83 @@ impl RafsInode for OndiskInodeWrapper {
...
+    /// Get name of the inode.
+    ///
+    /// # Safety
+    /// It depends on Self::validate() to ensure valid memory layout.
+    fn name(&self) -> OsString {
+        let mut curr_name = OsString::from("");
+        match self.name.borrow().as_ref() {
+            Some(name) => return name.clone(),
+            None => {
+                debug_assert!(self.is_dir());
```
imeoer, line 1239: Do we still depend on `debug_assert`? This will cause different behavior between debug and release binary. Maybe use `Result<OsString> here`.

jiangliu: then let's use "assert"?

src/bin/nydusd/fusedev.rs
jiangliu: I observed that the error message is mainly issued for quit. 
And in case real error happens, the caller should print the error message.

src/bin/nydusd/fusedev.rs
jiangliu: Good point, better to pop the error reason up.

jiangliu: Fixed

rafs/src/metadata/direct_v6.rs
jiangliu: Fixed

imeoer (approved):

PR #778, opened 2022-10-07
Second part to refine nydus-image

jiangliu:
Second part to refine nydus-image

### DIFF
--- a/src/bin/nydus-image/builder/directory.rs
+++ b/src/bin/nydus-image/builder/directory.rs
@@ -7,6 +7,7 @@ use std::fs::DirEntry;
 use std::io::Write;
 
 use anyhow::{Context, Result};
+use sha2::Digest;
 
 use crate::builder::Builder;
 use crate::core::blob::Blob;
@@ -153,6 +154,10 @@ impl Builder for DirectoryBuilder {
             ));
         };
 
+        if ctx.inline_bootstrap {
+            let (_, _) = blob_mgr.get_or_create_current_blob(ctx)?;
+        }
+
         // Dump blob file
         let mut blob = Blob::new();
         timing_tracer!(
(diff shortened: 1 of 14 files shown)


src/bin/nydus-image/core/context.rs
```diff
@@ -37,25 +37,46 @@ use super::prefetch::{Prefetch, PrefetchPolicy};
...
 pub const BUF_WRITER_CAPACITY: usize = 2 << 17;
 
-#[derive(Clone, Copy, Debug, PartialEq)]
-pub enum SourceType {
-    Directory,
-    StargzIndex,
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum ConversionType {
+    DirectoryToRafs,
+    DirectoryToStargz,
```
imeoer, line 43: Maybe we need to call it `EStargz`, it's a different format from stargz (it's not a blocker for this PR).

src/bin/nydus-image/builder/directory.rs
```diff
@@ -147,39 +149,57 @@ impl Builder for DirectoryBuilder {
...
+        // Dump bootstrap file
+        let blob_table = blob_mgr.to_blob_table(ctx)?;
+        bootstrap.dump(
+            ctx,
+            &mut bootstrap_mgr.bootstrap_storage,
+            &mut bootstrap_ctx,
+            &blob_table,
+        )?;
+
         if let Some((_, blob_ctx)) = blob_mgr.get_current_blob() {
```
imeoer, line 172: If no blob is generated, we still need to write bootstrap data and tar header (`TAR_BOOTSTRAP_NAME`) into the blob if `inline-bootstrap` is enabled.

imeoer:
Other LGTM!

src/bin/nydus-image/core/context.rs
jiangliu: Will fix in later patch.

src/bin/nydus-image/builder/directory.rs
jiangliu: Fixed

imeoer (approved):

Issue #608, opened 2022-07-25
【bug】nydusify pack can not save the blob file to the root path of oss bucket


zhangshuiyong:
**nydus Version:**
2.1.0-alpha2

**oss-backend-config.json:** 
<img width="439" alt="image" src="https://user-images.githubusercontent.com/1655233/180732044-e9c163c6-f322-43c9-a44b-2c952accc4ed.png">

**log:**
<img width="1324" alt="image" src="https://user-images.githubusercontent.com/1655233/180732441-3c548813-5b7a-47e7-9570-f05e3c6ed09a.png">

**Code:**
<img width="476" alt="image" src="https://user-images.githubusercontent.com/1655233/180732695-fcf12fc0-d6a4-44dd-b2a4-35580d9d6f43.png">

**Bug result: it will create a folder name "/" , then save the blob file into the folder.**

<img width="423" alt="image" src="https://user-images.githubusercontent.com/1655233/180732778-be37ab4e-4d09-4a3e-8423-80ca6f2e9d52.png">

imeoer:
![image](https://user-images.githubusercontent.com/1524576/180736854-6c578771-781d-47c3-aa00-f0e73bbed982.png)

Thanks for reporting, it looks like the root cause is here.

ccx1024cc:
@zhangshuiyong Nydus use specified prefix directly, instead of with `/` suffix. Files are not in `root` with empty prefix now. If u want, specify prefix = `/`.

imeoer:
@zhangshuiyong Fixed in `v2.1.1` release.

PR #780, opened 2022-10-09
refact: use specified object prefix and meta prefix directly

ccx1024cc:
[ISSUE](https://github.com/dragonflyoss/image-service/issues/608)

### DIFF
--- a/contrib/nydusify/pkg/packer/packer.go
+++ b/contrib/nydusify/pkg/packer/packer.go
@@ -67,7 +67,7 @@ func (cfg *BackendConfig) rawMetaBackendCfg() []byte {
 		"access_key_id":     cfg.AccessKeyID,
 		"access_key_secret": cfg.AccessKeySecret,
 		"bucket_name":       cfg.BucketName,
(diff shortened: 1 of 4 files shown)


contrib/nydusify/pkg/packer/packer.go
```diff
@@ -67,7 +69,10 @@ func (cfg *BackendConfig) rawMetaBackendCfg() []byte {
 		"access_key_id":     cfg.AccessKeyID,
 		"access_key_secret": cfg.AccessKeySecret,
 		"bucket_name":       cfg.BucketName,
-		"object_prefix":     cfg.MetaPrefix + "/",
+		"object_prefix":     defaultMetaPrefix,
```
imeoer, line 72: How do users put the object in the root directory of bucket?

ccx1024cc: > How do users put the object into the root directory of bucket?

Compatible to specified suffix / now.
User get it by prefix = "/"

imeoer (approved):

Issue #768, opened 2022-09-30
nydus-image inspect icheck panics (rafs v6)


changweige:
For rafs v6, disk inode does not point to its parent dir. So `icheck` is hard to implement and panic right now. We can fix this by using the lookup method.

jiangliu:
Fixed by https://github.com/dragonflyoss/image-service/pull/767？

PR #781, opened 2022-10-09
nydus-image: fix rafs v6 to get file's parent inode

sctb512:
Currently, the parent of file is None,
we can set the parent inode by traversing root directory.
This PR is addressing #768 

### DIFF
--- a/rafs/src/metadata/direct_v6.rs
+++ b/rafs/src/metadata/direct_v6.rs
@@ -33,6 +33,7 @@ use std::slice;
 use std::sync::Arc;
 
 use arc_swap::{ArcSwap, Guard};
+use nydus_utils::metrics::Inode;
 
 use crate::metadata::layout::MetaRange;
 use crate::metadata::{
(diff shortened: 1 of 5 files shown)


jiangliu:
I have different thought here. It's a little heavy to enhance get_inode_ext() to support such a case, it may  potentially cause serious performance issue. So I would prefer only to change `nydus-image icheck` implementation.

rafs/src/metadata/direct_v6.rs
```diff
@@ -356,6 +356,29 @@ impl RafsSuperInodes for DirectSuperBlockV6 {
...
+                        return Ok(Arc::new(inode));
+                    }
+
+                    if child.is_dir() {
+                        let child_inode = self.get_inode(child.ino(), false)?;
+                        stack.push_back(child_inode);
+                    }
+                }
+            }
+        }
```
changweige, line 381: We'd better make this optional in order not to affect nydusd runtime

rafs/src/metadata/direct_v6.rs
```diff
@@ -356,6 +356,29 @@ impl RafsSuperInodes for DirectSuperBlockV6 {
...
         }
+
+        let root_ino = self.root_ino();
+        let root = self.get_inode(root_ino, false)?;
+        let mut stack = VecDeque::new();
+        if root.is_dir() {
+            stack.push_back(root);
+        }
+        while !stack.is_empty() {
+            let parent = stack.pop_front().unwrap();
```
changweige, line 367: Don't directly unwrap unless is safe, which should be commented

rafs/src/metadata/direct_v6.rs
sctb512: Fixed.

PR #785, opened 2022-10-10
nydus-image: fix inspect to get correct path of rafs v6 file

sctb512:
Because rafs v6 doesn't support get_parent, the prefetch and icheck command of inspect will cause error. We fixed it by handling get_file_name get_file_name and path_from_ino for rafs v6 files separately. 

For hardlink, we will get all the paths of file:
image

This PR is addressing #768 

### DIFF
--- a/src/bin/nydus-image/inspect.rs
+++ b/src/bin/nydus-image/inspect.rs
@@ -3,11 +3,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
+    ffi::OsString,
     fs::Permissions,
     io::{Error, ErrorKind, Write},
     ops::DerefMut,
     os::unix::prelude::PermissionsExt,
-    path::Path,
+    path::{Path, PathBuf},
     sync::{Arc, Mutex},
 };
 
@@ -164,7 +165,8 @@ impl RafsInspector {
     fn cmd_stat_file(&self, file_name: &str) -> Result<Option<Value>, anyhow::Error> {
         // Stat current directory
         if file_name == "." {
-            return self.stat_single_file(self.cur_dir_ino);
+            let inode = self.rafs_meta.get_inode(self.cur_dir_ino, false)?;
+            return self.stat_single_file(Some(inode.parent()), self.cur_dir_ino);
         }
 
         // Walk through children inodes to find the file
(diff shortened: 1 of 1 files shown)


rafs/src/metadata/direct_v6.rs
```diff
@@ -33,6 +33,7 @@ use std::slice;
 use std::sync::Arc;
 
 use arc_swap::{ArcSwap, Guard};
+use nydus_utils::metrics::Inode;
```
jiangliu, line 36: This change seems wrong, we should use Inode from crate::metadata::Inode

rafs/src/metadata/direct_v6.rs
```diff
@@ -1159,14 +1160,49 @@ impl RafsInode for OndiskInodeWrapper {
         match self.parent_inode.get() {
             Some(parent) => parent,
             None => {
-                debug_assert!(self.is_dir());
```
jiangliu, line 1162: It would be better to keep `parent()` as is, and modify `nydus-image` to call `file_parent_inode()` directly.

bergwolf:
rafs v5 can have i_parent for every inode because it duplicates hardlink inodes and each would have different i_parent. For rafs v6, this is unsupported by design. If we still want to have such functionality, we need to be aware that for rafs v6 hardlinks, it is unreliable to find inode's parent by traversing from the root inode, and specially deal with it in `nydus-image inspect`.

rafs/src/metadata/direct_v6.rs
sctb512: Ok, i got it.

sctb512:
> nydus-image inspect

Ok, maybe deal with it in `nydus-image inspect` is better.

jiangliu:
> > nydus-image inspect
> 
> Ok, maybe deal with it in `nydus-image inspect` is better.

The fundamental issue is that v6 doesn't support "get_parent", so `nydus-image inspect` could only do best effort to find a parent. And the runtime can't guarenttee `get_parent()`

sctb512:
> > > nydus-image inspect
> > 
> > 
> > Ok, maybe deal with it in `nydus-image inspect` is better.
> 
> The fundamental issue is that v6 doesn't support "get_parent", so `nydus-image inspect` could only do best effort to find a parent. And the runtime can't guarenttee `get_parent()`

I got it.
It means try to find parent only in `nydus-image inspect`, right? 
For hardlink, if needed, we can traverse it completely to find all parent?

rafs/src/metadata/direct_v6.rs
sctb512: Sorry, I fixed it already.

PR #788, opened 2022-10-10
storage: retry timeout chunks for fscache ondemand path

kevinXYin:
for fscache ondemand path, if some requested chunks are set to pending by prefetch threads, and wait them timeout, will casue EIO to container side.

retry the timeout chunks on ondemand path, minimize EIOs.

### DIFF
--- a/storage/src/cache/cachedfile.rs
+++ b/storage/src/cache/cachedfile.rs
@@ -579,7 +579,28 @@ impl FileCacheEntry {
         }
 
         if !bitmap.wait_for_range_ready(chunk_index, count)? {
-            Err(eio!("failed to read data from storage backend"))
+            if prefetch {
+                return Err(eio!("failed to read data from storage backend"));
+            }
+            // if we are in ondemand path, retry for the timeout chunks
+            for chunk in chunks {
+                if self.chunk_map.is_ready(chunk.as_ref())? {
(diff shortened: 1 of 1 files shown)


storage/src/cache/cachedfile.rs
```diff
@@ -579,7 +579,30 @@ impl FileCacheEntry {
         }
 
         if !bitmap.wait_for_range_ready(chunk_index, count)? {
-            Err(eio!("failed to read data from storage backend"))
+            if !prefetch {
```
jiangliu, line 582: Minor suggestion:
```
if prefetch {
 return Err(eio!("failed to read data from storage backend"));
}
...
```
This function is a little over complex, will try to refactor it later

kevinXYin: Thanks , updated. 
Yes , this func is too long now , and I remember seems there's still issue for stargz part, may need farther check later.

jiangliu (approved):

changweige:
I didn't figure out why `wait_for_range_ready` is called after synchronous pulls of chunks. It looks like somewhat of a validation. But I am afraid something is wrong if wait_for_range_ready returns timeout ERROR since it is called after a synchronous download. It should always return true immediately.

PR #789, opened 2022-10-11
nydusd: enlarge default fuse server threads

changweige:
Now the default value is only 1, it affacts performance.


changweige:
I safely merge this PR since it is cherry-picked

PR #792, opened 2022-10-14
Improve implementation to convert stargz TOC into RAFS fs

jiangliu:
Improve implementation to convert stargz TOC into RAFS fs.

### DIFF
--- a/src/bin/nydus-image/main.rs
+++ b/src/bin/nydus-image/main.rs
@@ -14,8 +14,6 @@ extern crate serde;
 extern crate serde_json;
 #[macro_use]
 extern crate lazy_static;
-extern crate nydus_rafs as rafs;
-extern crate nydus_storage as storage;
 
 use std::fs::{self, metadata, DirEntry, File, OpenOptions};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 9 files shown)


src/bin/nydus-image/main.rs
```diff
@@ -989,6 +996,22 @@ impl Command {
         Ok(blob_id)
     }
 
+    fn get_blob_size(matches: &clap::ArgMatches, ty: ConversionType) -> Result<u64> {
+        if ty != ConversionType::StargzIndexToRef {
+            return Ok(0);
+        }
+
+        match matches.value_of("blob-data-size") {
+            None => bail!("no value specified for '--blob-data-size'"),
```
imeoer, line 1005: Does it seem must be calculated by `blob_size - footer_size` in snapshotter?

imeoer:
LGTM.

imeoer (approved):

src/bin/nydus-image/main.rs
jiangliu: It's the toc offset from estargz image. When snapshotter tries to downloaded the TOC, it already gets the `blob-data-size` by parsing the eStargz footer.

PR #793, opened 2022-10-14
nydus-image/v5: prefetch table should contain inode numbers rather its index

changweige:
Nydusd is performing prefetch by finding all inodes matching their inode numbers.


changweige:
@sctb512 Could you please help test this patch?

sctb512:
> @sctb512 Could you please help test this patch?

Absolutely, the result is that the prefetch table already contains the correct inode of hardlink.
image

Issue #765, opened 2022-09-29
Discuss about how to check mirror registry health


changweige:
Nydus has merged two features about deeply integrating nydusd with Dragonfly
- use Dragonfly as a pull-through cache a.k.a mirror registry
- work with Dragonfly as HTTP proxy useHTTPS accessing HTTPS registry with the capability to fallback to the original HTTPS registry.

For Dragonfly mirror mode, nydusd will continue to pull blob data from the next mirror if the current mirror fails, but there is no mechanism to recover access to the previously failed mirror.

To solve this, I can think of two methods:
1. add a health check endpoint to dragonfly
2. periodically get repo tags. If nydusd can successfully get repo tags from the failed mirror, it is treated recovered.

changweige:
This might be the correct way

> A properly secured registry should return 401 when the "/v2/" endpoint is hit
without credentials. The response should include a `WWW-Authenticate`
challenge, providing guidance on how to authenticate, such as with basic auth
or a token service. If the load balancer has health checks, it is recommended
to configure it to consider a 401 response as healthy and any others as down.
This will secure your registry by ensuring that configuration problems with
authentication don't accidentally expose an unprotected registry. If you're
using a less sophisticated load balancer, such as Amazon's Elastic Load
Balancer, that doesn't allow one to change the healthy response code, health
checks can be directed at "/", which will always return a `200 OK` response.

imeoer:
> This might be the correct way

Should also ensure that all registry providers (harbor, ghcr, docker hub, acr...) have implemented the same behavior.

changweige:
> > This might be the correct way
> 
> Should also ensure that all registry providers (harbor, ghcr, docker hub, acr...) have implemented the same behavior.

It comes from OCI distribution spec, and looks very reliable.https://github.com/distribution/distribution/blob/main/docs/deploying.md#importantrequired-http-headers

changweige:
The checker code has been merged.

PR #800, opened 2022-10-18
storage: add mirror health checking support

sctb512:
Currently, the mirror is set to unavailable if the failed times reach failure_limit. We added mirror health checking, which will recover unavailable mirror server. The failure_limit indicates the failed time at which the mirror is set to unavailable, The health_check_interval indicates the time interval to recover the unavailable mirror and the ping_url is the endpoint for checking mirror server health.

image

This PR is addressing #765 

### DIFF
--- a/api/src/http.rs
+++ b/api/src/http.rs
@@ -400,6 +400,7 @@ impl Default for ProxyConfig {
 
 /// Configuration for mirror.
 #[derive(Clone, Deserialize, Serialize, Debug)]
+#[serde(default)]
 pub struct MirrorConfig {
     /// Mirror server URL, for example http://127.0.0.1:65001.
     pub host: String,
@@ -412,6 +413,28 @@ pub struct MirrorConfig {
     /// e.g. when using Dragonfly server as mirror, authorization through it will affect performance.
     #[serde(default)]
     pub auth_through: bool,
+    /// Interval of mirror health checking, in seconds.
+    #[serde(default = "default_check_interval")]
+    pub health_check_interval: u64,
+    /// Failure count for which mirror is considered unavailable.
+    #[serde(default = "default_failure_limit")]
+    pub failure_limit: u8,
+    /// Ping URL to check mirror server health.
+    #[serde(default)]
+    pub ping_url: String,
+}
+
+impl Default for MirrorConfig {
+    fn default() -> Self {
+        Self {
+            host: String::new(),
+            headers: HashMap::new(),
+            auth_through: false,
+            health_check_interval: 5,
+            failure_limit: 5,
+            ping_url: String::new(),
+        }
+    }
 }
 
 #[derive(Debug)]
@@ -620,6 +643,14 @@ fn default_http_scheme() -> String {
 }
 
 fn default_http_timeout() -> u32 {
+    5
+}
+
+fn default_check_interval() -> u64 {
+    5
+}
+
+fn default_failure_limit() -> u8 {
     5
 }
 
--- a/docs/nydusd.md
+++ b/docs/nydusd.md
@@ -265,13 +265,21 @@ Currently, the mirror mode is only tested in the registry backend, and in theory
               // When Dragonfly does not cache data, nydusd will pull it from "X-Dragonfly-Registry".
               // If not set "X-Dragonfly-Registry", Dragonfly will pull data from proxy.registryMirror.url.
               "X-Dragonfly-Registry": "https://index.docker.io"
-            }
+            },
+            // This URL endpoint is used to check the health of mirror server, and if the mirror is unhealthy, 
+            // the request will fallback to the next mirror or the original registry server. 
+            // Use $host/v2 as default if left empty.
+            "ping_url": "http://127.0.0.1:40901/server/ping",
+            // Interval time (s) to check and recover unavailable mirror. Use 5 as default if left empty.
+            "health_check_interval": 5,
+            // Failure counts before disabling this mirror. Use 5 as default if left empty.
+            "failure_limit": 5,
           },
           {
             "host": "http://dragonfly2.io:65001",
             "headers": {
               "X-Dragonfly-Registry": "https://index.docker.io"
-            }
+            },
           }
         ],
         ...
--- a/storage/src/backend/connection.rs
+++ b/storage/src/backend/connection.rs
@@ -222,14 +222,14 @@ pub(crate) fn respond(resp: Response, catch_status: bool) -> ConnectionResult<Re
 pub(crate) struct Connection {
     client: Client,
     proxy: Option<Arc<Proxy>>,
-    mirrors: Vec<Arc<Mirror>>,
-    shutdown: AtomicBool,
+    pub mirrors: Vec<Arc<Mirror>>,
+    pub shutdown: AtomicBool,
 }
 
 #[derive(Debug)]
 pub(crate) struct Mirror {
     /// Information for mirror from configuration file.
-    config: MirrorConfig,
+    pub config: MirrorConfig,
     /// Mirror status, it will be set to false by atomic operation when mirror is not work.
     status: AtomicBool,
     /// Failed times requesting mirror, the status will be marked as false when failed_times = failure_limit.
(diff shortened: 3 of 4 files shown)


api/src/http.rs
```diff
@@ -400,18 +400,33 @@ impl Default for ProxyConfig {
 
 /// Configuration for mirror.
 #[derive(Clone, Deserialize, Serialize, Debug)]
+#[serde(default)]
 pub struct MirrorConfig {
     /// Mirror server URL, for example http://127.0.0.1:65001.
     pub host: String,
     /// HTTP request headers to be passed to mirror server.
-    #[serde(default)]
```
jiangliu, line 407: why removing this? Are `headers` and `auth_through` mandatory now?

sctb512: No, headers and auth_through is not mandatory.
We have implemented Default trait for MirrorConfig, so remove this. 😊
Are this right?

jiangliu: `#[serde(default)]` is for serde, that is to handle the case when this field is missed from json string. It serves different purpose of `#[derive(Default)]`

sctb512: Got it. I will fix this error. Thank you very much.

sctb512: Done.

storage/src/backend/connection.rs
```diff
@@ -363,7 +348,50 @@ impl Connection {
...
         }
-        // TODO: check mirrors' health
+        // Check mirrors' health
+        for mirror in connection.mirrors.iter() {
+            let conn = connection.clone();
+            let connect_timeout = config.connect_timeout;
+            // Spawn thread to update the health status of mirror server
+            let mirror_cloned = mirror.clone();
+            thread::spawn(move || {
+                let mirror_health_url = format!("{}/v2", mirror_cloned.config.host);
```
imeoer, line 358: Why use this URL to check healthy?

sctb512: Because we just check whether the mirror server is alive. If the server response is 401, we also think it is alive.

Maybe we should discuss it again.

sctb512:
Related to #801 as well.

storage/src/backend/connection.rs
imeoer: How about using the ping API provided by d7y dfdaemon: `127.0.0.1:40901/server/ping`, of course, the ping endpoint URL should support configuring by users.

sctb512: It is more reasonable.
How to deal with mirror servers without `ping` API? If users do not provide `ping` API, use mirror_server/v2 or disable it?

imeoer: It's ok to use `/v2` as default.

sctb512: Done. PTAL

docs/nydusd.md
```diff
@@ -265,7 +265,10 @@ Currently, the mirror mode is only tested in the registry backend, and in theory
               // When Dragonfly does not cache data, nydusd will pull it from "X-Dragonfly-Registry".
               // If not set "X-Dragonfly-Registry", Dragonfly will pull data from proxy.registryMirror.url.
               "X-Dragonfly-Registry": "https://index.docker.io"
-            }
+            },
+            // Endpoint of mirror server health checking (including scheme).
```
imeoer, line 269: This URL endpoint is used to check the health of mirror server, and if the mirror is unhealthy, the request will 
 fallback to the next mirror or the original registry server. Use `$host/v2` as default if left empty.

docs/nydusd.md
```diff
@@ -265,7 +265,10 @@ Currently, the mirror mode is only tested in the registry backend, and in theory
...
+            // Use host/v2 if not configure this item.
+            "ping_url": "http://127.0.0.1:40901/server/ping",
```
imeoer, line 271: Maybe `health_check_interval` and `failure_limit` need also to be explained here?

storage/src/backend/connection.rs
```diff
@@ -363,7 +348,55 @@ impl Connection {
                 });
             }
         }
-        // TODO: check mirrors' health
+        // Check mirrors' health
+        for mirror in connection.mirrors.iter() {
+            let conn = connection.clone();
+            let connect_timeout = config.connect_timeout;
```
imeoer, line 354: Should use `config.timeout` here?

storage/src/backend/connection.rs
```diff
@@ -363,7 +348,55 @@ impl Connection {
...
+                info!("Mirror health checking url: {}", mirror_health_url);
+
+                loop {
+                    // Try to recover the mirror server when it is unavailable.
+                    if !mirror_cloned.status.load(Ordering::Relaxed) {
+                        info!(
+                            "Mirror {} unhealthy, try to recover",
+                            mirror_cloned.config.host
+                        );
+                        let client = Client::new();
```
imeoer, line 372: Maybe put this out of the loop.

storage/src/backend/connection.rs
```diff
@@ -363,7 +348,55 @@ impl Connection {
...
+                        let _ = client
```
imeoer, line 373: Why ignore the error? maybe we will encounter a network error.

imeoer (commented):
Others LGTM.

storage/src/backend/connection.rs
sctb512: I refer to the `proxy` health checking code. What is the difference between `connect_timeout` and `timeout`?

docs/nydusd.md
sctb512: Yes, I have added it.

docs/nydusd.md
sctb512: Already replaced it. Thanks.

storage/src/backend/connection.rs
sctb512: I understand. It is related to the next comment.

storage/src/backend/connection.rs
sctb512: Because the mirror server is not available when we get an error.  Maybe add a `warn!` log line for it?

changweige:
We also need this feature in stable/v2.1 :-)

storage/src/backend/connection.rs
```diff
@@ -363,7 +348,62 @@ impl Connection {
                 });
             }
         }
-        // TODO: check mirrors' health
+        // Check mirrors' health
+        for mirror in connection.mirrors.iter() {
+            let conn = connection.clone();
```
changweige, line 353: It's better to extract this for loop into a new function as function `new` is already too long

storage/src/backend/connection.rs
```diff
@@ -449,92 +489,83 @@ impl Connection {
...
-            {
-                let data_cloned = data.as_ref().cloned();
-
-                for (key, value) in mirror.config.headers.iter() {
-                    headers.insert(
-                        HeaderName::from_str(key).unwrap(),
-                        HeaderValue::from_str(value).unwrap(),
-                    );
+        if !self.mirrors.is_empty() {
+            let mut requested_all_mirrors = true;
```
changweige, line 493: Can we rename it to `fallback_due_auth=false` ?

storage/src/backend/connection.rs
sctb512: Ok, I add two new function: `start_proxy_health_thread` and `start_mirror_health_thread`.

storage/src/backend/connection.rs
sctb512: Done.

imeoer (approved):

storage/src/backend/registry.rs
```diff
@@ -207,6 +222,19 @@ impl RegistryState {
                 e
             ))
         })?;
+        if let Ok(now_timestamp) = SystemTime::now().duration_since(UNIX_EPOCH) {
+            self.refresh_token_time
+                .store(Some(Arc::new(now_timestamp.as_secs() + ret.expires_in)));
```
imeoer, line 227: May `expires_in` be empty for some registries?

PR #806, opened 2022-10-20
action/nydus-test: stop on the first test failure

changweige:
By default, pytest will continue executing test even current test fails. It's hard to tell what to happen on such a environment. And it makes it hard to investigate the first failed case.


changweige:
This change should not affect the smoke test 🥺

jiangliu (approved):

PR #826, opened 2022-11-02
nydusify: drop label "nydus-blob-ids" from meta layer

changweige:
Image with layers more than 64 can't be pulled by containerd since the label is exceeding the label size limitation 4096bytes.

We should figure out another way to do GC in nydus-snapshotter


imeoer (approved):
LGTM!

PR #833, opened 2022-11-03
Adopt zlib random access algorithm to support OCIv1 compatible nydus image

jiangliu:
By adopting zlib random access algorithm, nydus can build a RAFS metadata from a targz file and use the targz file as RAFS data blob. By this way, we could associate a small RAFS metadata blob to existing OCIv1 images, which provides lazy loading and data deduplication capabilities with minimal resource overhead.
Now we have a flexible architecture to support all of OCIv1/eStargz/Nydus images as below:

截屏2022-11-03 23 29 58

### DIFF
--- a/rafs/src/fs.rs
+++ b/rafs/src/fs.rs
@@ -36,13 +36,12 @@ use nix::unistd::{getegid, geteuid};
 use serde::Deserialize;
 
 use nydus_api::http::{BlobPrefetchConfig, FactoryConfig};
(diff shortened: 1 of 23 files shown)


hsiangkao:
Fixes https://github.com/dragonflyoss/image-service/issues/686

imeoer (approved):

imeoer:
Let's move forward quickly.

PR #840, opened 2022-11-06
Reduce duplicated code by using utils/filemap

jiangliu:
Use the generic filemap from nydus-utils crate to avoid duplicated code.

Also reduce unsafe code, now there's only two unsafe left in direct v5.

Fix a lifetime related by xattr related code.


imeoer (approved):

jiangliu:
Still have more pending patch for this, hold on:)

PR #842, opened 2022-11-06
storage: remove unused code for refreshing registry tokens

sctb512:
There is no need to change 'grant_type' for refreshing registry tokens. Because we use the URL with cached 'grant_type' can get token as well.

Maybe we backport it to stable/v2.1 as well?

Related to https://github.com/dragonflyoss/image-service/pull/800


imeoer (approved):

imeoer:
@sctb512 v2.1.1 has been released, so let's ship it to the next release.

PR #847, opened 2022-11-07
Enhance RAFS v6 

jiangliu:
Enhance RAFS v6:
1) enforce stricter validation rules
2) refine direct_v6

### DIFF
--- a/rafs/src/metadata/layout/v6.rs
+++ b/rafs/src/metadata/layout/v6.rs
@@ -22,6 +22,8 @@ use nydus_storage::meta::{
 use nydus_storage::{RAFS_MAX_CHUNKS_PER_BLOB, RAFS_MAX_CHUNK_SIZE};
 use nydus_utils::{compress, digest, round_up, ByteSize};
 
+use crate::metadata::layout::v5::RafsV5ChunkInfo;
+use crate::metadata::layout::MetaRange;
 use crate::metadata::{layout::RafsXAttrs, RafsStore, RafsSuperFlags};
 use crate::{impl_bootstrap_converter, impl_pub_getter_setter, RafsIoReader, RafsIoWrite};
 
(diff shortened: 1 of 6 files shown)


rafs/src/metadata/layout/v6.rs
```diff
@@ -174,22 +181,41 @@ impl RafsV6SuperBlock {
...
+        if u32::from_le(self.s_blocks) == 0 {
             return Err(einval!("invalid blocks in Rafsv6 superblock"));
         }
 
-        if self.s_u != 0 {
+        if u16::from_le(self.s_u) != 0 {
             return Err(einval!("invalid union field in Rafsv6 superblock"));
         }
 
+        if u32::from_le(self.s_xattr_blkaddr) != 0 {
```
hsiangkao, line 200: this is the start blkaddr of shared xattrs

rafs/src/metadata/layout/v6.rs
```diff
@@ -174,22 +181,41 @@ impl RafsV6SuperBlock {
             )));
         }
 
-        if self.s_inos == 0 {
+        if u64::from_le(self.s_inos) == 0 {
```
hsiangkao, line 184: no need to convert endianness when comparing with 0

jiangliu: fixed

rafs/src/metadata/layout/v6.rs
jiangliu: I think it's designed for that, but not used yet.

rafs/src/metadata/layout/v6.rs
```diff
@@ -182,14 +189,34 @@ impl RafsV6SuperBlock {
...
-            return Err(einval!("invalid blocks in Rafsv6 superblock"));
-        }
-
         if self.s_u != 0 {
             return Err(einval!("invalid union field in Rafsv6 superblock"));
         }
 
+        if self.s_xattr_blkaddr != 0 {
+            return Err(einval!(
+                "invalid extended attribute address in Rafsv6 superblock"
```
hsiangkao, line 198: `unsupported shared extended attribute namespace in Rafsv6 superblock`

it indicates that it could be supported in Nydus in the future

PR #858, opened 2022-11-11
support ref image build and run

imeoer:
Create zran blob meta from targz fifo stream like:

```
nydus-image create --type targz-ref /path/to/fifo/layer.tar.gz --inline-bootstrap --blob-meta /path/to/blob.meta
```

It allows nydus lazily load OCI targz layer with the nydus zran ref image.

Related snapshotter implementation: https://github.com/containerd/nydus-snapshotter/pull/245


src/bin/nydus-image/builder/tarball.rs
```diff
@@ -543,11 +559,31 @@ impl Builder for TarballBuilder {
...
                     &mut bootstrap,
                     blob_mgr,
-                    &mut writer,
+                    bootstrap_writer,
                 )
             },
             "dump_bootstrap"
         )?;
+
+        if let Some((_, blob_ctx)) = blob_mgr.get_current_blob() {
```
jiangliu, line 568: Does the directory build need to write TOC too?

storage/src/cache/filecache/mod.rs
```diff
@@ -10,9 +10,10 @@ use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
 use std::sync::{Arc, RwLock};
 
 use nydus_api::CacheConfigV2;
+use tokio::runtime::Runtime;
+
```
jiangliu, line 14: How about removing the blank line?

storage/src/cache/filecache/mod.rs
```diff
@@ -182,6 +183,14 @@ impl FileCacheEntry {
             .backend
             .get_reader(blob_info.blob_id())
             .map_err(|_e| eio!("failed to get blob reader"))?;
+        let rafs_blob_reader = if blob_info.rafs_blob_digest() != &[0u8; 32] {
```
jiangliu, line 186: How about detecting the inline-bootstrap case by a flag from blob_info instead of checking digest is zero?

storage/src/cache/fscache/mod.rs
```diff
@@ -215,8 +215,17 @@ impl FileCacheEntry {
             .get_reader(blob_info.blob_id())
             .map_err(|_e| eio!("failed to get blob reader"))?;
         let blob_compressed_size = Self::get_blob_size(&reader, &blob_info)?;
+        let rafs_blob_reader = if blob_info.rafs_blob_digest() != &[0u8; 32] {
```
jiangliu, line 218: ditto

storage/src/cache/filecache/mod.rs
imeoer: Ok, how about adding a flag named `BLOB_FEATURE_INLINE`?

src/bin/nydus-image/builder/tarball.rs
imeoer: Fixed.

jiangliu (approved):

PR #860, opened 2022-11-11
Introduce Nydus configuration file format version 2

jiangliu:
Introduce Nydus configuration file format version 2 with:
1) better structured configuration items
2) support both json and Toml based configuration files


imeoer (commented):
@changweige Can the nydusd configuration reconcile with the snapshotter toml configuration?

changweige:
> @changweige Can the nydusd configuration reconcile with the snapshotter toml configuration?

I think it's OK to move to toml based nydusd configuration for nydus-snapshotter. Nydus-snapshotter has its own toml configuration file in which a file path points to a template of nydusd configuration regardless of nydusd's format. The only thing for nydus-snapshotter is to fill some necessary fields of the TOML configuration. But I suggest compatibility with the old JSON configuration format.

jiangliu:
> > @changweige Can the nydusd configuration reconcile with the snapshotter toml configuration?
> 
> I think it's OK to move to toml based nydusd configuration for nydus-snapshotter. Nydus-snapshotter has its own toml configuration file in which a file path points to a template of nydusd configuration regardless of nydusd's format. The only thing for nydus-snapshotter is to fill some necessary fields of the TOML configuration. But I suggest compatibility with the old JSON configuration format.

The v2 configuration format supports both json and toml, and it should be backward-compatible with v1 json configuration file. But new features will only be supported by v2 configuration format later.

api/src/config.rs
```diff
@@ -0,0 +1,1520 @@
...
+            _ => return false,
+        }
+
+        true
+    }
+
+    /// Get configuration information for localfs
+    pub fn get_localfs_config(&self) -> Result<&LocalFsConfig> {
+        if &self.backend_type != "localfs" {
+            Err(einval!("backend type is not 'localfs"))
```
changweige, line 210: should be 'localfs' ?

api/src/config.rs
```diff
@@ -0,0 +1,1520 @@
...
+            self.localfs
+                .as_ref()
+                .ok_or_else(|| einval!("no configuration information for localfs"))
+        }
+    }
+
+    /// Get configuration information for OSS
+    pub fn get_oss_config(&self) -> Result<&OssConfig> {
+        if &self.backend_type != "oss" {
+            Err(einval!("backend type is not 'OSS"))
```
changweige, line 221: should be 'oss' ?

api/src/config.rs
```diff
@@ -0,0 +1,1520 @@
...
+    pub batch_size: usize,
+    /// Whether to validate data digest.
+    #[serde(default)]
+    pub validate: bool,
+    /// Enable support of extended attributes.
+    #[serde(default)]
+    pub enable_xattr: bool,
+    /// Io statistics.
+    #[serde(default)]
+    pub iostats_files: bool,
```
changweige, line 512: How about revising the comment line?

> Record file operation metrics for each file. It's will consume many resources. Better keep it off in productive deployment

api/src/config.rs
```diff
@@ -0,0 +1,1520 @@
...
+    pub headers: HashMap<String, String>,
+    /// Whether the authorization process is through mirror, default to false.
+    /// true: authorization through mirror, e.g. Using normal registry as mirror.
+    /// false: authorization through original registry,
+    /// e.g. when using Dragonfly server as mirror, authorization through it may affect performance.
+    #[serde(default)]
+    pub auth_through: bool,
+    /// Interval for mirror health checking, in seconds.
+    #[serde(default = "default_check_interval")]
+    pub health_check_interval: u64,
```
changweige, line 616: u64 seems too large for the check time interval?

api/src/config.rs
```diff
@@ -0,0 +1,1520 @@
...
+    #[serde(default)]
+    pub mirrors: Vec<MirrorConfig>,
+}
+
+/// Container registry configuration information to access blobs.
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+pub struct RegistryConfig {
+    /// Registry http scheme, either 'http' or 'https'
+    #[serde(default = "default_http_scheme")]
+    pub scheme: String,
```
changweige, line 302: I remembered there was once a discussion about removing the `scheme` from the config file, just prefixing the host address with shame to construct a URL in the same style as contained. @imeoer .

docs/samples/blob_cache_entry_configuration_v2.toml
```diff
@@ -0,0 +1,140 @@
...
+registry_token = "bear_token"
+# The http scheme to access blobs.
+# It is used to workaround some P2P subsystem that requires a different scheme than the registry.
+blob_url_scheme = "https"
+# Redirect blob access to a different host regardless of the one specified in 'host'.
+blob_redirected_host = "redirect.registry.com"
+
+[backend.registry.proxy]
+# Access remote storage backend via proxy, e.g. Dragonfly dfdaemon server URL.
+url = "localhost:6789"
```
changweige, line 94: registry usually has port 5000

src/bin/nydusd/fs_service.rs
```diff
@@ -48,16 +49,10 @@ impl FsBackendCollection {
...
-                    serde_json::from_str(&cmd.config).map_err(DaemonError::Serde)?;
-                trim_backend_config!(
-                    config,
-                    "access_key_id",
-                    "access_key_secret",
-                    "auth",
-                    "token"
-                );
-                Some(config)
+                let cfg = ConfigV2::from_str(&cmd.config)
```
changweige, line 52: Could we handle compatibility here?

api/src/config.rs
```diff
@@ -0,0 +1,1520 @@
...
+    /// The field is a bearer token to be sent to registry to authorize registry requests.
+    #[serde(default)]
+    pub registry_token: Option<String>,
+    /// The http scheme to access blobs. It is used to workaround some P2P subsystem
+    /// that requires a different scheme than the registry.
+    #[serde(default)]
+    pub blob_url_scheme: String,
+    /// Redirect blob access to a different host regardless of the one specified in 'host'.
+    #[serde(default)]
+    pub blob_redirected_host: String,
```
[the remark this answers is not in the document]
changweige: So we can avoid concatenating strings when requesting HTTP server

api/src/config.rs
jiangliu: fixed

api/src/config.rs
jiangliu: fixed

api/src/config.rs
jiangliu: fixed

src/bin/nydusd/fs_service.rs
jiangliu: We have handled compabibility issues at ConfigV2::from_str(), which can handle both json/toml/v1/v2 configuration files.

jiangliu:
> Seems this PR does not consider V1 JSON based configuration compatibility. Did I miss something?

It's handled in ConfigV2::from_str()
```
   fn from_str(s: &str) -> Result<ConfigV2> {
        if let Ok(v) = serde_json::from_str::<ConfigV2>(s) {
            return if v.validate() {
                Ok(v)
            } else {
                Err(einval!("invalid configuration"))
            };
        }
        if let Ok(v) = toml::from_str::<ConfigV2>(s) {
            return if v.validate() {
                Ok(v)
            } else {
                Err(einval!("invalid configuration"))
            };
        }
        if let Ok(v) = serde_json::from_str::<RafsConfig>(s) {
            if let Ok(v) = ConfigV2::try_from(v) {
                if v.validate() {
                    return Ok(v);
                }
            }
        }
        Err(einval!("failed to parse configuration information"))
    }
```

api/src/config.rs
jiangliu: changed to u32

api/src/config.rs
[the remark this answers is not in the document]
jiangliu: We can make that change in a dedicated PR:)

imeoer (approved):

PR #870, opened 2022-11-17
storage: fix registry to avoid print bearer auth

sctb512:
This is a bug, we should backport it to `stable/v2.1` as well.


bergwolf (approved):
Thanks for taking care of the backport!

PR #880, opened 2022-11-22
fix too frequent retry

changweige:
tick() will complete when the next instant reaches which is a very short time rather than 1 second.
In addition, limit the total retry times


jiangliu (approved):

jiangliu:
Does the issue exist on the master?

PR #893, opened 2022-11-24
nydusd: automatically retry registry http scheme

uran0sH:
Signed-off-by: Wenyu Huang <huangwenyuu@outlook.com>


storage/src/backend/registry.rs
```diff
@@ -136,7 +136,7 @@ enum Auth {
 
 struct RegistryState {
     // HTTP scheme like: https, http
-    scheme: String,
+    scheme: RwLock<String>,
```
jiangliu, line 139: It would be better to use `AtomicBool` instead of `RwLock<String>`, to avoid acquiring lock on each request.

uran0sH: We need to change the value of the scheme. I am confused about how to use `AtomicBool`

jiangliu: > We need to change the value of the scheme. I am confused about how to use `AtomicBool`

use `is_https: AtomicBool` as a flag to choose https or http. And we can access/change value of `is_https` without locking:)

imeoer: Maybe use `type Scheme = AtomicBool<true>` as `https`, `type Scheme = AtomicBool<false>` as `http`, then implement `Display` for `Scheme` type.

storage/src/backend/registry.rs
```diff
@@ -487,8 +491,31 @@ impl RegistryReader {
...
+            resp = match self.request::<&[u8]>(
+                Method::GET,
+                url.as_str(),
+                None,
+                headers.clone(),
+                false,
+            ) {
+                Ok(res) => res,
+                Err(e) => {
+                    if &url[0..5] != "https" {
```
imeoer, line 505: Why not check `self.state.scheme` directly here.

storage/src/backend/registry.rs
```diff
@@ -487,8 +491,31 @@ impl RegistryReader {
...
+                        return Err(e);
+                    }
+                    if self.state.scheme.read().unwrap().clone() == "https" {
```
imeoer, line 508: We need to check if the error message matches `server gave HTTP response to HTTPS client` or `connect: connection refused` (not sure about the concrete message).

storage/src/backend/registry.rs
uran0sH: `self.state.scheme` has changed, but the `url` has not changed, in some cases

storage/src/backend/registry.rs
imeoer: @jiangliu @uran0sH Note: if users have specified the `scheme` field, for performance consideration, we shouldn't automatically fallback to `http`.

storage/src/backend/registry.rs
uran0sH: When we send `https` to `http` server, we will get this error: 
```
reqwest::Error { kind: Request, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv4(10.26.200.20)), port: Some(5000), path: "/v2/my-ubuntu-nydus/blobs/sha256:6fabb2fb4a5c017e9de1fb1dc044b69aa96ea29b0928da6330f674d51ab16e51", query: None, fragment: None }, source: hyper::Error(Connect, Error { code: -9847, message: "record overflow" }) }
```

imeoer: It seems the error happens when creating the TLS connection, maybe we can reference the golang implementation here: https://github.com/golang/go/blob/f21be2fdc6f1becdbed1592ea0b245cdeedc5ac8/src/net/http/client.go#L262

storage/src/backend/registry.rs
imeoer: Could you point out where the state has changed?

storage/src/backend/registry.rs
```diff
@@ -487,8 +507,30 @@ impl RegistryReader {
...
-                self.request::<&[u8]>(Method::GET, url.as_str(), None, headers.clone(), false)?;
+            resp = match self.request::<&[u8]>(
+                Method::GET,
+                url.as_str(),
+                None,
+                headers.clone(),
+                false,
+            ) {
+                Ok(res) => res,
+                Err(RegistryError::Request(ConnectionError::Common(_)))
```
imeoer, line 518: Does this error type match the TLS connection-related error?

storage/src/backend/registry.rs
```diff
@@ -487,8 +507,30 @@ impl RegistryReader {
...
+                    if &url[0..5] == "https" =>
```
imeoer, line 519: Why we can't check `state.scheme` here?

storage/src/backend/registry.rs
```diff
@@ -821,8 +887,8 @@ mod tests {
             "http://alibaba-inc.com/v2/nydusimage".to_owned()
         );
 
-        state.scheme = "unknown_schema".to_owned();
-        assert!(state.url("image", &[]).is_err());
+        // state.scheme = "unknown_schema".to_owned();
```
imeoer, line 890: Should remove it.

imeoer:
Now we should update the `scheme` description in https://github.com/dragonflyoss/image-service/blob/master/docs/nydusd.md#registry-backend like:

```
// Registry url scheme, leave empty to automatically detect, otherwise specify to https or http
"scheme": "",
```

storage/src/backend/registry.rs
```diff
@@ -487,8 +508,31 @@ impl RegistryReader {
...
+            resp = match self.request::<&[u8]>(
+                Method::GET,
+                url.as_str(),
+                None,
+                headers.clone(),
+                false,
+            ) {
+                Ok(res) => res,
+                Err(RegistryError::Request(ConnectionError::Common(e)))
+                    if &e.source().unwrap().source().unwrap().to_string()[0..14]
```
imeoer, line 520: Is the unwrap safe?

imeoer: Can we wrap this check into a helper func?

storage/src/backend/registry.rs
```diff
@@ -559,8 +603,27 @@ impl BlobReader for RegistryReader {
...
             .url(&format!("/blobs/sha256:{}", self.blob_id), &[])
             .map_err(RegistryError::Url)?;
+
         let resp =
-            self.request::<&[u8]>(Method::HEAD, url.as_str(), None, HeaderMap::new(), true)?;
+            match self.request::<&[u8]>(Method::HEAD, url.as_str(), None, HeaderMap::new(), true) {
+                Ok(res) => res,
+                Err(RegistryError::Request(ConnectionError::Common(e)))
+                    if &e.source().unwrap().source().unwrap().to_string()[0..14]
+                        == "error:1408F10B" =>
```
imeoer, line 612: Ditto.

imeoer:
cc @uran0sH Maybe rebase the master branch to resolve the conflicts.

docs/nydusd.md
```diff
@@ -181,8 +181,8 @@ We are working on enabling cloud-hypervisor support for nydus.
       "type": "registry",
       "config": {
         ...
-        // Registry url scheme, https or http
-        "scheme": "http",
+        // Registry url scheme, leave empty to automatically detect, otherwise specify to https or http. The true represents https and false represents http
+        "scheme": true,
```
imeoer, line 185: This is an incompatible change, can we still keep it as a string and provides optional values `https` or `http` ?

storage/src/backend/registry.rs
```diff
@@ -166,6 +186,16 @@ struct RegistryState {
     cached_bearer_auth: ArcSwapOption<BearerAuth>,
 }
 
+fn error_source2_string(e: &dyn Error) -> String {
```
imeoer, line 189: How about:

```
fn needs_fallback_http(e: &dyn Error) -> bool {
  match e.source() {
    if let Some(Some(err)) = err.source() {
      return err.to_string().contains("error:1408F10B")
    }
    false
  }
}
```

storage/src/backend/registry.rs
```diff
@@ -487,8 +517,30 @@ impl RegistryReader {
...
+                None,
+                headers.clone(),
+                false,
+            ) {
+                Ok(res) => res,
+                Err(RegistryError::Request(ConnectionError::Common(e)))
+                    if &error_source2_string(&e)[0..14] == "error:1408F10B" =>
+                {
+                    if self.state.scheme.0.load(Ordering::Relaxed) {
+                        self.state.scheme.0.store(false, Ordering::Relaxed);
```
imeoer, line 532: How about putting this logic into a helper func like `self.state.fallback_http`.

imeoer (commented):
@uran0sH It LGTM! I will do a final test on my registry server, thanks!

storage/src/backend/registry.rs
```diff
@@ -166,6 +186,16 @@ struct RegistryState {
     cached_bearer_auth: ArcSwapOption<BearerAuth>,
 }
 
+fn needs_fallback_http(e: &dyn Error) -> bool {
+    match e.source() {
+        Some(err) => match err.source() {
+            Some(err) => err.to_string().contains("error:1408F10B"),
```
imeoer, line 192: Not sure the error code stamp is compatible with all registry servers, I will test it on my machine, thanks!

imeoer (approved):

PR #894, opened 2022-11-25
nydusd: make DELETE /api/v2/blobs API support cull fscache blob

kevinXYin:
If only set blob_id parameter in DELETE /api/v2/blobs, will try to cull fscache blob. Then snapshotter can use this to delete fscache blob files.

https://github.com/containerd/nydus-snapshotter/pull/262

### DIFF
--- a/api/src/http.rs
+++ b/api/src/http.rs
@@ -158,6 +158,8 @@ pub enum ApiRequest {
     GetBlobObject(BlobCacheObjectId),
     /// Delete a blob cache entry
     DeleteBlobObject(BlobCacheObjectId),
+    /// Delete a blob cache file
+    DeleteBlobFile(String),
 }
 
 /// Kinds for daemon related error messages.
@@ -291,6 +293,8 @@ pub enum HttpError {
     CreateBlobObject(ApiError),
     /// Failed to delete blob object
     DeleteBlobObject(ApiError),
+    /// Failed to delete blob file
+    DeleteBlobFile(ApiError),
     /// Failed to list existing blob objects
     GetBlobObjects(ApiError),
 }
--- a/src/bin/nydusd/api_server_glue.rs
+++ b/src/bin/nydusd/api_server_glue.rs
@@ -87,6 +87,7 @@ impl ApiServer {
             ApiRequest::GetBlobObject(_param) => todo!(),
             ApiRequest::CreateBlobObject(entry) => self.create_blob_cache_entry(&entry),
             ApiRequest::DeleteBlobObject(param) => self.remove_blob_cache_entry(&param),
+            ApiRequest::DeleteBlobFile(blob_id) => self.blob_cache_gc(blob_id),
         };
 
         self.respond(resp);
@@ -324,6 +325,13 @@ impl ApiServer {
                 }
             }
         }
+    }
+
+    fn blob_cache_gc(&self, blob_id: String) -> ApiResponse {
+        self.get_daemon_object()?
+            .delete_blob(blob_id)
+            .map_err(|e| ApiError::DaemonAbnormal(e.into()))
+            .map(|_| ApiResponsePayload::Empty)
     }
 
     fn do_start(&self) -> ApiResponse {
--- a/src/bin/nydusd/daemon.rs
+++ b/src/bin/nydusd/daemon.rs
@@ -243,6 +243,10 @@ pub trait NydusDaemon: DaemonStateMachineSubscriber + Send + Sync {
 
     // For backward compatibility.
     fn get_default_fs_service(&self) -> Option<Arc<dyn FsService>>;
+
+    fn delete_blob(&self, _blob_id: String) -> DaemonResult<()> {
+        Ok(())
+    }
 }
 
 // State machine for Nydus daemon workflow.
--- a/src/bin/nydusd/fs_cache.rs
+++ b/src/bin/nydusd/fs_cache.rs
@@ -8,10 +8,12 @@ use std::cmp;
 use std::collections::hash_map::Entry::Vacant;
 use std::collections::HashMap;
 use std::convert::TryFrom;
-use std::fs::{File, OpenOptions};
+use std::env;
+use std::fs::{self, DirEntry, File, OpenOptions};
 use std::io::{Error, ErrorKind, Result, Write};
 use std::ops::Deref;
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
+use std::path::{Path, PathBuf};
 use std::ptr::read_unaligned;
 use std::string::String;
 use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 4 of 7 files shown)


api/src/http.rs
```diff
@@ -488,6 +488,8 @@ pub enum ApiRequest {
     GetBlobObject(BlobCacheObjectId),
     /// Delete a blob cache entry
     DeleteBlobObject(BlobCacheObjectId),
+    /// Delete a blob cache file
+    DeleteBlobFile(String),
```
bergwolf, line 492: Looks like the new blob APIs(including this one) are not documented in `api/openapi/nydus-api-v2.yaml` which still says `blob_objects`. Please update the openapi doc as well. Ideally, we should have a CI to validate the openapi interface...

src/bin/nydusd/daemon.rs
```diff
@@ -243,6 +243,10 @@ pub trait NydusDaemon: DaemonStateMachineSubscriber + Send + Sync {
 
     // For backward compatibility.
     fn get_default_fs_service(&self) -> Option<Arc<dyn FsService>>;
+
+    fn delete_blob(&self, _blob_id: String) -> DaemonResult<()> {
```
jiangliu, line 247: How about adding this method to `BlobCacheMgr` instead of `NydusDaemon`?

src/bin/nydusd/api_server_glue.rs
```diff
@@ -326,6 +327,13 @@ impl ApiServer {
         }
     }
 
+    fn blob_cache_gc(&self, blob_id: String) -> ApiResponse {
+        self.get_daemon_object()?
```
jiangliu, line 331: If we add `delete_blob()` method to `BlobCacheMgr`, then cand do
```
if let Some(blob_mgr) = DAEMON_CONTROLLER.get_blob_cache_mgr() {
     blob_mgr.delete_blob(blob_id).map_err(..).map(|_|...)
}
```

src/bin/nydusd/fs_cache.rs
```diff
@@ -248,6 +250,7 @@ pub struct FsCacheHandler {
     state: Arc<Mutex<FsCacheState>>,
     poller: Mutex<Poll>,
     waker: Arc<Waker>,
+    cache_dir: Mutex<PathBuf>,
```
jiangliu, line 253: Seems `cache_dir` is only initialized once and never updated, so no need for `Mutex` here?

src/bin/nydusd/fs_cache.rs
```diff
@@ -743,6 +747,166 @@ impl FsCacheHandler {
...
+                }
+            }
+            Err(err)
+        } else {
+            Ok(false)
+        }
+    }
+
+    fn cull(&self, cookie_dir: &Path, cookie_name: &str) -> Result<()> {
+        env::set_current_dir(&cookie_dir)?;
```
jiangliu, line 894: It seems a little dangerous to change current working directory, should we start a working thread to do this?

src/bin/nydusd/fs_cache.rs
```diff
@@ -743,6 +747,166 @@ impl FsCacheHandler {
...
+        let children = children.collect::<Result<Vec<DirEntry>>>()?;
+        let mut res = true;
+        let cwd_old = env::current_dir()?;
+
+        info!("try to cull blob {}", blob_id);
+
+        //calc blob path in all volumes then try to cull them
+        for child in children {
+            let path = child.path();
+            if !path.is_dir() || !child.file_name().to_str().unwrap().starts_with("Ierofs,") {
```
jiangliu, line 762: Please help to avoid `unwrap()` when possible:)

src/bin/nydusd/fs_cache.rs
```diff
@@ -743,6 +747,166 @@ impl FsCacheHandler {
...
+        volume_path: &Path,
+        volume_key: &str,
+        cookie_key: &str,
+    ) -> (PathBuf, String) {
+        //calc volume hash
+        let mut volume_hash_key: Vec<u8> =
+            Vec::with_capacity(self.round_up_u32(volume_key.len() + 2));
+        volume_hash_key.push(volume_key.len() as u8);
+        volume_hash_key.append(&mut volume_key.as_bytes().to_vec());
+        while volume_hash_key.len() != volume_hash_key.capacity() {
```
jiangliu, line 853: Seems it can be  simplified as `volume_hash_key.resize(volume_hash_key.capacity())`?

src/bin/nydusd/api_server_glue.rs
kevinXYin: Thanks , this should be better , will fix this.

src/bin/nydusd/fs_cache.rs
kevinXYin: Fscache_hash() need the data is a multiple of four bytes in size , so just padding 0 here as what the kernel does.

kevinXYin: Understood , thanks, will fix

src/bin/nydusd/fs_cache.rs
```diff
@@ -743,6 +747,166 @@ impl FsCacheHandler {
         unsafe { fscache_cread(fd as i32, hdr.msg_id as u64).unwrap() };
     }
 
+    pub fn cull_cache(&self, blob_id: String) -> Result<()> {
+        //should hold lock during cull_cache, cause we need change pwd when call inuse and cull
+        let children = fs::read_dir(self.cache_dir.lock().unwrap().clone())?;
```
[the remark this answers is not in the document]
kevinXYin: current working dir is a pre-process resource,  so do we need to launch a new process to do cache cull ?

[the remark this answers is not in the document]
jiangliu: That's a little heavy:( @bergwolf what's your thoughts here?
BTW, why does fscache depend on CWD?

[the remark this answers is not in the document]
kevinXYin: The cachefiles kernel APIs depend on CWD ,  inuse and cull cmd only accept file name, not a path. Before call inuse or cull the user daemon should chdir to the dir of blob file.  Maybe we can make these cachefiles kernel API more friendly in the future~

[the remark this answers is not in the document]
kevinXYin: we can not use chdir()  in a multi-threaded or concurrent application, in case of some race conditions. Actually this func will only called by api-server , which is  a single-thread server, and no other func will call chdir(), so I think it is safe, also we do not need a lock. Does this make sense?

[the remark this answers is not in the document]
jiangliu: Let's go that way first. If it turns out to be problem, we can fork a new process to do the cull.

src/bin/nydusd/api_server_glue.rs
kevinXYin: seems it is more more complicated to add delete_blob() to BlobCacheMgr , because we need to send  inuse and cull cmd via cachefiles devfd which is keeped in FsCacheHandler.

jiangliu (approved):

PR #896, opened 2022-11-26
clib: introduce nydus-clib framework

jiangliu:
Introduce nydus-clib framework, which provides static/dynamic to integrate nydus functionality into C programs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -755,9 +755,9 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.126"
+version = "0.2.137"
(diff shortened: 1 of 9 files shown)


imeoer:
Is this a lib to support lazily fetching remote image/file objects using nydus? Can we integrate it into any other languages?

jiangliu:
> Is this a lib to support lazily fetching remote image/file objects using nydus? Can we integrate it into any other languages?

Yes, we hope this library to complete the puzzle:
1) use nydus by in kernel fs EROFS,
2) use nydus by fuse
3) use nydus by library

So need help to define the Nydus API interfaces for applications to integrate Nydus.

PR #900, opened 2022-11-28
improve nydus-test

changweige:
Fix some bugs and enable e2e test for nydusify , remove useless test cases

### DIFF
--- a/contrib/nydus-test/framework/nydusify.py
+++ b/contrib/nydus-test/framework/nydusify.py
@@ -32,6 +32,9 @@ class NydusifyParam(LinuxCommand):
     def fs_version(self, fs_version):
         return self.set_param("fs-version", str(fs_version))
 
+    def target_insecure(self):
+        return self.set_flags("target-insecure")
+
 
 class Nydusify(LinuxCommand):
     def __init__(self, anchor: NydusAnchor):
@@ -46,7 +49,7 @@ class Nydusify(LinuxCommand):
         self.cmd = NydusifyParam(self.nydusify_bin)
         self.cmd.nydus_image(self.image_builder).work_dir(self.work_dir)
 
-    def convert(self, source, suffix="_converted", target_ref=None, fs_version=5):
+    def convert(self, source, suffix="_converted", target_ref=None, fs_version=6):
         """
         A reference to image looks like registry/namespace/repo:tag
         Before conversion begins, split the reference into those parts.
@@ -66,7 +69,10 @@ class Nydusify(LinuxCommand):
                 self.__converted_image,
             )
 
-        self.cmd.source(source).target(target_ref).fs_version(fs_version)
+        self.cmd.source(source).target(target_ref).fs_version(
+            fs_version
+        ).target_insecure()
+
         self.target_ref = target_ref
 
         cmd = str(self.cmd)
--- a/contrib/nydus-test/functional-test/test_nydusify.py
+++ b/contrib/nydus-test/functional-test/test_nydusify.py
@@ -15,6 +15,7 @@ ANCHOR = NydusAnchor()
 
 FS_VERSION = ANCHOR.fs_version
 
+
 @pytest.mark.parametrize(
     "source",
     [
--- a/contrib/nydus-test/functional-test/test_nydusify.py
+++ b/contrib/nydus-test/functional-test/test_nydusify.py
@@ -1,10 +1,8 @@
 import pytest
 import posixpath
 import time
-import platform
 
 from nydus_anchor import NydusAnchor
-from oss import OssHelper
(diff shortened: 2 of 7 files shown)


jiangliu (approved):

contrib/nydus-test/functional-test/test_nydusify.py
```diff
@@ -160,97 +150,7 @@ def test_build_cache(
...
 
 
-@pytest.mark.skip(reason="Get rid of OSS dependency!")
-@pytest.mark.parametrize(
-    "source",
-    [
-        "docker.io/busybox:latest",
-    ],
-)
-def test_upload_oss(
```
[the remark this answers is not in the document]
changweige: In fact, the case is always skipped now. We can drop the changes about removing the case of this PR.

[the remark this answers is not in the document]
bergwolf: Thanks. We should try to enable the test instead. There is OSS AK configured for nydus-test. Let me take a look.

contrib/nydus-test/functional-test/test_nydusify.py
```diff
@@ -335,16 +323,16 @@ def test_cross_platform_multiplatform(
...
+    # cached_layers = [c["digest"] for c in records]
+    # cached_layers.sort()
     # >       assert cached_layers == converted_layers
     # E       AssertionError: assert None == ['sha256:3f18...af3234b4c257']
     # E         +None
     # E         -['sha256:3f18b27a912188108c8590684206bd9da7d81bbfd0e8325f3ef0af3234b4c257']
-    for r in converted_layers:
-        assert r in cached_layers
+    # for r in converted_layers:
+    #     assert r in cached_layers
```
bergwolf, line 335: Any reason why the build cache tests are disabled?

contrib/nydus-test/functional-test/test_nydusify.py
```diff
@@ -13,22 +11,15 @@
 
 ANCHOR = NydusAnchor()
 
-FS_VERSION = ANCHOR.fs_version
```
bergwolf, line 16: Why is `fs_version` ignored?

contrib/nydus-test/framework/nydusify.py
```diff
@@ -46,7 +49,7 @@ def __init__(self, anchor: NydusAnchor):
         self.cmd = NydusifyParam(self.nydusify_bin)
         self.cmd.nydus_image(self.image_builder).work_dir(self.work_dir)
 
-    def convert(self, source, suffix="_converted", target_ref=None, fs_version=5):
+    def convert(self, source, suffix="_converted", target_ref=None, fs_version=6):
```
bergwolf, line 52: Let's drop the default value for `fs_version` to make sure that callers always set it properly.

bergwolf: Same applies to the `check` method.

contrib/nydus-test/functional-test/test_nydusify.py
changweige: It has malfunctioned. In order to run nydusify test suites, I have to disable it first~

contrib/nydus-test/functional-test/test_nydusify.py
```diff
@@ -80,11 +71,11 @@ def test_basic_conversion(
     workload_gen.finish_torture_read()
 
 
+@pytest.mark.skip(reason="build cache has bug trying use https to access repo")
```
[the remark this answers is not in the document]
imeoer: @bergwolf @changweige nydusify always use https to access source/target/build cache registry, and then fallback to plain http if https failure. I noticed that in one case this might not work and it has been fixed in this [commit](https://github.com/dragonflyoss/image-service/pull/966/commits/559e277a9f931516def19db3d13b8c3bd26e83f8), please take a look.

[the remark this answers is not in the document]
changweige: @imeoer 

The problem still exists.

```
INFO[2022-12-30T11:13:18+08:00] [CACH] Check layer                            ChainID="sha256:0310f4cd9792b14930d36723933ce1811a4d0aeb1b5a57984ef49b0cdf8886e5" Time="956.344µs"
WARN[2022-12-30T11:13:18+08:00] Failed to get cache record: Push cached bootstrap layer: failed to do request: Head "https://127.0.0.1:5000/v2/nydus/golang/blobs/sha256:62d40f76fdad58cd102ad1f5f0567ae65d842c629425c429bb90c959c545c993": http: server gave HTTP response to HTTPS client
INFO[2022-12-30T11:13:18+08:00] [SOUR] Mount layer

```

PR #908, opened 2022-11-30
utils: fix a warning generated when decoding chunk data

jiangliu:
The refactor introduced a regression which tries to decompress an uncompressed data chunk, thus generating a warning about decompression failure. Fix it by check whether the chunk data is compressed or not.


liubogithub (approved):
Looks good.

bergwolf (approved):

PR #909, opened 2022-11-30
storage: enhance retryable registry http scheme

imeoer:
Check the tls connection error with `wrong version number` keywords,
it's more reliable than the specific error code.

### DIFF
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -186,16 +186,6 @@ struct RegistryState {
     cached_bearer_auth: ArcSwapOption<BearerAuth>,
 }
 
-fn needs_fallback_http(e: &dyn Error) -> bool {
-    match e.source() {
-        Some(err) => match err.source() {
-            Some(err) => err.to_string().contains("error:1408F10B"),
-            None => false,
-        },
-        None => false,
-    }
-}
-
 impl RegistryState {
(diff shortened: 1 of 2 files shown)


storage/src/backend/registry.rs
```diff
@@ -189,7 +189,14 @@ struct RegistryState {
 fn needs_fallback_http(e: &dyn Error) -> bool {
     match e.source() {
         Some(err) => match err.source() {
-            Some(err) => err.to_string().contains("error:1408F10B"),
+            Some(err) => {
+                let msg = err.to_string().to_lowercase();
+                let fallback = msg.contains("ssl") && msg.contains("record");
```
bergwolf, line 194: This is a literal string match. I wonder if it works with different registries.

imeoer: It's a simplified client implementation referenced in go: https://github.com/golang/go/issues/44855, it seems for different client using different ssl library, we got different error message:

curl:

```
curl https://localhost:5000/v2
curl: (35) error:0A00010B:SSL routines::wrong version number
```

rust reqwest with current openssl lib:

```
error:1408F10B:SSL routines:ssl3_get_record:wrong version number:ssl/record/ssl3_record.c:331
```

jiangliu (approved):

storage/src/backend/registry.rs
imeoer: @bergwolf Updated, PTAL.

PR #912, opened 2022-12-01
storage: simplify implementation of cachefile

jiangliu:
Simplify implementation of cachefile by:
1) remove compressor and digester and directly get those info from
   blob info object.
2) rename is_compressed to is_raw_data, the cache data may be
   uncompressed and encrypted in future.


changweige (approved):
LGTM

PR #921, opened 2022-12-05
Fscache chunkmap restore

kevinXYin:
restore chunkmap from blob file after create FileCacheEntry for fscache mode


jiangliu:
fscache related code should be disabled for macos, it's not supported by macos yet.

storage/src/cache/cachedfile.rs
```diff
@@ -51,33 +51,44 @@ impl FileCacheMeta {
         blob_file: String,
         blob_info: Arc<BlobInfo>,
         reader: Option<Arc<dyn BlobReader>>,
+        sync: bool,
```
jiangliu, line 54: How about abstracting the `sync` related code as a new function `try_open_cached_xxx()`?

kevinXYin:
> fscache related code should be disabled for macos, it's not supported by macos yet.

thanks , will fix this

storage/src/cache/cachedfile.rs
kevinXYin: But we still need to download the meta data , right ?  if introduce a new func seems they are doing same things

storage/src/cache/fscache/mod.rs
```diff
@@ -257,4 +275,69 @@ impl FileCacheEntry {
             prefetch_config,
         })
     }
+
+    //restore index_map from blob file via seek_hole
+    #[cfg(target_os = "linux")]
```
bergwolf, line 280: I wonder if we should disable fscahce on non-linux target OS altogether. fscache is surely a Linux thing.

kevinXYin: Yes, that should be better , but it seems hard to completely isolate the code for fscache and disable it. Any advises? @jiangliu

jiangliu: It would be better to isolate fscache for Linux only. It's ok to move on with this PR and open another issue for conditionally enable fscache for linux only.

jiangliu (approved):

PR #923, opened 2022-12-05
Backoff retry

changweige:
Introduce BackOff delayer to mitigate backend pressure

Don't retry immediately since the registry can return "too many requests" error. We'd better to be slow to retry.

### DIFF
--- a/src/bin/nydusctl/commands.rs
+++ b/src/bin/nydusctl/commands.rs
@@ -401,31 +401,28 @@ Commit:                 {git_commit}
                 if let Some(b) = i.get("backend_collection") {
                     if let Some(fs_backends) = b.as_object() {
                         if !fs_backends.is_empty() {
-                            println!("Backend list:")
+                            println!("Instances:")
                         }
 
                         for (mount_point, backend_obj) in fs_backends {
                             let backend: FsBackendDesc =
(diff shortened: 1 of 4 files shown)


src/bin/nydusctl/commands.rs
```diff
@@ -401,31 +401,28 @@ Commit:                 {git_commit}
...
-            Prefetch:               {enabled}
-            Prefetch Merging Size:  {merging_size}
-        "#,
-                                        meta_mode = rafs_cfg.mode,
-                                        enabled = cache_cfg.prefetch.enable,
-                                        merging_size = cache_cfg.prefetch.batch_size,
+                                        "\tPrefetch Merging Size:  {}",
+                                        cache_cfg.prefetch.batch_size
                                     );
+                                    print!("")
```
imeoer, line 425: Is this necessary?

imeoer (commented):
Others LGTM.

src/bin/nydusctl/commands.rs
changweige: The print() adds a black line as a delimiter between each instance's information. So I think it is necessary since it makes the print and format neater

jiangliu (approved):

Issue #831, opened 2022-11-03
support s3 storage backend


imeoer:
Currently nydus only supports Aliyun OSS, we need to support more object storage backends, for example AWS S3, the following components need to be supported:

- [x] nydusify
- [ ] acceld
- [x] nydusd

loheagn:
Hi, @imeoer, I want to take this issue. Can you assign it to me?

imeoer:
@loheagn Cool, let's move forward!

CSUDaXiang:
hello,Have you made any progress on this issue recently?

loheagn:
Hi, @imeoer @CSUDaXiang , I opened a PR to add s3 support to `nydusify convert`. For more information, please refer to the PR #871.

The next target is to add s3 support to nydusd.

PR #925, opened 2022-12-06
storage: nydusd supports s3 backend

loheagn:
As we have supported nydusify using s3 services as the storage backend to store blobs, this patch supports nydusd fetching blobs from s3 services.

By now, using s3 backend, we can use `nydusify convert` to convert an oci image to nydus image and use `nydusify check` to verify the conversion.

Signed-off-by: Nan Li <loheagn@icloud.com>

### DIFF
--- a/storage/src/backend/mod.rs
+++ b/storage/src/backend/mod.rs
@@ -21,14 +21,22 @@ use nydus_utils::metrics::{BackendMetrics, ERROR_HOLDER};
 use crate::utils::{alloc_buf, copyv};
 use crate::StorageError;
 
-#[cfg(any(feature = "backend-oss", feature = "backend-registry"))]
+#[cfg(any(
+    feature = "backend-oss",
+    feature = "backend-registry",
+    feature = "backend-s3"
+))]
 pub mod connection;
 #[cfg(feature = "backend-localfs")]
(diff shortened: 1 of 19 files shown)


loheagn:
We can support other nydusify subcommands (such as `build`) to use s3 backend and complete the documents in the next pr.

loheagn:
update #831

storage/src/backend/mod.rs
```diff
@@ -16,19 +16,28 @@
 use std::sync::Arc;
 
 use fuse_backend_rs::file_buf::FileVolatileSlice;
+
```
jiangliu, line 19: please remove this blank line:)

storage/src/backend/object_storage.rs
```diff
@@ -0,0 +1,214 @@
+use std::fmt::Debug;
```
jiangliu, line 1: Please help to add a license header here.

jiangliu:
Great work!
Only several syntax related comments.

loheagn:
> Great work! Only several syntax related comments.

@jiangliu Thanks for the review! All the comments are fixed.

hsiangkao:
Would you mind updating README.md as well?

PR #934, opened 2022-12-08
Fix 2.1 broken compressed cache

changweige:
From my test, the compressed local blob cache can no longer work on branch v2.1. This PR fixes the feature and tries to improve runtime efficiency a bit.


imeoer (approved):

changweige:
Please merge this after v2.1.2 is released, it is not ready to get into v2.1.2

PR #953, opened 2022-12-11
Misc code cleanup

jiangliu:
Misc code cleanup

### DIFF
--- a/src/bin/nydus-image/core/feature.rs
+++ b/src/bin/nydus-image/core/feature.rs
@@ -44,7 +44,7 @@ impl TryFrom<&str> for Feature {
 
     fn try_from(f: &str) -> std::result::Result<Self, Self::Error> {
         match f {
-            "blob_toc" => Ok(Self::BlobToc),
+            "blob-toc" => Ok(Self::BlobToc),
             _ => bail!(
                 "{} `{}`, please try upgrading to the latest nydus-image",
                 ERR_UNSUPPORTED_FEATURE,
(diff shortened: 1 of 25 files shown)


storage/src/meta/toc.rs
```diff
@@ -335,56 +333,54 @@ impl TocEntryList {
...
         if entry_type != EntryType::Regular {
-            return Err(anyhow!("Tar entry type for ToC is not a regular file"));
+            return Err(eother!("Tar entry type for ToC is not a regular file"));
         }
         let entry_size = header
             .entry_size()
-            .map_err(|_| anyhow!("failed to get entry size from tar header"))?;
+            .map_err(|_| eother!("failed to get entry size from tar header"))?;
         if entry_size != size as u64 {
-            return Err(anyhow!(
```
changweige, line 370: Why not use `anyhow`

src/bin/nydus-image/core/feature.rs
```diff
@@ -45,6 +45,7 @@ impl TryFrom<&str> for Feature {
     fn try_from(f: &str) -> std::result::Result<Self, Self::Error> {
         match f {
             "blob_toc" => Ok(Self::BlobToc),
```
imeoer, line 47: Snapshotter converter has changed to `blob-toc`, we can remove it safely.

storage/src/meta/toc.rs
imeoer: It seems the `anyhow` crate only be introduced in `nydus-image`.

src/bin/nydus-image/core/feature.rs
imeoer: @jiangliu It seems the `blob_toc` still be checked in `src/bin/nydus-image/main.rs`.

jiangliu: Fixed by changing to "blob-toc".

PR #960, opened 2022-12-14
storage: use async runtime to download blobs meta info

changweige:
For a image with 100 layers and 1000 nodes in K8s cluster, 100 thousands requests are poured into registry and object storage system. The pressure is too high. So use existing tokio runtime with limited threads to download those blob metas

### DIFF
--- a/storage/src/cache/fscache/mod.rs
+++ b/storage/src/cache/fscache/mod.rs
@@ -237,6 +237,7 @@ impl FileCacheEntry {
                 blob_file_path,
                 blob_info.clone(),
                 Some(rafs_blob_reader),
+                None,
                 true,
             )?
         } else {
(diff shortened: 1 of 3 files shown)


storage/src/cache/fscache/mod.rs
```diff
@@ -237,6 +237,7 @@ impl FileCacheEntry {
                 blob_file_path,
                 blob_info.clone(),
                 Some(rafs_blob_reader),
+                None,
```
imeoer, line 240: Why pass `None` here?

changweige: Seems fscache needs to download blob metas synchronously

imeoer: Ok, it seems implemented in `init_blobcache`, maybe we can also migrate to async runtime later.

imeoer (approved):

PR #961, opened 2022-12-15
rafs: fix overflow panic of rafs v6 lookup

mofishzz:
The directory for v6 is stored in the following way： ...name1name2
The first subdirectory we got here is ".".

When looking for some file whose ascii value is less than "." like "*", we need to move forward to the block index -1. This caused uszie's index "pirvot" to attempt to subtract with overflow and then panic.

Fixes: 50ca1a1 ("rafs: optimize entry search in rafs v6")
Signed-off-by: Huang Jianan <jnhuang@linux.alibaba.com>


changweige (approved):
Looks good

PR #965, opened 2022-12-18
Add more methods to handle ToC

jiangliu:
Add more methods to handle ToC

### DIFF
--- a/storage/src/meta/toc.rs
+++ b/storage/src/meta/toc.rs
@@ -425,7 +425,7 @@ impl TocEntryList {
 
     fn parse(buf: &[u8], size: usize, location: &TocLocation) -> Result<Self> {
         if location.validate_digest {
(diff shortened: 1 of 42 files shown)


storage/src/meta/toc.rs
```diff
@@ -397,17 +476,138 @@ impl TocEntryList {
...
+                let p = path
+                    .as_ref()
+                    .to_path_buf()
+                    .with_extension("toc_downloading");
+                let mut file = OpenOptions::new()
+                    .create(true)
+                    .write(true)
+                    .truncate(true)
+                    .open(p.as_path())?;
+                bootstrap.extract_from_reader(reader.clone(), &mut file)?;
```
imeoer, line 514: Is the file leak if failed here?

jiangliu: fixed

PR #966, opened 2022-12-19
nydusify: some minor fixups

imeoer:
- nydusify: fix panic if only --target be specified
- nydusify: fix a http fallback case for build cache


bergwolf (approved):
lgtm!

PR #970, opened 2022-12-21
nydusify: refactor with converter package

imeoer:
```
Refactor converter workflow using the converter package in
acceld project:

https://github.com/goharbor/acceleration-service/pull/84

This allows nydusify to reuse acceld/snapshotter codes and
to convert layers concurrently, as well as to support the
zran conversion later.

This patch causes nydusify to rely on the containerd daemon
service, we will remove this restriction in the next commit.
```


liubogithub (commented):
why do we copy containerd.fetch() and containerd.pull()?

liubogithub (approved):

imeoer:
> why do we copy containerd.fetch() and containerd.pull()?

The part of codes is not exported, so we must port it.

Issue #971, opened 2022-12-21
support local disk backend 


adamqqqplay:
Currently nydus supports 4 different backends, they are registry, localfs, oss and s3, which can already meet the existing needs. But in Confidential Computing or Hybrid Image scenarios, containers may need to read data from block devices, so we could add local disk backends to nydus to support more scenarios.

The following components need to be supported:

- [ ]  nydusify
- [ ]  acceld
- [ ]  nydusd
- [ ]  nydus-snapshotter

imeoer:
If there is no user demand for the time being, let's close it first.

PR #972, opened 2022-12-21
storage: introduce new backend localdisk in nydusd

adam3q:
The localdisk backend adds support for storing images in disks. In this scenario, each layer of the blob is stored in partitions, and multiple partitions are addressed in the local raw disk via the GUID partition table (GPT), which means that this disk stores the entire image.

Currently, generating a localdisk image through `nydusify` is not supported for the time being. You need to use the `nydus-localdisk` tool to complete this step.
Test document located at: https://github.com/adamqqqplay/nydus-localdisk/blob/master/README.md

This PR updates: #971 

Signed-off-by: Qinqi Qu <quqinqi@linux.alibaba.com>


storage/src/backend/localdisk.rs
```diff
@@ -0,0 +1,246 @@
...
+                partition.length
+            );
+            table_guard.insert(name, partition);
+        }
+
+        Ok(())
+    }
+
+    // Disk names in GPT tables cannot store full 64-byte blob ids, so we should truncate them to 32 bytes.
+    fn truncate_blob_id(blob_id: &str) -> LocalDiskResult<&str> {
```
changweige, line 209: How about storing binary rather than strings? Then 32 bytes can hold the whole blob id

changweige:
How to create partitions filled with blobs and run with this PR?

adamqqqplay:
> How to create partitions filled with blobs and run with this PR?

https://github.com/adamqqqplay/nydus-localdisk
You could use this small tools to generate a localdisk image file and then mount it to the server via a loop device. I'll be porting this feature to nydusify shortly.

storage/src/backend/localdisk.rs
adamqqqplay: > How about storing binary rather than strings? Then 32 bytes can hold the whole blob id

Very good suggestion, I will try to modify this.

adamqqqplay: I tried to use the storage method of Hex to String, but there are still problems in the implementation, and I will improve it later.

In the future, I will use two different methods to obtain the blob_id in get_blob() to maintain compatibility with different versions of the localdisk image.

storage/src/backend/localdisk.rs
```diff
@@ -0,0 +1,273 @@
...
```
imeoer, line 209: If the blob id is truncated in GPT table, how do we ensure no conflicts between different blobs with the same sha256 prefix?

docs/nydusd.md
```diff
@@ -151,6 +151,25 @@ We are working on enabling cloud-hypervisor support for nydus.
 }
 ```
 
+##### Localdisk Backend
+
```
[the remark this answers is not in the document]
adamqqqplay: Ok, I'll add more description here.

storage/src/backend/localdisk.rs
adamqqqplay: When the output of SHA-256 is truncated to 128 bits, its collision resistance is still 2^64, which is still enough to prevent blob collisions. And in the SHA standard, the result is also allowed to be truncated.

Please refer to:
https://security.stackexchange.com/questions/34796/truncating-the-output-of-sha256-to-128-bits
https://security.stackexchange.com/questions/72673/how-bad-is-it-to-truncate-a-hash/72685#72685

Despite this, I'll try to store the full blob id in the future, while maintaining compatibility with images generated in the past.

adamqqqplay: > If the blob id is truncated in GPT table, how do we ensure no conflicts between different blobs with the same sha256 prefix?

Problem fixed in latest commit. Now the full 64-bit layer digest (blob id) stored in the Partition Name and Partition GUID respectively.

At the same time, forward and backward compatibility is maintained.

docs/nydusd.md
```diff
@@ -151,6 +151,31 @@ We are working on enabling cloud-hypervisor support for nydus.
 }
 ```
 
+##### Localdisk Backend
```
imeoer, line 154: Maybe mark this `Localdisk Backend (Experimental)` before we have supported it in nydusify.

adamqqqplay: > Maybe mark this `Localdisk Backend (Experimental)` before we have supported it in nydusify.

OK, that's good.

imeoer (approved):

PR #974, opened 2022-12-23
storage: add new backend `HttpProxy`

loheagn:
This pr adds a new storage backend `HttpProxy` which can access blobs through a http proxy server.

The http proxy server can be local (using unix socket) or remote (using `https://` or using `http://`).

`HttpProxy` uses two API endpoints to access the blobs:
- `HEAD /path/to/blobs` to get the blob size
- `GET /path/to/blobs` to read the blob

The http proxy server should respect [the `Range` header](https://www.rfc-editor.org/rfc/rfc9110.html#name-range) to compute the offset and length of the blob.

The example config files for this new backend may be:

```jsonc
// for remote usage
{
  "backend": {
    "type": "http-proxy",
    "config": {
      "addr": "http://127.0.0.1:9977",
      "path": "/namespace/<repo>/blobs"
    }
  }
}
```

or

```jsonc
// for local unix socket
{
  "backend": {
    "type": "http-proxy",
    "config": {
      "addr": "/path/to/unix/socket/file"
    }
  }
}
```

There is also a test in `http_proxy.rs` to make sure `HttpProxy` works well, which setups a simple http server and generates a `HttpProxy` backend to get contents from the server.

Signed-off-by: Nan Li <loheagn@icloud.com>

### DIFF
--- a/api/src/config.rs
+++ b/api/src/config.rs
@@ -238,6 +238,8 @@ pub struct BackendConfigV2 {
     pub s3: Option<S3Config>,
     /// Configuration for container registry backend.
     pub registry: Option<RegistryConfig>,
+    /// Configuration for local http proxy.
+    pub http_proxy: Option<HttpProxyConfig>,
 }
 
 impl BackendConfigV2 {
@@ -271,6 +273,28 @@ impl BackendConfigV2 {
             "registry" => match self.registry.as_ref() {
                 Some(v) => {
                     if v.host.is_empty() || v.repo.is_empty() {
+                        return false;
+                    }
+                }
+                None => return false,
+            },
+
+            "http-proxy" => match self.http_proxy.as_ref() {
+                Some(v) => {
+                    let is_valid_unix_socket_path = |path: &str| {
+                        let path = Path::new(path);
+                        path.is_absolute() && path.exists()
+                    };
+                    if v.addr.is_empty()
+                        || !(v.addr.starts_with("http://")
+                            || v.addr.starts_with("https://")
+                            || is_valid_unix_socket_path(&v.addr))
+                    {
+                        return false;
+                    }
+
+                    // check if v.path is valid url path format
+                    if Path::new(&v.path).join("any_blob_id").to_str().is_none() {
                         return false;
                     }
                 }
@@ -323,6 +347,17 @@ impl BackendConfigV2 {
             self.registry
                 .as_ref()
                 .ok_or_else(|| einval!("no configuration information for registry"))
+        }
+    }
+
+    /// Get configuration information for http proxy
+    pub fn get_http_proxy_config(&self) -> Result<&HttpProxyConfig> {
+        if &self.backend_type != "http-proxy" {
+            Err(einval!("backend type is not 'http-proxy'"))
+        } else {
+            self.http_proxy
+                .as_ref()
+                .ok_or_else(|| einval!("no configuration information for http-proxy"))
         }
     }
 }
@@ -407,6 +442,35 @@ pub struct S3Config {
     /// S3 secret
     #[serde(default)]
     pub access_key_secret: String,
+    /// Skip SSL certificate validation for HTTPS scheme.
+    #[serde(default)]
+    pub skip_verify: bool,
+    /// Drop the read request once http request timeout, in seconds.
+    #[serde(default = "default_http_timeout")]
+    pub timeout: u32,
+    /// Drop the read request once http connection timeout, in seconds.
+    #[serde(default = "default_http_timeout")]
+    pub connect_timeout: u32,
+    /// Retry count when read request failed.
+    #[serde(default)]
+    pub retry_limit: u8,
+    /// Enable HTTP proxy for the read request.
+    #[serde(default)]
+    pub proxy: ProxyConfig,
+    /// Enable mirrors for the read request.
+    #[serde(default)]
+    pub mirrors: Vec<MirrorConfig>,
+}
+
+/// Http proxy configuration information to access blobs.
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+pub struct HttpProxyConfig {
+    /// Address of http proxy server, like `http://xxx.xxx` or `https://xxx.xxx` or `/path/to/unix.sock`.
+    pub addr: String,
+    /// Path to access the blobs, like `/<_namespace>/<_repo>/blobs`.
+    /// If the http proxy server is over unix socket, this field will be ignored.
+    #[serde(default)]
+    pub path: String,
     /// Skip SSL certificate validation for HTTPS scheme.
     #[serde(default)]
     pub skip_verify: bool,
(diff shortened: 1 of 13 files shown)


imeoer:
Thanks @loheagn for the contribution, it can make `nydus-image unpack` faster!

My suggestion is that maybe we should implement the content store backend with HTTP over unix domain socket instead of TCP, which is simpler and requires only implementing two distribution APIs:

- `HEAD` blob size;
- `GET` blob data (HTTP range get);

It's very like `contrib/nydus-backend-proxy` server (it receives HTTP requests and proxy to a local blob directory).

api/src/config.rs
```diff
@@ -427,6 +441,14 @@ pub struct S3Config {
     pub mirrors: Vec<MirrorConfig>,
 }
 
+/// ContentProxyProxy configuration information to access blobs.
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+pub struct LocalHttpProxyConfig {
```
jiangliu, line 446: How about `LocalUdsConfig`?
We can define an APIs to retrieve blobs from Unix domain socket, no matter it's containerd content store or something else.

jiangliu: Or `ContainerdContentStoreProxy`? @imeoer

loheagn: But this backend driver only support http protocol now. I think `LocalHttpProxy` may be better.

imeoer: It's ok for me if using `LocalHttpProxy`, maybe we can merge the feature with [nydus-backend-proxy](https://github.com/dragonflyoss/image-service/tree/master/contrib/nydus-backend-proxy) (not in this PR)? So that we can use `LocalHttpProxy` to request an HTTP server that can proxy to local blob directory/containerd content store.

api/src/config.rs
```diff
@@ -427,6 +441,14 @@ pub struct S3Config {
...
+    pub socket_path: String,
```
imeoer, line 447: Maybe we can name it `addr`? so that we can support `http://`.

storage/Cargo.toml
```diff
@@ -16,6 +16,8 @@ hex = "0.4.3"
 hmac = { version = "0.12.1", optional = true }
 http = { version = "0.2.8", optional = true }
 httpdate = { version = "1.0", optional = true }
+hyper = {version = "0.14.11", optional =  true}
```
imeoer, line 19: Is this necessary? Is `reqwest` haven't supported unix domain socket?

storage/src/backend/http_proxy.rs
```diff
@@ -0,0 +1,541 @@
...
+                    .map_err(|e| HttpProxyError::ReadResponseBody(e).into())
+                    .map(|bytes| bytes.to_vec()),
+                Err(e) => Err(e),
+            }
+        })
+    }
+}
+
+impl BlobReader for HttpProxyReader {
+    // Note! If the blob is compressed, the function will always return 0
```
imeoer, line 195: Is this comment useful?

storage/Cargo.toml
loheagn: Yes. `reqwest` has not supported unix socket yet. The issue https://github.com/seanmonstar/reqwest/issues/39 is still open.

api/src/config.rs
```diff
@@ -427,6 +462,34 @@ pub struct S3Config {
     pub mirrors: Vec<MirrorConfig>,
 }
 
+/// Http proxy configuration information to access blobs.
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+pub struct HttpProxyConfig {
+    /// Address of http proxy server, like `http://xxx.xxx` or `https://xxx.xxx` or `path/to/unix.sock`.
```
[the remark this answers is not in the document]
loheagn: Yes. If the user chooses to use unix socket, the `addr` should be a unix path 
string like `/path/to/unix.sock`. The whole backend config may be like:

```json
{
  "backend": {
    "type": "http-proxy",
    "config": {
      "addr": "/path/to/unix/socket/file"
    }
  }
}
```

api/src/config.rs
```diff
@@ -325,6 +349,17 @@ impl BackendConfigV2 {
                 .ok_or_else(|| einval!("no configuration information for registry"))
         }
     }
+
+    /// Get configuration information for http proxy
+    pub fn get_http_proxy_config(&self) -> Result<&HttpProxyConfig> {
+        if &self.backend_type != "http-proxy" {
```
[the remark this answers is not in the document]
loheagn: I think both are ok. But for some personal reasons, I prefer `http-proxy`.

storage/src/backend/http_proxy.rs
```diff
@@ -0,0 +1,541 @@
+// Copyright 2022 Ant Group. All rights reserved.
```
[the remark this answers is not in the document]
loheagn: OK.

storage/src/backend/http_proxy.rs
loheagn: Yes, it's useful. If the blobs are comporessed, we cannnot get the real blob size before we complete the stream reading.

Here is an example in nydus-snapshotter https://github.com/containerd/nydus-snapshotter/blob/3d37ee2f6834ca1626abcd69063f96e1d69d397f/pkg/converter/convert_unix.go#L233.

api/src/config.rs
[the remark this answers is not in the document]
imeoer: Okay, let's update the comment

api/src/config.rs
```diff
@@ -276,6 +278,28 @@ impl BackendConfigV2 {
                 }
                 None => return false,
             },
+
+            "http_proxy" => match self.http_proxy.as_ref() {
```
imeoer, line 282: Can we update to `http-proxy`?

loheagn: Oh, my bad...

imeoer:
Please also add SOB for the commits.

imeoer:
Hi @loheagn, Could also update the doc in [here](https://github.com/dragonflyoss/image-service/blob/master/docs/nydusd.md#use-different-storage-backends)? Thanks!

loheagn:
> Hi @loheagn, Could also update the doc in [here](https://github.com/dragonflyoss/image-service/blob/master/docs/nydusd.md#use-different-storage-backends)? Thanks!

No problem. The docs were updated.

docs/nydusd.md
```diff
@@ -293,6 +293,51 @@ Currently, the mirror mode is only tested in the registry backend, and in theory
...
+      }
+    }
+  }
+}
+```
+
+or
+
+```
+// for remote usage
```
imeoer, line 326: Should be `for local usage` here.

loheagn: fixed :)

changweige:
need rebase to master since anolis test script is updated

imeoer (approved):

PR #979, opened 2022-12-29
rafs: add a configuration warmup to make window for prefetching

changweige:
Thus let rafs have a bigger chance to directly hit the local file cache to reduce the Registry request.

From the test results, the container startup has the same latency with much fewer requests to the registry

- with warmup enabled
```
{"repo": "golang", "bench": "golang:latestnydusv6", "pull_time": " 1.867576", "create_time": " 2.041314", "run_time": " 0.466915", "elapsed": " 4.375805"}

Backend Type:       "registry"
Read Amount:        355171173 Bytes (338.71762561798096 MB)
Read Count:         179
Read Errors:        0

Block Sizes:             <1K     1K~     4K~     16K~    64K~    128K~   512K~   1M~
Average Latency(millis): 0       156     52      32      57      65      32      56

Block Sizes:             <1K     1K~     4K~     16K~    64K~    128K~   512K~   1M~
Request Count:           0       1       1       2       2       3       2       168
```

- with warmup disabled
```
{"repo": "golang", "bench": "golang:latestnydusv6", "pull_time": " 1.464904", "create_time": " 0.608780", "run_time": " 2.441874", "elapsed": " 4.515558"}

Backend Type:       "registry"
Read Amount:        355171173 Bytes (338.71762561798096 MB)
Read Count:         747
Read Errors:        0

Block Sizes:             <1K     1K~     4K~     16K~    64K~    128K~   512K~   1M~
Average Latency(millis): 9       10      13      20      108     28      23      64

Block Sizes:             <1K     1K~     4K~     16K~    64K~    128K~   512K~   1M~
Request Count:           308     172     57      18      4       16      6       166
```

### DIFF
--- a/api/src/config.rs
+++ b/api/src/config.rs
@@ -586,6 +586,8 @@ pub struct RafsConfigV2 {
     /// Filesystem prefetching configuration.
     #[serde(default)]
     pub prefetch: PrefetchConfigV2,
+    #[serde(default)]
+    pub warmup: u64,
 }
 
 impl RafsConfigV2 {
@@ -908,9 +910,13 @@ struct RafsConfig {
     /// Record file name if file access trace log.
     #[serde(default)]
     pub latest_read_files: bool,
-    // ZERO value means, amplifying user io is not enabled.
+    /// ZERO value means, amplifying user io is not enabled.
     #[serde(default = "default_batch_size")]
     pub amplify_io: usize,
+    /// Let Rafs wait for a period before completing mount. So prefetch
+    /// can download data as much as possible. In millisecond.
+    #[serde(default)]
+    pub warmup: u64,
 }
 
 impl TryFrom<RafsConfig> for ConfigV2 {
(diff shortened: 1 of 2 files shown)


imeoer:
Good point, but how long is a good time to wait?

changweige:
> Good point, but how long is a good time to wait?

It relates to end-users' production environment and deployment. By test, I found that 1800ms best fits in our production environment

jiangliu:
Instead of sleep for a while, how about adding interface to `wait for those blob meta to be ready`?

changweige:
> Instead of sleep for a while, how about adding interface to `wait for those blob meta to be ready`?

The total time consumed by downloading blobs' metadata could be very short if the registry is under low pressure and the image has a small number of layers, which means the `wait` will be complete very soon. I am afraid that nydusd does not have enough time to download data by prefetch

jiangliu:
Sorry, hope to work this after 2.2-rc0 is out.
I think we may use a condvar to synchronize meta downloading status instead of a sleep.

PR #987, opened 2023-01-04
read prefetch list(runtime) from file

Desiki-high:
change the nydusd  arg --prefetch as a file path, we can read the prefetch list from the file. For example:
we can use `--prefetch-files /tmp/test` and the `/tmp/test` is like 
```
/etc
/bin
/src/test.py
```
When nydusd is mounted, this file is read by line to form a prefetch list

### DIFF
--- a/api/openapi/nydus-api-v1.yaml
+++ b/api/openapi/nydus-api-v1.yaml
@@ -348,10 +348,8 @@ components:
           description: usually to be the metadata source
           type: string
         prefetch_files:
-          description: files that need to be prefetched
-          type: array
-          items:
-            type: string
+          description: local file path which recorded a file list to be prefetched
+          type: string
         config:
           description: inline request, use to configure fs backend.
           type: string
--- a/api/openapi/nydus-api-v1.yaml
+++ b/api/openapi/nydus-api-v1.yaml
@@ -348,7 +348,7 @@ components:
           description: usually to be the metadata source
           type: string
         prefetch_files:
-          description: local file path which recorded a file list to be prefetched
+          description: local file path which recorded files/directories to be prefetched and separated by newlines
           type: string
         config:
           description: inline request, use to configure fs backend.
--- a/contrib/nydus-test/functional-test/test_nydus.py
+++ b/contrib/nydus-test/functional-test/test_nydus.py
@@ -582,7 +582,7 @@ def test_shared_blobcache(nydus_anchor: NydusAnchor, nydus_image, rafs_conf: Raf
             .apisock(tempfile.NamedTemporaryFile().name)
             .prefetch_files(os.path.abspath(tmp_file))
             .set_mountpoint(mountpoint)
-        )  
+        )
         return rafs
 
     cases = []
@@ -613,7 +613,7 @@ def test_shared_blobcache(nydus_anchor: NydusAnchor, nydus_image, rafs_conf: Raf
 
     for case in cases:
         case[0].umount()
-    
+
 
 # @pytest.mark.skip(reason="ECS can't pass this case!")
 @pytest.mark.parametrize("sig", [signal.SIGTERM, signal.SIGINT])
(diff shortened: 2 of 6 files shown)


src/bin/nydusd/main.rs
```diff
@@ -220,10 +220,10 @@ fn append_fs_options(app: Command) -> Command {
     .arg(
         Arg::new("prefetch-files")
             .long("prefetch-files")
-            .help("List of files/directories to prefetch")
+            .help("A file include the list of files/directories to prefetch")
```
changweige, line 223: "A file including xxx"

changweige (commented):
Is it a breaking change? Please also update the test cases in nydus-test.

api/openapi/nydus-api-v1.yaml
```diff
@@ -348,10 +348,8 @@ components:
           description: usually to be the metadata source
           type: string
         prefetch_files:
-          description: files that need to be prefetched
-          type: array
-          items:
-            type: string
+          description: local file path which recorded a file list to be prefetched
```
bergwolf, line 351: Better be explicit that the list of file paths is separated with newlines.

jiangliu (approved):

jiangliu:
It's a break change, any comments from @changweige ?

jiangliu:
It may be worth to set a upper limit on prefetch-list-file?

changweige:
> It's a breaking change, any comments from @changweige ?

Though I am not a big fan of changing the semantics of nydusd arguments, it still looks good. Hopefully, there is no user enormously depending on the old semantics

contrib/nydus-test/functional-test/test_nydus.py
```diff
@@ -564,7 +564,7 @@ def test_pseudo_fs(nydus_anchor, nydus_image: RafsImage, rafs_conf: RafsConf):
     nc.umount_rafs("/pseudo3")
 
 
-def test_shared_blobcache(nydus_anchor: NydusAnchor, nydus_image, rafs_conf: RafsConf):
+def test_shared_blobcache(nydus_anchor: NydusAnchor, nydus_image, rafs_conf: RafsConf, tmp_path):
```
changweige, line 567: We can't add a parameter to the test function if it is not a fixture. Did you successfully run the test?

contrib/nydus-test/functional-test/test_nydus.py
```diff
@@ -573,13 +573,16 @@ def test_shared_blobcache(nydus_anchor: NydusAnchor, nydus_image, rafs_conf: Raf
...
+
     def make_rafs(mountpoint):
         rafs = (
             NydusDaemon(nydus_anchor, nydus_image, rafs_conf)
             .apisock(tempfile.NamedTemporaryFile().name)
-            .prefetch_files("/")
+            .prefetch_files(os.path.abspath(tmp_file))
             .set_mountpoint(mountpoint)
-        )
+        )
```
changweige, line 585: Please trim trailing whitespace, perhaps try some kind of formater

contrib/nydus-test/functional-test/test_nydus.py
```diff
@@ -747,10 +751,14 @@ def test_specified_prefetch(
     prefetching_files.append("/a/b/c/d")
     prefetching_files.append(os.path.join("/", "f/g/h/"))
 
-    specified_files = " ".join([os.path.join("/", d) for d in prefetching_files])
+    tmp_file = tmp_path / "prefetch.txt"
+    with open(tmp_file, "w") as f:
+        for d in prefetching_files:
+            f.write(os.path.join("/", d) + "\n")
```
changweige, line 757: This can be simplified by using `writelines`

changweige:
Otherwise, looks good to me

contrib/nydus-test/functional-test/test_nydus.py
Desiki-high: Yes,I run these test functions successfully . But I'm not familiar with pytest especially the fixture, so I'm not sure.

Issue #990, opened 2023-01-06
make v2.2 (master branch) stable


imeoer:
Issues that need to be fixed:

- [x] `nydus-image subcommand` not work with `--bootstrap`, for example `nydus-image check --bootstrap`
- [x] `nydus-image` Tar/TargzToRafs losts file xattrs
- [x] zran TarToRef not works
- [x] nydusd throws error if image chunk size > 1MB (not reproduced)
- [x] test basic compatibility with v2.1, v0.1.0 binaries
- [x] prefetch module throws some errors:

```
[2023-01-14 03:18:12.418347 +00:00] ERROR [/src/error.rs:22] Error:
	"entry not found index 1 chunk_info_array.len 2, last_end 0x24, end 0x224, blob compressed size 0x224"
	at storage/src/meta/mod.rs:1385
	note: enable `RUST_BACKTRACE=1` env to display a backtrace
```

- [x] nydusd throws error if run `nginx:latest-v5` image built by `nydus-image v0.1.0` + `nydusify v0.1.0`:

```
[2023-01-14 03:26:41.377899 +00:00] ERROR [/src/error.rs:22] Error:
	"data digest value doesn't match"
	at storage/src/cache/mod.rs:359
	note: enable `RUST_BACKTRACE=1` env to display a backtrace
```
```
It's an false alarm for old images with digested chunk map.
```

- [x] oss backend broken: invalid signature

- [x] nydusd panic if run zran image `nginx:latest-nydus-zran`:

```
thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 6 but the index is 18446744073709551615': storage/src/cache/cachedfile.rs:721
``` 

We can reproduce aboves in this PR https://github.com/dragonflyoss/image-service/pull/983 (with `cd smoke && make test`)

CIs need to be fixed:

- [x] [convert ci](https://github.com/dragonflyoss/image-service/actions/workflows/convert.yml) broken
- [x] [nydus test ci](https://github.com/dragonflyoss/image-service/actions/workflows/it.yml) broken
- [x] [nydus release ci](https://github.com/dragonflyoss/image-service/actions/workflows/release.yml) broken

jiangliu:
Per my tests, "nydus-image check --bootstrap bootstrap" and "nydus-image check bootstrap" works as expected.
But "nydus-image inspect --bootstrap bootstrap" doesn't work because we don't support `--bootstrap` for inspect, it has been changed to `nydus-image inspect bootstrap`

PR #994, opened 2023-01-07
nydus-image: fix a bug in handle xattr for tarballs

jiangliu:
There's a bug in xattr handling for tarball which causing losing of xattrs.

Fixes: https://github.com/dragonflyoss/image-service/issues/990


imeoer (approved):
LGTM!

PR #995, opened 2023-01-08
Enhance the way to handle `--bootstrap` option for nydus-image

jiangliu:
Enhance the way to handle `--bootstrap` option for nydus-image

### DIFF
--- a/src/bin/nydus-image/main.rs
+++ b/src/bin/nydus-image/main.rs
@@ -9,7 +9,6 @@ extern crate clap;
 extern crate anyhow;
 #[macro_use]
 extern crate log;
-extern crate serde;
 #[macro_use]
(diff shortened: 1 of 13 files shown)


src/bin/nydus-image/main.rs
```diff
@@ -848,13 +849,6 @@ impl Command {
         event_tracer!("euid", "{}", geteuid());
         event_tracer!("egid", "{}", getegid());
 
-        // Validate output bootstrap file
-        if !blob_inline_meta {
```
imeoer, line 842: It's ok to disable validation for create subcommand, but we'd better do `nydus-image check` after `nydus-image merge` in the converter tool.

imeoer (approved):

PR #1014, opened 2023-01-17
Fix a bug in rafs v6 introduced during code refactor

jiangliu:
Fix a bug in rafs v6 introduced during code refactor, which calculates wrong offset for inode with xattrs.


hsiangkao (approved):

hsiangkao:
May also need @liubogithub to look at this as well.

PR #1017, opened 2023-01-18
rafs: fix a underflow bug in rafs v6 implementation

jiangliu:
Fix a underflow bug in find_target_block() of RAFS v6. `last` is usize instead of isize so it may underflow.

### DIFF
--- a/rafs/src/metadata/direct_v6.rs
+++ b/rafs/src/metadata/direct_v6.rs
@@ -640,16 +640,15 @@ impl OndiskInodeWrapper {
         Ok(chunks)
     }
 
-    fn find_target_block(&self, name: &OsStr) -> Result<usize> {
+    fn find_target_block(&self, name: &OsStr) -> Result<Option<usize>> {
         let inode = self.disk_inode();
         if inode.size() == 0 {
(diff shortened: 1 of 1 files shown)


rafs/src/metadata/direct_v6.rs
```diff
@@ -640,16 +640,15 @@ impl OndiskInodeWrapper {
         Ok(chunks)
     }
 
-    fn find_target_block(&self, name: &OsStr) -> Result<usize> {
+    fn find_target_block(&self, name: &OsStr) -> Result<Option<usize>> {
         let inode = self.disk_inode();
         if inode.size() == 0 {
-            return Err(enoent!());
```
hsiangkao, line 646: so when we get Err()?

jiangliu: There are several `map_err()?` in the function body.

jiangliu: And the caller `get_child_by_name()` while return `enoent()` if `find_target_block()` returns `Ok(None)`.

hsiangkao (approved):

PR #1021, opened 2023-01-19
smoke: DescartesIterator must return item when hasNext() returns true

ccx1024cc:
In order to return Item from `Next()` when `hasNext()` returns true, DescartesIterator stores the information of next item, which involves user-customized `skip()` function. It updates information only when `Next()` is called.

Signed-off-by: 泰友 <cuichengxu.ccx@antgroup.com>


jiangliu (approved):

imeoer (approved):
LGTM!

PR #1022, opened 2023-01-19
storage: avoid file leakage in error handling path

jiangliu:
If error happens during preparing chunkmap and blob.meta files for data blobs, some files may be left over. So change code to avoid file leakages.

### DIFF
--- a/src/bin/nydus-image/main.rs
+++ b/src/bin/nydus-image/main.rs
@@ -237,6 +237,7 @@ fn prepare_cmd_args(bti_string: &'static str) -> App {
                     Arg::new("blob-offset")
                         .long("blob-offset")
                         .help("File offset to store RAFS data, to support storing data blobs into tar files")
+                        .hide(true)
                         .default_value("0"),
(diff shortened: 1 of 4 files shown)


src/bin/nydus-image/main.rs
```diff
@@ -237,6 +237,7 @@ fn prepare_cmd_args(bti_string: &'static str) -> App {
                     Arg::new("blob-offset")
                         .long("blob-offset")
                         .help("File offset to store RAFS data, to support storing data blobs into tar files")
+                        .hide(true)
```
imeoer, line 240: Does this use as an internal arg?

imeoer (approved):

src/bin/nydus-image/main.rs
jiangliu: It's for backward compatibility.

jiangliu: And hope it can be phased out in 3.0

PR #1027, opened 2023-01-26
Move generic code from nydusd into nydus crate

jiangliu:
Move generic code from nydusd into nydus crate, so they can be reused by other projects.


src/lib.rs
```diff
@@ -2,38 +2,143 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+//<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
```
bergwolf, line 5: ??

bergwolf (commented):
@jiangliu Could you explain the purpose of the PR? Which component is going to consume the nydus crate and if the nydus crate is a library crate, why not move it to the top level as other library crates?

jiangliu:
> @jiangliu Could you explain the purpose of the PR? Which component is going to consume the nydus crate and if the nydus crate is a library crate, why not move it to the top level as other library crates?

The target audience is `image-rs` for confidential containers.
I reused `nydus` crate for simplicity. And it's Ok to split out as a dedicated library crate.

src/lib.rs
jiangliu: A leftover, fixed

imeoer (approved):
LGTM

PR #1031, opened 2023-01-29
nydus-test: don't fetch iostats per file metrics

changweige:
The files' metrics have a large size ending up with uncompleted json string. The root cause why the buffer is so uncomleted is still in fog. Moreover, we are going to remove files's metrics in the future.


jiangliu:
Aha, the dbs-uhttps has a limitation of message size, is it caused by that?

changweige:
> Aha, the dbs-uhttps has a limitation of message size, is it caused by that?

It's possible, however, due to de fact that metrics of all files' iostats might be removed in the future. Shall we just remove the test observations points?

jiangliu (approved):

PR #1038, opened 2023-01-31
github: recover unit test for macos

jiangliu:
The unit test for macos got lost during refactor, so recover it.


imeoer (commented):
It seems to slow down the action, we'd better put it in the daily test.

adamqqqplay:
> It seems to slow down the action, we'd better put it in the daily test.

I also agree, it's too slow right now and might delay the PR merge.

PR #1051, opened 2023-02-03
storage: fix bugs in prefetch data for ZRan images

jiangliu:
ZRan images may have holes between compressed chunk data, which breaks the prefetch algorithm. So fix the bug by special handling of prefetch for ZRan images.


changweige:
need rebase to master since anolis test script is updated

PR #1052, opened 2023-02-03
Fix anolis test

changweige:
Update smoke tests on anolis and docs about how to start nydus-snapshotter service.
In addition, simplify how to start nydus-snapshotter by removing cli flags with the same value as default value

### DIFF
--- a/docs/nydus-fscache.md
+++ b/docs/nydus-fscache.md
@@ -89,13 +89,9 @@ make
 mkdir -p /var/lib/containerd-nydus
 
 ./bin/containerd-nydus-grpc \
- --config-path /etc/nydus/nydusd-config.fscache.json \
- --daemon-mode shared \
+ --nydusd-config /etc/nydus/nydusd-config.fscache.json \
  --fs-driver fscache \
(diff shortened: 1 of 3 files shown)


docs/nydus-fscache.md
```diff
@@ -89,13 +89,10 @@ make
 mkdir -p /var/lib/containerd-nydus
 
 ./bin/containerd-nydus-grpc \
- --config-path /etc/nydus/nydusd-config.fscache.json \
+ --nydusd-config /etc/nydus/nydusd-config.fscache.json \
  --daemon-mode shared \
```
imeoer, line 93: Is it necessary?

imeoer (commented):
others LGTM!

docs/nydus-fscache.md
changweige: You are right, we can also remove the flag as `--fs-driver fscache` enforces it.

changweige: fixed

imeoer (approved):

PR #1062, opened 2023-02-08
zran: fix an assert when generating ZRAN images

jiangliu:
Fix an assert when generating ZRAN images.


utils/src/compress/zlib_random.rs
```diff
@@ -422,14 +421,21 @@ impl<R: Read> Read for ZranReaderState<R> {
         loop {
             // Reload the input buffer when needed.
             if self.stream.avail_in() == 0 {
+                error!("last byte ptr {:x}", self.stream.stream.next_in as usize);
```
imeoer, line 424: Is it debug message?

jiangliu: sorry, fixed

imeoer (approved):

PR #1067, opened 2023-02-09
nydusify: fix overlayfs mount options for check

imeoer:
To fix the error when mount OCI v1 image in check subcommand:

```
error: mount options is too long
```

The mount options has 4k buffer size limitation, we will
encounter the issue in a huge image with mass layers.

We need to shorten the length of lowerdir paths in overlayfs
option, change `sha256:xxx` to `layer-N` to alleviate the issue.

### DIFF
--- a/contrib/nydusify/pkg/checker/rule/filesystem.go
+++ b/contrib/nydusify/pkg/checker/rule/filesystem.go
@@ -330,6 +330,19 @@ func (rule *FilesystemRule) Validate() error {
 		return nil
 	}
 
+	// Cleanup temporary directories
+	defer func() {
+		if err := os.RemoveAll(rule.SourcePath); err != nil {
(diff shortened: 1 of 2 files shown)


hsiangkao:
You could also use new mount api called fsopen to mount, which doesn't have 4k limitation.

imeoer:
> You could also use new mount api called fsopen to mount, which doesn't have 4k limitation.

It looks great! we used the mount method from containerd package, maybe we can improve that for it.

adamqqqplay:
@imeoer Will this change cause compatibility issues? Or is this just an internal change in nydusify.

imeoer:
@adamqqqplay No, it's an internal change for `nydusify check` subcommand.

adam3q (approved):
LGTM

jiangliu (approved):

PR #1072, opened 2023-02-11
nydus-service: split service framework from nydus into nydus-service

jiangliu:
Split service framework from nydus into nydus-service crate, to reduce dependency when used by other project like image-rs.


imeoer (approved):

liubogithub (approved):

liubogithub:
Looks good.

PR #1073, opened 2023-02-11
nydusify: speed up conversion with tar2rafs

imeoer:
Default to enable `--type tar-rafs` option for nydus-image, which
converting OCI tar blob stream into nydus blob directly, eliminating
the need to decompress it to a local directory first, thus greatly
accelerating the pack process.

For compatibility, we still support env `NYDUS_DISABLE_TAR2RAFS=true`
variable to disable the optimization.

It's an internal feature, so we just only bump acceleration-service
and nydus-snapshotter packages to the latest version.

Related PR: https://github.com/containerd/nydus-snapshotter/pull/352


imeoer:
Maybe try `--type tar-rafs` and `--features blob-toc` first, then fallback if the builder is not supported.

jiangliu (approved):

PR #1082, opened 2023-02-15
nydus-image: add `backend-config` support to `nydus-image unpack`

loheagn:
The `nydus-image unpack` only support reading nydus blob from the blob files on local machine (by the argument `blob`) by now. This patch adds a new argument `backend-config` to this command to allow `nydus-image unpack` to read blob data from all kinds of backends.

### DIFF
--- a/src/bin/nydus-image/main.rs
+++ b/src/bin/nydus-image/main.rs
@@ -24,9 +24,11 @@ use clap::parser::ValueSource;
 use clap::{Arg, ArgAction, ArgMatches, Command as App};
 use nix::unistd::{getegid, geteuid};
 use nydus::get_build_time_info;
-use nydus_api::{BuildTimeInfo, ConfigV2};
+use nydus_api::{BuildTimeInfo, ConfigV2, LocalFsConfig};
 use nydus_app::setup_logging;
 use nydus_rafs::metadata::{RafsSuper, RafsSuperConfig, RafsVersion};
+use nydus_storage::backend::localfs::LocalFs;
+use nydus_storage::backend::BlobBackend;
 use nydus_storage::device::BlobFeatures;
 use nydus_storage::factory::BlobFactory;
 use nydus_storage::meta::format_blob_features;
@@ -526,6 +528,12 @@ fn prepare_cmd_args(bti_string: &'static str) -> App {
                         .required_unless_present("bootstrap"),
                 )
                 .arg(
+                    Arg::new("backend-config")
+                        .long("backend-config")
+                        .help("config file of backend")
+                        .required(false),
+                )
+                .arg(
                     Arg::new("bootstrap")
                         .short('B')
                         .long("bootstrap")
@@ -955,10 +963,7 @@ impl Command {
             Some(args) => Some(import_chunk_dict(args, config, &rs.meta.get_config())?),
         };
 
-        let cfg_file = matches.get_one::<String>("backend-config").unwrap();
-        let cfg = ConfigV2::from_file(cfg_file)?;
-        let backend_cfg = cfg.get_backend_config()?;
-        let backend = BlobFactory::new_backend(backend_cfg, "compactor")?;
+        let backend = Self::get_backend(matches, "compactor")?;
 
         let config_file_path = matches.get_one::<String>("config").unwrap();
         let file = File::open(config_file_path)
@@ -984,9 +989,25 @@ impl Command {
         if output.is_empty() {
             return Err(anyhow!("invalid empty --output option"));
         }
-        let blob = matches.get_one::<String>("blob").map(|s| s.as_str());
 
-        OCIUnpacker::new(bootstrap, blob, output)
+        let blob = matches.get_one::<String>("blob").map(|s| s.as_str());
+        let backend: Option<Arc<dyn BlobBackend + Send + Sync>> = match blob {
+            Some(blob_path) => {
+                let blob_path = PathBuf::from(blob_path);
+                let local_fs_conf = LocalFsConfig {
+                    blob_file: blob_path.to_str().unwrap().to_owned(),
+                    dir: Default::default(),
+                    alt_dirs: Default::default(),
+                };
+                let local_fs = LocalFs::new(&local_fs_conf, Some("unpacker"))
+                    .with_context(|| format!("fail to create local backend for {:?}", blob_path))?;
+
+                Some(Arc::new(local_fs))
+            }
+            None => Self::get_backend(matches, "unpacker").ok(),
+        };
+
+        OCIUnpacker::new(bootstrap, backend, output)
             .with_context(|| "fail to create unpacker")?
             .unpack(config)
             .with_context(|| "fail to unpack")
@@ -1208,6 +1229,20 @@ impl Command {
         }
 
         Ok(Arc::new(config))
+    }
+
+    fn get_backend(
+        matches: &ArgMatches,
+        blob_id: &str,
+    ) -> Result<Arc<dyn BlobBackend + Send + Sync>> {
+        let cfg_file = matches
+            .get_one::<String>("backend-config")
+            .context("missing backend-config argument")?;
+        let cfg = ConfigV2::from_file(cfg_file)?;
+        let backend_cfg = cfg.get_backend_config()?;
+        let backend = BlobFactory::new_backend(backend_cfg, blob_id)?;
+
+        Ok(backend)
     }
 
     fn get_blob_id(matches: &ArgMatches) -> Result<String> {
--- a/src/bin/nydus-image/unpack/mod.rs
+++ b/src/bin/nydus-image/unpack/mod.rs
@@ -10,15 +10,13 @@ use std::str;
 use std::sync::Arc;
 
 use anyhow::{Context, Result};
-use nydus_api::{ConfigV2, LocalFsConfig};
+use nydus_api::ConfigV2;
 use nydus_rafs::{
     metadata::{RafsInodeExt, RafsSuper},
     RafsIterator,
 };
-use nydus_storage::{
-    backend::{localfs::LocalFs, BlobBackend, BlobReader},
-    device::BlobInfo,
-};
+use nydus_storage::backend::BlobBackend;
+use nydus_storage::device::BlobInfo;
 use tar::{Builder, Header};
 
 use self::pax::{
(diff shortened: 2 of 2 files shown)


jiangliu:
How about using "-C, --config <config>" instead? We are going to switch from "--backend-config" to "--config" and "--backend-config" will get deprecated.

loheagn:
> How about using "-C, --config " instead? We are going to switch from "--backend-config" to "--config" and "--backend-config" will get deprecated.

@jiangliu But I think this will be conflict with the `backend` in the `Config` when load rafs. https://github.com/dragonflyoss/image-service/blob/master/rafs/src/metadata/mod.rs#L685

jiangliu:
> > How about using "-C, --config " instead? We are going to switch from "--backend-config" to "--config" and "--backend-config" will get deprecated.
> 
> @jiangliu But I think this will be conflict with the `backend` in the `Config` when load rafs. https://github.com/dragonflyoss/image-service/blob/master/rafs/src/metadata/mod.rs#L685

Are you going to support this usage scenario?
1) nydus meta blob has already been extracted to a local directory
2) nydus data blobs haven't been downloaded to the local directory

loheagn:
> > > How about using "-C, --config " instead? We are going to switch from "--backend-config" to "--config" and "--backend-config" will get deprecated.
> > 
> > 
> > @jiangliu But I think this will be conflict with the `backend` in the `Config` when load rafs. https://github.com/dragonflyoss/image-service/blob/master/rafs/src/metadata/mod.rs#L685
> 
> Are you going to support this usage scenario?
> 
> 1. nydus meta blob has already been extracted to a local directory
> 2. nydus data blobs haven't been downloaded to the local directory

Yes. 

For example, by now, when we use the `coverter.Unpack` method in nydus-snapshotter to unpack a nydus image layer, we need to extract both the bootstrap part and the blob data part to the local disk first to allow the nydus-image to read to blob data (use the localfs backend). But with this patch, we can just need to extract the bootstrap part and set up a local http server to serve the blob data part. Then the `backend-config` can be used to pass the information about this local http server.

jiangliu:
> > > > How about using "-C, --config " instead? We are going to switch from "--backend-config" to "--config" and "--backend-config" will get deprecated.
> > > 
> > > 
> > > @jiangliu But I think this will be conflict with the `backend` in the `Config` when load rafs. https://github.com/dragonflyoss/image-service/blob/master/rafs/src/metadata/mod.rs#L685
> > 
> > 
> > Are you going to support this usage scenario?
> > 
> > 1. nydus meta blob has already been extracted to a local directory
> > 2. nydus data blobs haven't been downloaded to the local directory
> 
> Yes.
> 
> For example, by now, when we use the `coverter.Unpack` method in nydus-snapshotter to unpack a nydus image layer, we need to extract both the bootstrap part and the blob data part to the local disk first to allow the nydus-image to read to blob data (use the localfs backend). But with this patch, we can just need to extract the bootstrap part and set up a local http server to serve the blob data part. Then the `backend-config` can be used to pass the information about this local http server.

Can we let nydusd to handle the bootstrap part too?

loheagn:
> > > > > How about using "-C, --config " instead? We are going to switch from "--backend-config" to "--config" and "--backend-config" will get deprecated.
> > > > 
> > > > 
> > > > @jiangliu But I think this will be conflict with the `backend` in the `Config` when load rafs. https://github.com/dragonflyoss/image-service/blob/master/rafs/src/metadata/mod.rs#L685
> > > 
> > > 
> > > Are you going to support this usage scenario?
> > > 
> > > 1. nydus meta blob has already been extracted to a local directory
> > > 2. nydus data blobs haven't been downloaded to the local directory
> > 
> > 
> > Yes.
> > For example, by now, when we use the `coverter.Unpack` method in nydus-snapshotter to unpack a nydus image layer, we need to extract both the bootstrap part and the blob data part to the local disk first to allow the nydus-image to read to blob data (use the localfs backend). But with this patch, we can just need to extract the bootstrap part and set up a local http server to serve the blob data part. Then the `backend-config` can be used to pass the information about this local http server.
> 
> Can we let nydus-image to handle the bootstrap part too? That's a feature I expect, then nydus-image can process remote images.



> > > > > How about using "-C, --config " instead? We are going to switch from "--backend-config" to "--config" and "--backend-config" will get deprecated.
> > > > 
> > > > 
> > > > @jiangliu But I think this will be conflict with the `backend` in the `Config` when load rafs. https://github.com/dragonflyoss/image-service/blob/master/rafs/src/metadata/mod.rs#L685
> > > 
> > > 
> > > Are you going to support this usage scenario?
> > > 
> > > 1. nydus meta blob has already been extracted to a local directory
> > > 2. nydus data blobs haven't been downloaded to the local directory
> > 
> > 
> > Yes.
> > For example, by now, when we use the `coverter.Unpack` method in nydus-snapshotter to unpack a nydus image layer, we need to extract both the bootstrap part and the blob data part to the local disk first to allow the nydus-image to read to blob data (use the localfs backend). But with this patch, we can just need to extract the bootstrap part and set up a local http server to serve the blob data part. Then the `backend-config` can be used to pass the information about this local http server.
> 
> Can we let nydus-image to handle the bootstrap part too? That's a feature I expect, then nydus-image can process remote images.

@jiangliu Yes, I think we can. 

We need to implement the logic of extracting the bootstrap and blob data from the whole nydus image layer (which is implemented in nydus-snapshotter by now) and let the caller tell nydus-image the loacation of the image layer by some argument (may be `--image-path`).

What's your option? @imeoer

jiangliu:
> > > > > > How about using "-C, --config " instead? We are going to switch from "--backend-config" to "--config" and "--backend-config" will get deprecated.
> > > > > 
> > > > > 
> > > > > @jiangliu But I think this will be conflict with the `backend` in the `Config` when load rafs. https://github.com/dragonflyoss/image-service/blob/master/rafs/src/metadata/mod.rs#L685
> > > > 
> > > > 
> > > > Are you going to support this usage scenario?
> > > > 
> > > > 1. nydus meta blob has already been extracted to a local directory
> > > > 2. nydus data blobs haven't been downloaded to the local directory
> > > 
> > > 
> > > Yes.
> > > For example, by now, when we use the `coverter.Unpack` method in nydus-snapshotter to unpack a nydus image layer, we need to extract both the bootstrap part and the blob data part to the local disk first to allow the nydus-image to read to blob data (use the localfs backend). But with this patch, we can just need to extract the bootstrap part and set up a local http server to serve the blob data part. Then the `backend-config` can be used to pass the information about this local http server.
> > 
> > 
> > Can we let nydus-image to handle the bootstrap part too? That's a feature I expect, then nydus-image can process remote images.
> 
> > > > > > How about using "-C, --config " instead? We are going to switch from "--backend-config" to "--config" and "--backend-config" will get deprecated.
> > > > > 
> > > > > 
> > > > > @jiangliu But I think this will be conflict with the `backend` in the `Config` when load rafs. https://github.com/dragonflyoss/image-service/blob/master/rafs/src/metadata/mod.rs#L685
> > > > 
> > > > 
> > > > Are you going to support this usage scenario?
> > > > 
> > > > 1. nydus meta blob has already been extracted to a local directory
> > > > 2. nydus data blobs haven't been downloaded to the local directory
> > > 
> > > 
> > > Yes.
> > > For example, by now, when we use the `coverter.Unpack` method in nydus-snapshotter to unpack a nydus image layer, we need to extract both the bootstrap part and the blob data part to the local disk first to allow the nydus-image to read to blob data (use the localfs backend). But with this patch, we can just need to extract the bootstrap part and set up a local http server to serve the blob data part. Then the `backend-config` can be used to pass the information about this local http server.
> > 
> > 
> > Can we let nydus-image to handle the bootstrap part too? That's a feature I expect, then nydus-image can process remote images.
> 
> @jiangliu Yes, I think we can.
> 
> We need to implement the logic of extracting the bootstrap and blob data from the whole nydus image layer (which is implemented in nydus-snapshotter by now) and let the caller tell nydus-image the loacation of the image layer by some argument (may be `--image-path`).
> 
> What's your option? @imeoer

We may discuss offline. I hope we can enhance nydus-image to directly handle remote image, the same to nydusd.

jiangliu:
Does this work for you?
```
nydus-image unpack -C nydusd.conf --outputb image.tar.gz path/to/bootstrap.file
```

api/src/config.rs
```diff
@@ -250,6 +250,7 @@ pub struct BackendConfigV2 {
     /// Configuration for container registry backend.
     pub registry: Option<RegistryConfig>,
     /// Configuration for local http proxy.
+    #[serde(rename = "http-proxy")]
     pub http_proxy: Option<HttpProxyConfig>,
```
jiangliu, line 254: This deserves a dedicated PR:)

src/bin/nydus-image/main.rs
```diff
@@ -17,16 +17,19 @@ extern crate lazy_static;
 use std::fs::{self, metadata, DirEntry, File, OpenOptions};
 use std::os::unix::fs::FileTypeExt;
 use std::path::{Path, PathBuf};
+use std::result::Result::Ok;
```
jiangliu, line 20: This may be unnecessary.

loheagn: Yes. Fixed :)

api/src/config.rs
loheagn: Here's the pr https://github.com/dragonflyoss/image-service/pull/1086 .

loheagn:
> Does this work for you?
> 
> ```
> nydus-image unpack -C nydusd.conf --outputb image.tar.gz path/to/bootstrap.file
> ```

NO. The argument `--bootstrap` is required.

src/bin/nydus-image/main.rs
```diff
@@ -1210,6 +1231,20 @@ impl Command {
...
+            .context("missing backend-config argument")?;
+        let cfg = ConfigV2::from_file(cfg_file)?;
+        let backend_cfg = cfg.get_backend_config()?;
+        let backend = BlobFactory::new_backend(backend_cfg, blob_id)?;
+
+        Ok(backend)
+    }
+
     fn get_blob_id(matches: &ArgMatches) -> Result<String> {
         let mut blob_id = String::new();
```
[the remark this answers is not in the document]
loheagn: You are so smart 💯

jiangliu:
> > Does this work for you?
> > ```
> > nydus-image unpack -C nydusd.conf --outputb image.tar.gz path/to/bootstrap.file
> > ```
> 
> NO. The argument `--bootstrap` is required.

With nydus 2.2, the "--bootstrap" option for `nydus-image unpack` should be optional:)
I simplified it by making "--bootstrap" optional.

imeoer (approved):

PR #1104, opened 2023-02-23
rafs: enforce stricter and safer validation of RAFS v6 images

jiangliu:
Enforce stricter and safer validations of RAFS v6 images,
- return error instead of panic
- ensure Dirent content doesn't cross EROFS_BLOCK_SIZE boundary
- ensure symlink size is smaller than EROFS_BLOCK_SIZE
- ensure xattr name and value are within range
- validate block index for flat plain and flat inline inodes
- add helpers to reduce duplicated code


mofishzz (approved):
LGTM!

imeoer (approved):

PR #1108, opened 2023-02-24
Correctly generate mapped-blkaddr for RAFS devslot array

jiangliu:
Correctly generate `mapped-blkaddr` for RAFS devslot array, it helps to expose a RAFS v6 image as a block device.


mofishzz (approved):

mofishzz (approved):
LGTM

PR #1116, opened 2023-02-28
nydusify: enable pigz by default

imeoer:
```
We should use pigz for supporting parallel gzip decompression, so that
improve the conversion speed when unpack gzip layer for source image.

We still allow users to specify the env `CONTAINERD_DISABLE_PIGZ=1` to
disable the feature when encounter any decompression error.

See https://github.com/containerd/containerd/blob/33c0eafb17dff6de43f4a530fcf05b02ed24330b/archive/compression/compression.go#L261
```

### DIFF
--- a/contrib/nydusify/cmd/nydusify.go
+++ b/contrib/nydusify/cmd/nydusify.go
@@ -863,11 +863,6 @@ func main() {
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


jiangliu (approved):

changweige:
Should we cherry-pick this to stable 2.1?

imeoer:
@changweige It's ok, will do that.

PR #1134, opened 2023-03-03
smoke: Add case for localfsDir option

ccx1024cc:
Test case for an stale option `localfsDir`.

### DIFF
--- a/smoke/tests/native_layer_test.go
+++ b/smoke/tests/native_layer_test.go
@@ -5,6 +5,7 @@
 package tests
 
 import (
+	"os"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


smoke/tests/native_layer_test.go
```diff
@@ -156,6 +157,65 @@ func (n *NativeLayerTestSuite) testMakeLayers(ctx tool.Context, t *testing.T) {
 	tool.Verify(t, ctx, lowerLayer.FileTree)
 }
 
+// specify blob dir and cache dir with option `localfs-dir`
+func (n *NativeLayerTestSuite) TestSharedCacheDirWithBlobDir(t *testing.T) {
```
imeoer, line 161: Why do we need to add an independent case for `local-fs-dir`, can we reuse `TestMakeLayers` ?

ccx1024cc: > Why do we need to add an independent case for `local-fs-dir`, can we reuse `TestMakeLayers` ?

Because the workflow is not same, `testMakeLayers` does not support this option.

imeoer: Can we make it reusable?

PR #1137, opened 2023-03-03
builder: support `--parent-bootstrap` for merge

imeoer:
This option allows merging multiple bootstraps of upper layer with
the bootstrap of a parent image, so that we can implement container
commit operation for nydus image.

Related PR: https://github.com/containerd/nydus-snapshotter/pull/403


src/bin/nydus-image/merge.rs
```diff
@@ -191,15 +211,8 @@ impl Merger {
...
-                let mut found = false;
-                for (idx, blob) in blob_mgr.get_blobs().iter().enumerate() {
-                    if blob.blob_id == blob_ctx.blob_id {
-                        blob_idx_map.push(idx as u32);
-                        found = true;
-                    }
-                }
-                if !found {
-                    blob_idx_map.push(blob_mgr.len() as u32);
+                if blob_idx_map.get(&blob.blob_id()).is_none() {
```
jiangliu, line 214: `!blob_idx_map.contains(&blob.blob_id())`?

rafs/src/metadata/mod.rs
```diff
@@ -699,21 +699,11 @@ impl RafsSuper {
             // Old converters extracts bootstraps from data blobs with inlined bootstrap
             // use blob digest as the bootstrap file name. The last blob in the blob table from
             // the bootstrap has wrong blod id, so we need to fix it.
-            let mut fixed = false;
             let blobs = rs.superblock.get_blob_infos();
             for blob in blobs.iter() {
                 // Fix blob id for new images with old converters.
                 if blob.has_feature(BlobFeatures::INLINED_FS_META) {
                     blob.set_blob_id_from_meta_path(path.as_ref())?;
-                    fixed = true;
```
jiangliu, line 708: Why removing this backward compatible handling?
If it's not needed any, please help to remove about comments:)

imeoer: We still need to keep `Fix blob id for new images with old converters.`, but remove `Fix blob id for old images with old converters.`.

In fact, there is no way to check if a separate old bootstrap file was inline to the blob, for example, for an merged bootstrap generated by old converter, we can't set the blob id it references to as the filename, otherwise it will break blob table on loading rafs.

The bug can be reproduced by these steps:

```
nydus-image (v1.0.0) create --blob /blob --bootstrap /bootstrap /source

nydus-image (v2.2.0) check --bootstrap /bootstrap

RAFS filesystem metadata is valid, referenced data blobs:
	 0: old-v5-bootstrap, compressed data size 0x0, compressed file size 0x0, uncompressed file size 0x0, chunks: 0x0, features:
	 1: old-v5-bootstrap, compressed data size 0x0, compressed file size 0x0, uncompressed file size 0x0, chunks: 0x0, features:
	 2: old-v5-bootstrap, compressed data size 0x0, compressed file size 0x0, uncompressed file size 0x0, chunks: 0x0, features:
	 3: old-v5-bootstrap, compressed data size 0x0, compressed file size 0x0, uncompressed file size 0x0, chunks: 0x0, features:
```

Ths blob ids in output are broken.

jiangliu (approved):

PR #1141, opened 2023-03-06
Move RAFS filesystem builder into nydus-rafs crate

jiangliu:
Move RAFS filesystem builder into nydus-rafs crate so it can be reused. Also reduce memory consumption during image conversion.


bergwolf:
Which component is supposed to reuse the builder?

jiangliu:
> Which component is supposed to reuse the builder?

Image-rs for confidential containers：）

imeoer (approved):

PR #1147, opened 2023-03-13
Introduce new tarfs mode to Nydus

jiangliu:
When using containerd overlayfs snapshotter to handle OCIv1 images, it works as below:
- download compressed blobs from registry
- uncompress compressed blobs into tar streams/files
- unpack tar streams to directories on local filesystem
- mount multiple directories into a container rootfs by overlayfs
    
Here we introduce a new work mode to nydus, called as TARFS, which works as below:
- download compressed blobs from registry
- uncompress compressed blobs into tar streams/files
- build RAFS/EROFS meta blob from tar streams/files
- optionally merge multiple RAFS/EROFS meta blobs into one
- mount the generated RAFS filesystem by mount.erofs
    
By introducing TARFS mode to RAFS, it helps to avoid generating a bunch of small files by `untar`, which speeds up image preparation and garbage collection. It may also help to reduce levels of overlayfs by merging multiple image layers into one final RAFS filesystem.
    
The TARFS mode of RAFS filesystem has several special behavior, compared to current RAFS, as below:
- Instead of generating RAFS data blob, it directly use tar files as RAFS data blob.
- Tar files are uncompressed, so data blobs for TARFS mode are uncompressed.
- Tar files will also be directly used as local cache files.
- There's no chunk compression info, chunk digest, TOC etc, generated for TARFS mode.
- Block size is 512 bytes instead of 4K, because tar files are 512 bytes aligned.
    
 Now we have three ways to make of OCIv1 images:
| Mode     |    TAR-TARFS   | TARGZ-REF  | TARGZ-RAFS |
| -------- | -------| --------|--------| 
| Generate meta blob? |            Y |                      Y      |                 Y |
|    Generate chunk data?       |     N           |            N            |           Y |
|    Generate blob.meta?        |     N            |           Y              |         Y |
|    Generate data blobs?       |     N            |           Y(for blob.meta)  |      Y
|    Data in data blobs?           |  Not generated   |        blob.meta       |        chunk data & blob.meta |
|    Chunk alignment?             |   512            |         4096            |        4096 |
|    Chunk dedup?                   | N                 |      Y                      |  Y |
|    Lazy loading?                    |  N                |       Y                     |  Y |
    
    Note: RAFS in TARFS mode is designed to be used locally only. In other words, it's a way to implement snapshotter, instead of an image format for sharing.

### DIFF
--- a/rafs/src/builder/core/blob.rs
+++ b/rafs/src/builder/core/blob.rs
@@ -55,6 +55,7 @@ impl Blob {
                 Self::finalize_blob_data(ctx, blob_mgr, blob_writer)?;
             }
             ConversionType::TarToRef
+            | ConversionType::TarToTarfs
             | ConversionType::TargzToRef
             | ConversionType::EStargzToRef => {
                 // Use `sha256(tarball)` as `blob_id` for ref-type conversions.
@@ -68,6 +69,9 @@ impl Blob {
                         }
                     } else if let Some(tar_reader) = &ctx.blob_tar_reader {
                         blob_ctx.compressed_blob_size = tar_reader.position();
+                        if ctx.conversion_type == ConversionType::TarToTarfs {
+                            blob_ctx.uncompressed_blob_size = blob_ctx.compressed_blob_size;
+                        }
                         if blob_ctx.blob_id.is_empty() {
                             let hash = tar_reader.get_hash_object();
                             blob_ctx.blob_id = format!("{:x}", hash.finalize());
--- a/rafs/src/builder/core/bootstrap.rs
+++ b/rafs/src/builder/core/bootstrap.rs
@@ -152,6 +152,7 @@ impl Bootstrap {
         let index = nodes.len() as u32 + 1;
         let parent = &mut nodes[tree.node.index as usize - 1];
         let parent_ino = parent.inode.ino();
+        let block_size = ctx.v6_block_size();
 
         // Maybe the parent is not a directory in multi-layers build scenario, so we check here.
         if parent.is_dir() {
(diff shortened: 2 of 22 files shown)


rafs/src/builder/core/blob.rs
```diff
@@ -62,12 +62,14 @@ impl Blob {
...
                         let reader = zran.lock().unwrap().reader();
                         blob_ctx.compressed_blob_size = reader.get_data_size();
+                        blob_ctx.current_compressed_offset = blob_ctx.compressed_blob_size;
                         if blob_ctx.blob_id.is_empty() {
                             let hash = reader.get_data_digest();
                             blob_ctx.blob_id = format!("{:x}", hash.finalize());
                         }
                     } else if let Some(tar_reader) = &ctx.blob_tar_reader {
                         blob_ctx.compressed_blob_size = tar_reader.position();
+                        blob_ctx.current_compressed_offset = blob_ctx.compressed_blob_size;
```
bergwolf, line 72: This needs to be backported. Could you send it via a seperate PR? /cc @imeoer

bergwolf (commented):
It's not clear how the PR is supposed to work. Did you include **ALL** of your modification?

rafs/src/builder/core/blob.rs
jiangliu: I was going to split it out as a dedicated PR, but found it's a false alarm:(
The assertion should be
```
assert!(self.current_compressed_offset >= self.compressed_blob_size);
```
instead of
```
assert!(self.current_compressed_offset > self.compressed_blob_size);
```

So the fixup is unnecessary:)

jiangliu:
With the EROFS patchset from https://kernel.source.codeaurora.cn/pub/scm/linux/kernel/git/xiang/erofs.git/?h=dev-test, you can test the PR as below:
- cargo build --features block-nbd
- modprobe nbd
- mkdir images
- mkdir mnt
- tar cvf src.tar src
- sha256sum src.tar
- cp src.tar images/d070105c2e4773da6bc94ba593ac15112cc613f57a14bac0233aa51ba3220087
- target/debug/nydus-image create -t tar-tarfs -D images/ src.tar 
- target/debug/nydusd nbd -D images/ -B images/6e40836c5536df7c7bdad85136135fcd2ca38816fe33ae254640fcad30a454f2 /dev/nbd0 &
- mount -t erofs /dev/nbd0 ./mnt/
- cat mnt/src/lib.rs

imeoer (approved):

PR #1151, opened 2023-03-16
Enhance `nydus-image merge` to support tarfs

jiangliu:
Enhance `nydus-image merge` to support tarfs, so we can merge multiple RAFS filesystems in TARFS mode into a single one.


imeoer (approved):
LGTM!

PR #1152, opened 2023-03-16
rafs: reclaim used memory of tailer page in case of flat plain mode

jiangliu:
When the directory inode is in flat plain mode, there may be space left in the tailer page, so reclaim it.


hsiangkao (changes requested):
incorrect change, since for the plain mode, and whole block will be mapped even it's beyond i_size.

PR #1156, opened 2023-03-17
rafs: define dedicated RafsV6Inode to reduce memory consumption

jiangliu:
There are several unused fields in RafsV5Inode when used for v6, so define dedicated RafsV6Inode to reduce memory consumption.


imeoer (approved):
LGTM

PR #1157, opened 2023-03-17
rafs: do not include chunk info array in the metadata blob

jiangliu:
Now we can reconstruct chunk info from RAFS v6 data blob with embedded chunk digest, so no need to embed the chunk info array in the metadata blob anymore, which helps to dramatically reduce size of metadata blob for big images.


adamqqqplay:
cargo-deny can be fixed in PR #1240

adam3q (commented):
need rebase

PR #1168, opened 2023-03-21
rafs: fix incorrect blob id in merged TARFS

jiangliu:
When merging multiple RAFS filesystems in TARFS mode into one, the generated data blob is incorrectly, it actually the meta blob id instead of data blob id.


adam3q (approved):
LGTM

PR #1177, opened 2023-03-28
nydus: fix a possible panic caused by SubCmdArgs::is_present()

jiangliu:
Fix a possible panic caused by SubCmdArgs::is_present().


imeoer (approved):
LGTM

PR #1184, opened 2023-03-30
rafs: fix a incorrect mapped_blkaddr for multi layer images

jiangliu:
When generating a RAFS filesystem with multiple data blobs, the mapped_blkaddr for second and following-on blobs are incorrect.


imeoer (approved):

imeoer:
@jiangliu Does this need to be backported to v2.2?

PR #1186, opened 2023-03-30
storage: fix http fallback handle

imeoer:
If we attempt to establish a TLS connection with the HTTP registry server,
we are likely to encounter these types of error:

- Error `wrong version number` from openssl library;
- Error `connection refused` from standard library;

Before this, only the first type of error was handled. This commit handles
the second type of error, which was reproduced by running a local insecure
harbor registry.


adam3q (approved):
LGTM

jiangliu (approved):

PR #1187, opened 2023-03-31
storage: use global thread pool to support prefetch

jiangliu:
Currently we use a prefetch manager per backend, this design has two drawbacks:
- spawn too many worker threads for prefetch
- the network ratelimit has been defunct

So change the design to use a global prefetch manager with thread pool to supporrt all prefetch requests.


ccx1024cc:
Is it for process-internal enhancement? Furthermore, what is the typical case, in which multiple backends runs in the same process?

ccx1024cc (changes requested):
A test for improvement is required.

PR #1200, opened 2023-04-07
service: add coco feature in Cargo.toml

taoohong:
Add feature coco to Cargo.toml, so that confidential containers can apply this feature to use nydus to download images.


jiangliu:
We have fixed a go version, which fixes the smoke test failure. Could you please help to force push again?

imeoer (approved):

jiangliu (approved):

PR #1204, opened 2023-04-10
nydus-image: support generate tarfs bootstrap from tar stream via stdin

kevinXYin:
Add '--blob-stdin-tar-stream' option to 'nydus-image create -t tar-tarfs' to generate tarfs bootstrap from tar stream via stdin instead of SOURCE tar file.

### DIFF
--- a/rafs/src/builder/compact.rs
+++ b/rafs/src/builder/compact.rs
@@ -578,6 +578,7 @@ impl BlobCompactor {
             Default::default(),
             None,
             false,
+            false,
             Features::new(),
         );
         let mut bootstrap_mgr =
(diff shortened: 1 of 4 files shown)


jiangliu:
How about using FIFO instead of stdin? Once nydus-image reads prefetch list from stdin, so it would be better to use FIFO instead.

imeoer:
> How about using FIFO instead of stdin? Once nydus-image reads prefetch list from stdin, so it would be better to use FIFO instead.

+1

kevinXYin:
> How about using FIFO instead of stdin? Once nydus-image reads prefetch list from stdin, so it would be better to use FIFO instead.

Thanks , I will try FIFO. And check if there is performance drop using FIFO compared to stdin, cause may need additional memory copies and interaction. I think we need to make image convert as fast as possible for tarfs usage.

ccx1024cc (changes requested):
Please add e2e test cases for this new option, then update doc. thx

PR #1202, opened 2023-04-07
nydus-image: support small chunks mergence

hangvane:
Add the support of merging small file chunks into one batch chunk, for `ConversionType::DirectoryToRafs`, `ConversionType::EStargzToRafs`, `ConversionType::TargzToRafs`, and `ConversionType::TarToRafs`.

# basic usage

Add the `--batch-size` arg to command to enable chunk mergence for supported conversion types:

```shell
nydus-image create --bootstrap ~/bootstrap --blob-dir ~/blobs ~/source --batch-size 0x100000

nydus-image create --type estargz-rafs --bootstrap ~/bootstrap --blob-dir ~/blobs ~/source.tar.gz --batch-size 0x100000

nydus-image create --type targz-rafs --bootstrap ~/bootstrap --blob-dir ~/blobs ~/source.tar.gz --batch-size 0x100000

nydus-image create --type tar-rafs --bootstrap ~/bootstrap --blob-dir ~/blobs ~/source.tar --batch-size 0x100000
```
# benchmarks

- 4K-aligned: append zeros to chunk data buffer so that the data to be compressed is 4K aligned.
- continuous: a big chunk between small chunks will cut the batch chunk.
- 1/2 filtered: do not merge chunks with their uncompressed size bigger than 1/2 batch size.

| repositories | tags | 4K-aligned | continuous | 1/2 filtered | size (MiB) | % | 
|--|--|:--:|:--:|:--:|--:|--:|
| wordpress | latest | - | - | - | 207.36 | - |
| wordpress | latest-rafs | - | - | - | 211.17 | +1.84% |
| wordpress | latest-rafs-batch-64K | ✖ | ✖ | ✖ | 205.63 | -0.83% |
| wordpress | latest-rafs-batch-64K | ✖ | ✖ | ✔ | 205.62 | -0.84% |
| wordpress | latest-rafs-batch-64K | ✖ | ✔ | ✖ | 205.76 | -0.77% |
| wordpress | latest-rafs-batch-64K | ✔ | ✖ | ✖ | 206.17 | -0.57% |
| wordpress | latest-rafs-batch-64K | ✔ | ✖ | ✔ | 206.16 | -0.58% |
| wordpress | latest-rafs-batch-64K | ✖ | ✔ | ✔ | 205.85 | -0.72% |
| wordpress | latest-rafs-batch-64K | ✔ | ✔ | ✖ | 206.28 | -0.52% |
[...]

PR #1206, opened 2023-04-11
feat: support the arg batch-size for small chunks mergence for nydusify

Desiki-high:
Add the batch-size arg  for nydusify and the batch-size arg is new feature for nydus-image, which support merge small file chunks into one batch chunk.

That is the second part of work , the first part work is in [nydus-snapshotter](https://github.com/containerd/nydus-snapshotter/pull/445)

This PR is related to https://github.com/dragonflyoss/image-service/pull/1202.


ccx1024cc:
Please add smoke test cases for new option.

imeoer (approved):

Issue #1198, opened 2023-04-07
Fail to start v6 by stable/v2.1


ccx1024cc:
### Version
Based on latest stable/v2.1. More details as below.
```shell
# nydusify --version
Nydusify version
Version         : v2.1.5
Revision        : ef2033c2e21d36938410069f07d32a5095a19398
Go version      : go1.19.7
Build time      : 2023-04-06T18:45:32

# nydus-image --version
Version:        v2.1.5
Git Commit:     ef2033c2e21d36938410069f07d32a5095a19398
Build Time:     2023-04-06T10:29:47.360755167Z
Profile:        release
Rustc:          rustc 1.61.0 (fe5b13d68 2022-05-18)

# nydusd --version
Version:        v2.1.5
Git Commit:     ef2033c2e21d36938410069f07d32a5095a19398
Build Time:     2023-04-06T10:29:47.360755167Z
Profile:        release
Rustc:          rustc 1.61.0 (fe5b13d68 2022-05-18)
```

### Reproduction
```shell
# sudo nydusify convert --source wordpress:latest --target localhost:5077/nydus/wordpress:2.1-v6 --fs-version 6
# sudo nerdctl run -it  --rm --snapshotter nydus  --network host localhost:5077/nydus/wordpress:2.1-v6

# Panic by nydusd
[2023-04-07 11:08:34.004217 +08:00] ERROR [/home/morgan/.cargo/registry/src/github.com-1ecc6299db9ec823/log-panics-2.1.0/src/lib.rs:130] thread 'fuse_server' panicked at 'Chunk is not fully loaded': storage/src/device.rs:424
   0: backtrace::capture::Backtrace::create
   1: backtrace::capture::Backtrace::new

```

PR #1213, opened 2023-04-14
rafs: fix a regression caused by commit 2616fb2c0502e4

jiangliu:
Fix a regression caused by commit 2616fb2c0502e4.


hsiangkao (approved):

hsiangkao:
Please help verify if https://github.com/dragonflyoss/image-service/issues/1198 works now @ccx1024cc

PR #1219, opened 2023-04-15
feat: add the benchmark test for nydus

Desiki-high:
# To compare the container startup speed of nydus and oci, add the new feature.
## misc/benchmark/*.py is the benchmark workflow python script:
1. bench.py: bench the pull image time, create container time and start container time
2. benchmark.py : the benchmark satrt with four mode(oci, nydus without prefetch , nydus with prefetch all, nydus with prefetch file list)
3. convert.py: convet the image from oci to nydus by nydusify
4. metrics.py: use the nydus metrics api to collet the files which accessd by container in start time
5. prefetch_list.py: sort the file list by access time, maybe the more smart strategy will add in the future
6. util.py: the utility functions used in benchmark

## the other files in misc/benchmark is the config files or environment files:
1. cni_bridge.py: used by cni plugins
2. nydusd_config.json: nydusd config file
3. containerd_config.toml: snapshotter config file with containerd
4. nydus-snapshotter.service: systemd service to creat nydusd process

### we use wondershape to limit to local net bandwith(10M in [config.yml](https://github.com/dragonflyoss/image-service/compare/master...Desiki-high:benchmark?expand=1#diff-924ec53aea92a71d8b43a94cc3e5a1260b1be0cbc934fcd21fbb4fc2b65057be))
1. env.IMAGE: choose the image for benchmark(if you modify this please also modiy the config.yaml)
2. benchmark-oci: benchmark for oci
3. benchmark-nydus-no-prefetch: benchmark for nydsu image (without prefetch)
4. benchmark-nydus-all-prefetch: benchmark for nydsu image (prefetch /)
5. benchmark-nydus-filelist-prefetch: benchmark for nydsu image (prefetch the file list create by prefetch_list.py )
### upload the benchmark result in the artifacts

PR #1226, opened 2023-04-19
change the pull_request event to pull_request_target

Desiki-high:
In #1219 , [we can't comment on PR](https://github.com/dragonflyoss/image-service/actions/runs/4739197938/jobs/8413885944), so can we change the event from pull_request to pull_request_target to enable the action havs permission to comment on PR.

**But I'm not sure if this will cause any security issues?**


Desiki-high:
When we change the event in PR, the workfolw will not work. This is also why I create a new PR.

PR #1229, opened 2023-04-19
feat: add more types of file to smoke

ccx1024cc:
Including:
    * regular file with chinese name
    * regular with long name
    * symbolic link of deleted file
    * large regular file of 13MB
    * regular file with hole at both head and tail
    * empty regular file


jiangliu (approved):

sctb512:
Important bugfix. Can we backport these to v2.1.0?

Issue #1221, opened 2023-04-17
Proposal: Compare the current benchmark tests with the master branch's.


Desiki-high:
Reference: https://github.com/dragonflyoss/image-service/pull/1219#issuecomment-1510614518.

We need to compare the benchmark tests between the PR and the master branch's.

adamqqqplay:
@Desiki-high Hi, Thanks for your work, the benchmarks are now working. 

I have a small suggestion, can you let github-actions bot modify the previous comments instead of re-comments after each test. In this way, the comment page of the PR can be made clearer.

PR #1234, opened 2023-04-20
[WIP]action: replace the benchmark comment

Desiki-high:
We should modify the previous benchmark result comment instead of re-comment after each becnhmark test

Ref:https://github.com/dragonflyoss/image-service/issues/1221#issuecomment-1515834305


hsiangkao:
lack of zran-no-prefetch?

PR #1239, opened 2023-04-21
action: add zran-no-prefetch benchmark and make smoke.yml tidy

Desiki-high:
- Benchmark in smoke test has too many same steps, we can move them to shell and make [smoke.yml](https://github.com/dragonflyoss/image-service/tree/master/.github/workflows/smoke.yml) tidy by prepare_env.sh and benchmark_summary.sh.
- Add zran-no-prefetch benchmark.
- Change the benchmark-result order, set the oci in the first.
- Since https://github.com/dragonflyoss/image-service/pull/1238, we move the benchmark-result from PR comment to action summary, we should enable benchmark-result in push and schedule.
- Use the stable wordpress tag 6.11.
- Delete the artifacts after benchmark-result download all artifacts. Certainly, the nydus-integration-test had downloaded the artifacts.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -14,10 +14,8 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  SNAPSHOOTER_VERSION: 0.7.3
-  NERDCTL_VERSION: 1.3.0
-  CNI_PLUGINS_VERSION: 1.2.0
   IMAGE: wordpress
+  TAG: 6.1.1
 
 jobs:
   contrib-build:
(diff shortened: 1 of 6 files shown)


.github/workflows/smoke.yml
```diff
@@ -148,23 +146,12 @@ jobs:
...
-        sudo tar -xzvf cni-plugins-linux-amd64-v${{env.CNI_PLUGINS_VERSION}}.tgz -C /opt/cni/bin
-        sudo install -D misc/benchmark/cni_bridge.conf /etc/cni/net.d/bridge.conf
-    - name: Prepare OCI Registry Environment
-      run: |
-        sudo docker run -d --restart=always -p 5000:5000 --name registry registry
-        sudo docker pull ${{env.IMAGE}} && docker tag ${{env.IMAGE}}:latest localhost:5000/${{env.IMAGE}}:latest
-        sudo docker push localhost:5000/${{env.IMAGE}}:latest
-        git clone https://github.com/magnific0/wondershaper.git
-        sudo install -D -m 755 wondershaper/wondershaper /usr/local/bin
+        sudo install -D -m 755 misc/benchmark/prepare_env.sh prepare_env.sh
```
imeoer, line 151: No need to install the script.

.github/workflows/smoke.yml
```diff
@@ -193,32 +180,13 @@ jobs:
...
-        sudo install -D misc/benchmark/cni_bridge.conf /etc/cni/net.d/bridge.conf
-        sudo install -D misc/benchmark/nydusd_config.json /etc/nydus/config.json
-        sudo install -D misc/benchmark/containerd_config.toml /etc/containerd/config.toml
-        sudo systemctl restart containerd
-        sudo install -D misc/benchmark/nydus-snapshotter.service /etc/systemd/system/nydus-snapshotter.service
-        sudo systemctl start nydus-snapshotter
-    - name: Prepare Nydus Registry Environment
-      run: |
-        sudo docker run -d --restart=always -p 5000:5000 --name registry registry
+        sudo install -D -m 755 misc/benchmark/prepare_env.sh prepare_env.sh
```
imeoer, line 183: Ditto.

.github/workflows/smoke.yml
```diff
@@ -247,32 +215,51 @@ jobs:
...
+        sudo ./prepare_env.sh nydus
+        sudo docker pull ${{env.IMAGE}}:${{env.TAG}} && docker tag ${{env.IMAGE}}:${{env.TAG}} localhost:5000/${{env.IMAGE}}:${{env.TAG}}
+        sudo docker push localhost:5000/${{env.IMAGE}}:${{env.TAG}}
+        sudo DOCKER_CONFIG=$HOME/.docker nydusify convert \
+          --source localhost:5000/${{env.IMAGE}}:${{env.TAG}} \
+          --target localhost:5000/${{env.IMAGE}}:${{env.TAG}}_nydus \
+          --fs-version 6 \
+          --oci-ref \
+          --platform linux/amd64,linux/arm64
```
imeoer, line 227: It seems we just only run the test on `linux/amd64` platform, `linux/arm64` is unnecessary.

Desiki-high: > It seems we just only run the test on `linux/amd64` platform, `linux/arm64` is unnecessary.

Actually, we don't need to declaration the platform, so I delete the opt.

.github/workflows/smoke.yml
Desiki-high: fix

.github/workflows/smoke.yml
Desiki-high: fix

imeoer (approved):

PR #1243, opened 2023-04-22
action: add read-amount and read-count for benchmark

Desiki-high:
We should add the ```read-amount``` and ```read-count``` for nydus benchmark to compare nydus, zran with different batchsize, config and so on.
> For oci benchmark we use "-" for these fields.


Desiki-high:
cargo-deny is fixed by https://github.com/dragonflyoss/image-service/pull/1238.

imeoer (approved):
LGTM!

PR #1256, opened 2023-04-28
contrib: support nydus-overlayfs and ctr-remote on different platforms

### LINKED ISSUE #401: Add arm support for CI and release
We need to run CI and get the binary release for ARM platform.

mofishzz:
Otherwise, the binary we compiled cannot run on other platforms such as arm.

Fix: #401 #407


adam3q (approved):
LGTM, thanks!

imeoer (approved):

PR #1261, opened 2023-05-06
action: reuse rust caches

### LINKED ISSUE #1260: Reuse the rust caches
### And the caches is approaching total cache storage [limit](https://github.com/dragonflyoss/image-service/actions/caches). We have too many rust caches:
image
image
image
image

## So maybe we can reuse these caches by ```shared-key``` like https://github.com/Swatinem/rust-cache/issues/127?
1. https://github.com/dragonflyoss/image-service/blob/7a48992bce92aad5116907734e388719c3dafe5b/.github/workflows/benchmark.yml#L48
2. https://github.com/dragonflyoss/image-service/blob/7a48992bce92aad5116907734e388719c3dafe5b/.github/workflows/convert.yml#L51
3. https://github.com/dragonflyoss/image-service/blob/7a48992bce92aad5116907734e388719c3dafe5b/.github/workflows/smoke.yml#L88
4. https://github.com/dragonflyoss/image-service/blob/7a48992bce92aad5116907734e388719c3dafe5b/.github/workflows/smoke.yml#L112
5. https://github.com/dragonflyoss/image-service/blob/7a48992bce92aad5116907734e388719c3dafe5b/.github/workflows/smoke.yml#L603
6. https://github.com/dragonflyoss/image-service/blob/7a48992bce92aad5116907734e388719c3dafe5b/.github/workflows/smoke.yml#L619

Desiki-high:
### changes
1. update the Swatinem/rust-cache version
2. reuse the rust cache in ```nydus-build```, ```nydus-build-master``` and ``` nydus-unit-tes```.
### reason
1. ```nydus-build``` and ```nydus-build-master``` have same dep.
2. ```nydus-unit-test``` dependences have partial similarity with the ```nydus-build```, and it will use the cache create by ```nydus-build``` because of slower. So it will not block the smoke test.

close #1260.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -7,6 +7,7 @@ on:
   workflow_dispatch:
 
 env:
+  CARGO_TERM_COLOR: always
   IMAGE: wordpress
   TAG: 6.1.1
 
@@ -48,6 +49,7 @@ jobs:
       uses: Swatinem/rust-cache@v2.2.0
       with:
         cache-on-failure: true
+        shared-key: nydus-build
     - name: Build Nydus
       run: |
         rustup component add rustfmt clippy
--- a/.github/workflows/convert.yml
+++ b/.github/workflows/convert.yml
@@ -51,6 +51,7 @@ jobs:
       uses: Swatinem/rust-cache@v2.2.0
       with:
         cache-on-failure: true
+        shared-key: nydus-build
     - name: Build Nydus
       run: |
         rustup component add rustfmt clippy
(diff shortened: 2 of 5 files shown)


.github/workflows/benchmark.yml
```diff
@@ -48,6 +49,7 @@ jobs:
       uses: Swatinem/rust-cache@v2.2.0
       with:
         cache-on-failure: true
+        shared-key: nydus-build-ubuntu-latest
```
imeoer, line 52: Should be enough if use `nydus-build`.

imeoer (approved):

PR #1262, opened 2023-05-06
action: fix benchmark on schedule without cache will panic

Desiki-high:
When we can't find the last benchmark result cache, the workfolow will panic.So we should cover the case when we can't find cache.

- no cache: https://github.com/dragonflyoss/image-service/actions/runs/4899401706/attempts/1#summary-13278927308
- find cache: https://github.com/dragonflyoss/image-service/actions/runs/4899401707#summary-13278921048


imeoer (approved):
LGTM!

PR #1264, opened 2023-05-07
fix warnings in Release and Integration Test

Desiki-high:
### Update actions version
1. Since we update the rust cache version in release from v1 to v2.2.0 in #1260 , the target-dir is useless, and the ./target will be cached default.
2. We should update the actions to use Node.js 16: actions/checkout@v2, actions/setup-go@v2, actions/cache@v2, actions/upload-artifact@v2.

### changelogs
1. https://github.com/Swatinem/rust-cache/releases/tag/v2.0.0 
2. https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

**Ref: warnings in https://github.com/dragonflyoss/image-service/actions/runs/4903861322**

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -45,11 +45,8 @@ jobs:
       - name: Cache cargo
         uses: Swatinem/rust-cache@v2.2.0
         with:
-          target-dir: |
-            ./target
           cache-on-failure: true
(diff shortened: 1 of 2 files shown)


imeoer (approved):
LGTM!

PR #1275, opened 2023-05-13
Fixes two bugs for rafs v5

jiangliu:
Fixes two bugs for rafs v5:
- record correct inode number for v5
- avoid a possible panic for v5

### DIFF
--- a/rafs/src/metadata/md_v5.rs
+++ b/rafs/src/metadata/md_v5.rs
@@ -172,12 +172,12 @@ impl RafsSuper {
             if let Ok(ni) = self.get_inode(next_ino, false) {
                 if ni.is_reg() {
                     let next_size = ni.size();
-                    let next_size = if next_size < window_size {
+                    let next_size = if next_size == 0 {
+                        continue;
+                    } else if next_size < window_size {
                         next_size
                     } else if window_size >= self.meta.chunk_size as u64 {
                         window_size / self.meta.chunk_size as u64 * self.meta.chunk_size as u64
-                    } else if next_size == 0 {
-                        continue;
                     } else {
                         break;
                     };
--- a/rafs/src/builder/core/bootstrap.rs
+++ b/rafs/src/builder/core/bootstrap.rs
@@ -65,6 +65,14 @@ impl Bootstrap {
         }
         ctx.prefetch.insert_if_need(&tree.node);
         nodes.push_back(tree.node.clone());
+        bootstrap_ctx.inode_map.insert(
+            (
+                tree.node.layer_idx,
+                tree.node.info.src_ino,
+                tree.node.info.src_dev,
+            ),
+            vec![tree.node.index],
+        );
 
         Self::build_rafs(ctx, bootstrap_ctx, &mut tree, &mut nodes)?;
         if ctx.fs_version.is_v6() && !bootstrap_ctx.layered {


rafs/src/builder/core/v5.rs
```diff
@@ -187,7 +187,7 @@ impl Bootstrap {
 
         // Set super block
         let mut super_block = RafsV5SuperBlock::new();
-        let inodes_count = bootstrap_ctx.inode_map.len() as u64;
+        let inodes_count = bootstrap_ctx.inode_map.len() as u64 + 1;
```
imeoer, line 190: Have this plus one before?

jiangliu: No, I found the inodes count is incorrect for v5.

jiangliu: Yes, the root inode is missing here:
```
root@903f8fcca5c8:/nydus# target/debug/nydus-image inspect images/0de6bc4212c22aa2bcc8026697cbb93db330421c7a3a09f3ae2cfa9b90c4332c 
[2023-05-15 09:35:34.626189 +00:00] INFO RAFS v5 super block features: HASH_BLAKE3 | EXPLICIT_UID_GID | COMPRESSION_ZSTD
Inspecting RAFS :> stats

    Version:                5
    Inodes Count:           1
    Chunk Size:             1024KB
    Root Inode:             1
    Flags:                  HASH_BLAKE3 | EXPLICIT_UID_GID | COMPRESSION_ZSTD
    Blob table offset:      0x2008
    Blob table size:        0x0
    Prefetch table offset:  0x2008
    Prefetch table entries: 0x0
    Chunk table offset:     0x0
    Chunk table size:       0x0
    
Inspecting RAFS :> ls
-    2        "text.txt"
Inspecting RAFS :> 
```

rafs/src/metadata/md_v5.rs
```diff
@@ -172,12 +172,12 @@ impl RafsSuper {
             if let Ok(ni) = self.get_inode(next_ino, false) {
                 if ni.is_reg() {
                     let next_size = ni.size();
-                    let next_size = if next_size < window_size {
+                    let next_size = if next_size == 0 {
+                        continue;
+                    } else if next_size < window_size {
```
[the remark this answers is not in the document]
jiangliu: Refined the commit message.

rafs/src/builder/core/v5.rs
bergwolf: Earlier on, we had
```
        tree.node.index = RAFS_ROOT_INODE;
        tree.node.inode.set_ino(RAFS_ROOT_INODE);
        // Filesystem walking skips root inode within subsequent while loop, however, we allow
        // user to pass the source root as prefetch hint. Check it here.
        ctx.prefetch.insert_if_need(&tree.node);

        bootstrap_ctx.inode_map.insert(
            (tree.node.layer_idx, tree.node.src_ino, tree.node.src_dev),
            vec![tree.node.index],
        );
```
Should we add it back instead?

rafs/src/metadata/md_v5.rs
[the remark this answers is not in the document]
bergwolf: Cannot find it. Did you forget to push the new code? :)

rafs/src/builder/core/v5.rs
jiangliu: The inode_map is used to detect hardlinks for regular files, so it doesn't make sense to add directories inode (including `/`) into the inode_map. How about adding a comments for the `+`` instead?

bergwolf: But we have all the other directory inodes (except for the root inode) in the `inode_map`, no?

jiangliu: fair enough:)

bergwolf (approved):
lgtm! Thanks!

PR #1281, opened 2023-05-17
action:clean cache after branch closes

Desiki-high:
Caches have branch scope restrictions in place, which means some caches have limited usage options. We should clean cache after branch closes.
See https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries.

### DIFF
--- a/.github/workflows/clean-cache.yml
+++ b/.github/workflows/clean-cache.yml
@@ -1,0 +1,33 @@
+name: Cleanup caches by a branch
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Cleanup
+        run: |
+          gh extension install actions/gh-actions-cache
+
+          REPO=${{ github.repository }}
+          BRANCH="refs/pull/${{ github.event.pull_request.number }}/merge"
+
+          echo "Fetching list of cache key"
+          cacheKeysForPR=$(gh actions-cache list -R $REPO -B $BRANCH | cut -f 1 )
+
+          ## Setting this to not fail the workflow while deleting cache keys.
+          set +e
+          echo "Deleting caches..."
+          for cacheKey in $cacheKeysForPR
+          do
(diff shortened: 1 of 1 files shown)


bergwolf:
> Once a repository has reached its maximum cache storage, the cache eviction policy will create space by deleting the oldest caches in the repository

Are we seeing any issues with this behavior? IOW, the description is not clear why we need to clean up caches instead of letting GH handle it.

Desiki-high:
> > Once a repository has reached its maximum cache storage, the cache eviction policy will create space by deleting the oldest caches in the repository
> 
> Are we seeing any issues with this behavior? IOW, the description is not clear why we need to clean up caches instead of letting GH handle it.

In our repo, the cache approach total cache storage limit always happened.   The reson is pr changed the depences, such as go.sum or Cargo. toml, created new cache.   After pr merged, GH will created new cache, the cache created by pr branch is still exited.
If we let GH handle it, GH will choose cache by time.  Absolutely, the cache we deleted will never used again.  So we can use this workflow to prevent cache thrashing. I think it just a optional solution, I'm not sure whther it can resolve the problem that cache approach total cache storage limit.

Desiki-high:
> > > Once a repository has reached its maximum cache storage, the cache eviction policy will create space by deleting the oldest caches in the repository
> > 
> > 
> > Are we seeing any issues with this behavior? IOW, the description is not clear why we need to clean up caches instead of letting GH handle it.
> 
> In our repo, the cache approach total cache storage limit always happened. The reson is pr changed the depences, such as go.sum or Cargo. toml, created new cache. After pr merged, GH will created new cache, the cache created by pr branch is still exited. If we let GH handle it, GH will choose cache by time. Absolutely, the cache we deleted will never used again. So we can use this workflow to prevent cache thrashing. I think it just a optional solution, I'm not sure whther it can resolve the problem that cache approach total cache storage limit.

The specific scenario is that the cache I saved in benchmark is deleted by GH once. Since benchmark doesn't run every day, the last cache might be replaced by the one generated by pr. I'm just trying to reduce the possibility of this happening.

imeoer (approved):

PR #1287, opened 2023-05-18
app: merge crate nydus-app into crate nydus

jiangliu:
Merge nydus-api crate into nydus crate, to reduce number of crates.


liubogithub (approved):
Looks OK.

PR #1292, opened 2023-05-19
cargo: fix fuse-backend-rs dependency in cargo.toml

adam3q:
#1283

## Details
The previous changes only upgraded the value in cargo.lock, we should also upgrade cargo.toml.

## Types of changes

## Checklist


liubogithub:
Can you please also put the following detail in patch commit log?
`The previous changes only upgraded the value in cargo.lock, we should also upgrade cargo.toml.`

adamqqqplay:
> Can you please also put the following detail in patch commit log? `The previous changes only upgraded the value in cargo.lock, we should also upgrade cargo.toml.`

OK, fixed.

liubogithub (approved):

PR #764, opened 2022-09-28
[WIP] storage: add ut for registry anonymous authorization

imeoer:
And update some comments for authorization workflow.

Signed-off-by: Yan Song <imeoer@linux.alibaba.com>

PR #1294, opened 2023-05-22
[WIP] storage: change token fetch to POST request

imeoer:
Try to make compatible with more registries, for example Google Artifact Registry.

## Details
Try to make compatible with more registries, for example Google Artifact Registry.

## Types of changes

## Checklist

### DIFF
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -260,7 +260,7 @@ impl RegistryState {
 
         let token_resp = connection
             .call::<&[u8]>(
-                Method::GET,
+                Method::POST,
                 auth.realm.as_str(),
                 Some(&query),
                 Some(ReqBody::Form(form)),


storage/src/backend/registry.rs
```diff
@@ -260,7 +260,7 @@ impl RegistryState {
 
         let token_resp = connection
             .call::<&[u8]>(
-                Method::GET,
+                Method::POST,
```
changweige, line 263: Could you please provide some references proving POST method is more generic?  They are too different. 🤣

imeoer: Will ref the [codes](https://github.com/containerd/containerd/blob/878132923db35f31c8c7b324d9cd1e5f59046064/remotes/docker/auth/fetch.go#L117), but we still need more [tests](https://github.com/dragonflyoss/image-service/pull/764/files).

lilic:
Hello 👋 Great project, really interested in using this in Google Artifact Registry! I am just curious if there is a reason this is still in WIP PR and if it will be picked up? Thank you!

storage/src/backend/registry.rs
philandstuff: from https://docs.docker.com/registry/spec/auth/oauth/

> Not all token servers implement oauth2. If the request to the endpoint returns 404 using the HTTP POST method, refer to [Token Documentation](https://docs.docker.com/registry/spec/auth/token/) for using the HTTP GET method supported by all token servers.

imeoer:
Fixed in master.

imeoer:
@lilic Should be fixed in https://github.com/dragonflyoss/image-service/pull/1425

Issue #1306, opened 2023-05-31
nydusify mount subcommand shouldn't require backend option


imeoer:
```
$ nydusify mount --target localhost:5000/nginx:latest-nydus
...
FATA[2023-05-31T02:20:35Z] Required flag "backend-type" not set
```

When the blobs of target image are all stored in the registry, the `nydusify mount` subcommand shouldn't require `--backend-type` and `--backend-config` options, just like the `nydusify check` subcommand.

lyj199907:
@imeoer I will try to fix it later

PR #1315, opened 2023-06-12
fix th e cmd 'mount' of nydusify

lihuahua123:
this is the solution of issue https://github.com/dragonflyoss/image-service/issues/1306.(#1306) 

## Details
we fix the bug, the nydusify `mount` subcommand will not need the `--backend-type`  and  `--backend-config` options when  the blobs of target image are all stored in the registry,

- [x] sds

## Types of changes

## Checklist


lihuahua123:
i will fix it later

PR #1317, opened 2023-06-12
Fix nydusify

lihuahua123:
this is the solution of issue https://github.com/dragonflyoss/image-service/issues/1306.(#1306) 

## Details
we fix the bug, the nydusify `mount` subcommand will not need the `--backend-type`  and  `--backend-config` options when  the blobs of target image are all stored in the registry,

## Types of changes

## Checklist

### DIFF
--- a/contrib/nydusify/cmd/nydusify.go
+++ b/contrib/nydusify/cmd/nydusify.go
@@ -26,6 +26,11 @@ import (
 	"github.com/dragonflyoss/image-service/contrib/nydusify/pkg/provider"
 	"github.com/dragonflyoss/image-service/contrib/nydusify/pkg/utils"
 	"github.com/dragonflyoss/image-service/contrib/nydusify/pkg/viewer"
+
+	"encoding/base64"
+	"encoding/json"
+	dockerconfig "github.com/docker/cli/cli/config"
+	"github.com/docker/distribution/reference"
 )
 
 var (
@@ -169,6 +174,14 @@ func getPrefetchPatterns(c *cli.Context) (string, error) {
 	}
 
 	return patterns, nil
+}
+
+type RegistryBackendConfig struct {
+	Scheme     string `json:"scheme"`
+	Host       string `json:"host"`
+	Repo       string `json:"repo"`
+	Auth       string `json:"auth,omitempty"`
+	SkipVerify bool   `json:"skip_verify,omitempty"`
 }
 
 func main() {
@@ -650,7 +663,7 @@ func main() {
 				&cli.StringFlag{
 					Name:     "backend-type",
 					Value:    "",
-					Required: true,
+					Required: false,
 					Usage:    "Type of storage backend, possible values: 'oss', 's3'",
 					EnvVars:  []string{"BACKEND_TYPE"},
 				},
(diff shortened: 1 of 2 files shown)


contrib/nydusify/cmd/nydusify.go
```diff
@@ -171,6 +176,14 @@ func getPrefetchPatterns(c *cli.Context) (string, error) {
 	return patterns, nil
 }
 
+type RegistryBackendConfig struct {
```
imeoer, line 179: Can we share the code with `check` subcommand implementation?

contrib/nydusify/cmd/nydusify.go
```diff
@@ -700,8 +713,40 @@ func main() {
 				if err != nil {
 					return err
 				} else if backendConfig == "" {
-					// TODO get auth from docker configuration file
-					return errors.Errorf("backend configuration is empty, please specify option '--backend-config'")
+
+					backendType = "registry"
```
imeoer, line 717: Can we share these codes with `check` subcommand implementation?

contrib/nydusify/pkg/viewer/viewer.go
```diff
@@ -61,11 +61,29 @@ func New(opt Opt) (*FsViewer, error) {
 	}
 
 	mode := "cached"
-	digestValidate := true
+	digestValidate := false
 	if opt.FsVersion == "6" {
 		mode = "direct"
 		digestValidate = false
 	}
+	// targetParsed, err := targetParser.Parse()
```
imeoer, line 69: Why add these useless codes here.

contrib/nydusify/cmd/nydusify.go
lihuahua123: Because the chaos of commit, i have submit a new pull-request #1317. And this quetion has been solve.

contrib/nydusify/cmd/nydusify.go
lihuahua123: I have checked this part, and find it will be bad to observe the logic of the code if it shares these codes with check subcommand implementation due to the different parameter. The code is pasted as follows
```
//contrib/nydusify/cmd/nydusify.go
if err != nil {
    return err
} else if backendConfig == "" {

    backendType = "registry"
    parsed, err := reference.ParseNormalizedNamed(c.String("target"))

    if err != nil {
        return err
    }

    host := reference.Domain(parsed)
    repo := reference.Path(parsed)

    config := dockerconfig.LoadDefaultConfigFile(os.Stderr)
    authConfig, err := config.GetAuthConfig(host)
    if err != nil {
        return errors.Wrap(err, "get docker registry auth config")
    }

    var auth string
    if authConfig.Username != "" && authConfig.Password != "" {
        auth = base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", authConfig.Username, authConfig.Password)))
    }
    skipVerify := false
    if c.Bool("target-insecure") {
        skipVerify = true
    }
    scheme := "http"

    backendConfigStruct := rule.RegistryBackendConfig{scheme, host, repo, auth, skipVerify}
    bytes, err := json.Marshal(backendConfigStruct)
    if err != nil {
        return errors.Wrap(err, "parse registry backend config")
    }
    backendConfig = string(bytes)

}


//contrib/nydusify/pkg/checker/rule/filesystem.go
if rule.NydusdConfig.BackendType == "" {
    rule.NydusdConfig.BackendType = "registry"

    if rule.NydusdConfig.BackendConfig == "" {
        config := dockerconfig.LoadDefaultConfigFile(os.Stderr)
        authConfig, err := config.GetAuthConfig(host)
        if err != nil {
            return nil, errors.Wrap(err, "get docker registry auth config")
        }

        var auth string
        if authConfig.Username != "" && authConfig.Password != "" {
            auth = base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", authConfig.Username, authConfig.Password)))
        }
        skipVerify := false
        if rule.TargetInsecure {
            skipVerify = true
        }
        scheme := "https"
        if rule.PlainHTTP {
            scheme = "http"
        }

        backendConfig := RegistryBackendConfig{scheme, host, repo, auth, skipVerify}
        bytes, err := json.Marshal(backendConfig)
        if err != nil {
            return nil, errors.Wrap(err, "parse registry backend config")
        }
        rule.NydusdConfig.BackendConfig = string(bytes)
    }
}
```

contrib/nydusify/pkg/viewer/viewer.go
lihuahua123: This part has been solved in the new pull-request #1317

contrib/nydusify/cmd/nydusify.go
imeoer: @lihuahua123 Please don't submit a new one, just use `git rebase -i $commit^` to `edit` the commit and then force push by `git push -f $branch`.

contrib/nydusify/cmd/nydusify.go
imeoer: I think we can create a func named `NewRegistryBackendConfig(...) (string, error)` to reuse the part of codes at least, right?

PR #1326, opened 2023-06-18
rafs: add special handling of invalid zero blob index

jiangliu:
Fix a possible panic in debug release caused by invalid image data.

## Details
The rafs v6 format reserves blob index 0 for meta blobs, so ensure invalid zero blob index doesn't cause abnormal behavior.

## Types of changes

## Checklist


imeoer (approved):
LGTM

PR #1333, opened 2023-06-20
feat: nydus support encrypted images

taoohong:
Extend native nydus v6 to support handling encrypted containers images:
* An encrypted nydus image is composed of encrypted bootstrap and chunk-level encrypted data blobs. The bootstrap is encrypted by the Ocicrypt and the data blobs are encrypted by aes-128-xts with randomly generated key and iv at chunk-level.
* The chunk encryption key and iv are stored in the blob info reusing some items of the structure to save reserved space.
* For every data blob, all the chunk data, conpression context. table and compression context table header are encrypted.
* Encrypted chunk data will be decrypted first, and then be decompressed while bing fetched by the storage backend.

Related [issue](https://github.com/dragonflyoss/image-service/issues/1334)


utils/src/crypt.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2022 Alibaba Cloud. All rights reserved.
+// Copyright (C) 2023 Alibaba Cloud. All rights reserved.
```
jiangliu, line 1: Should it be 2022-2023?

builder/src/compact.rs
```diff
@@ -552,6 +552,9 @@ impl BlobCompactor {
                         build_ctx.blob_features,
                         build_ctx.compressor,
                         build_ctx.digester,
+                        build_ctx.cipher,
```
jiangliu, line 555: Seems it helps to define a new helper `BlobContext::new_with_builid_ctx()`?

builder/src/core/blob.rs
```diff
@@ -194,8 +195,22 @@ impl Blob {
         if !compressed {
             compressor = compress::Algorithm::None;
         }
+
+        let encrypted_ci_data = if encrypt {
+            if let Some(cipher_ctx) = &blob_ctx.cipher_ctx {
```
jiangliu, line 200: Seems this could be abstracted as a helper.

builder/src/core/blob.rs
```diff
@@ -212,18 +227,32 @@ impl Blob {
...
-        match compressed_data {
+        match encrypted_ci_data {
             Cow::Owned(v) => blob_ctx.write_data(blob_writer, &v)?,
             Cow::Borrowed(v) => {
                 let buf = v.to_vec();
                 blob_ctx.write_data(blob_writer, &buf)?;
             }
         }
-        blob_ctx.write_data(blob_writer, header.as_bytes())?;
+        blob_ctx.write_data(blob_writer, &encrypted_header)?;
```
jiangliu, line 255: Could you please help to check whether the `to_vec()` at line 251 is necessary or not?

taoohong: It's necessary, when compressed is false, the compressed data equals &blob_ctx.blob_meta_info.as_byte_slice().

utils/src/crypt.rs
taoohong: done

bergwolf:
> The bootstrap is encrypted by the Ocicrypt and the data blobs are encrypted by aes-128-xts with randomly generated key and iv at chunk-level.

What is `iv`? And I thought we agreed on using convergent encryption for shared chunk data encryption. Any reason why it got dropped?

taoohong:
> > The bootstrap is encrypted by the Ocicrypt and the data blobs are encrypted by aes-128-xts with randomly generated key and iv at chunk-level.
> 
> The IV is used to ensure that the same plaintext block encrypted with the same data encryption key produces a different ciphertext block each time. 



> > The bootstrap is encrypted by the Ocicrypt and the data blobs are encrypted by aes-128-xts with randomly generated key and iv at chunk-level.
> 
> What is `iv`? And I thought we agreed on using convergent encryption for shared chunk data encryption. Any reason why it got dropped?

The IV is used to ensure that the same plaintext block encrypted with the same data encryption key produces a different ciphertext block each time.  It's usually suggested to be provided when using AES in xts mode.
About `Convergent encryption` you mean CipherContext with convergent_encryption set true? It's not really dropped, the related codes are moved to `utils/crypt.rs` to make them capable of being more commonly and easily used by storage or rafs subsystems.

rafs/src/metadata/layout/v6.rs
```diff
@@ -1413,11 +1423,23 @@ struct RafsV6Blob {
...
-
-    reserved2: [u8; 48],
+    // When using encryption mod, used for cipher_iv last 8 bytes.
+    // 0                  7                 15
+    // +------------------+------------------+
+    // |  blob_meta_size  | cipher_iv[8..16] |
+    // |     8bytes       |      8bytes      |
+    // +------------------+------------------+
+    //  \_         cipher_iv[0..16]        _/
+    cipher_iv: [u8; 8],
```
bergwolf, line 1438: ok, google tells that it is an cipher's `initialization vector`. But is it needed for decryption at all? If not, we should not save it in v6 blob.

bergwolf: ok, after looking at the rest of the PR, I see that IV is needed for decryption as well.

utils/src/crypt.rs
```diff
@@ -320,6 +337,74 @@ impl Cipher {
         out.truncate(count + rest);
         Ok(out)
     }
+
+    pub fn generate_key_for_aes_xts(size: usize) -> Result<Vec<u8>, Error> {
+        let mut buf = vec![0u8; size];
+        if let Err(e) = rand::rand_bytes(&mut buf) {
+            Err(eother!(format!(
+                "failed to generate key for Aes256Xts, {}",
```
bergwolf, line 345: not just Aes256Xts right ?

utils/src/crypt.rs
```diff
@@ -320,6 +337,74 @@ impl Cipher {
...
+pub struct CipherContext {
+    key: Vec<u8>,
+    iv: Vec<u8>,
+    convergent_encryption: bool,
+}
+
+impl CipherContext {
+    /// Create a new instance of [CipherContext].
+    pub fn new(key: Vec<u8>, iv: Vec<u8>, convergent_encryption: bool) -> Result<Self, Error> {
+        if key.len() != 32 {
```
bergwolf, line 374: The check and the one below, are only for Aes128Xts? If so, please rename this as `new_aes128xts`. `Cipher` is implemented to work for different cipher algorithms. It's confusing `CipherContext` only works for aes128xts. If the limitation is necessary, please make sure to highlight it with proper function names and comments.

utils/src/crypt.rs
taoohong: Yeah, I will fix it. : )

utils/src/crypt.rs
taoohong: I'll fix it. THX!

bergwolf (dismissed):
Thanks!

bergwolf:
@taoohong The implementation looks good. Could you add some integration tests (in the `smoke/tests/` directory) so that it is actually tested?

taoohong:
> Thanks!



> @taoohong The implementation looks good. Could you add some integration tests (in the `smoke/tests/` directory) so that it is actually tested?

I'd like to. THX for your review 👍

jiangliu (approved):

PR #1338, opened 2023-06-26
[WIP] nydus-image: fix unpack from specify blob dir

imeoer:
Fix the panic when user specify `--blob-dir` on unpack:

```
$ nydus-image unpack --bootstrap /path/to/bootstrap \
    --blob-dir /path/to/blobs \
    --output /path/to/oci.tar

Error: fail to unpack

Caused by:
    both blob path or blob backend config are not specified
```

## Details

## Types of changes

## Checklist

### DIFF
--- a/src/bin/nydus-image/main.rs
+++ b/src/bin/nydus-image/main.rs
@@ -1172,7 +1172,13 @@ impl Command {
 
                 Some(Arc::new(local_fs))
             }
-            None => Self::get_backend(matches, "unpacker").ok(),
+            None => {
+                Some(BlobFactory::new_backend(
+                    config.backend.as_ref().unwrap(),
+                    "unpacker",
+                )?)
+                // Self::get_backend(matches, "unpacker").ok()
+            }
         };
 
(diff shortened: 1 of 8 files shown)


adamqqqplay:
Really helpful fix, I thought it was a problem with my usage, so I could only use the --blob parameter instead of --blob-dir.

Thanks a lot!

src/bin/nydus-image/main.rs
```diff
@@ -1172,7 +1172,13 @@ impl Command {
 
                 Some(Arc::new(local_fs))
             }
-            None => Self::get_backend(matches, "unpacker").ok(),
+            None => {
+                Some(BlobFactory::new_backend(
+                    config.backend.as_ref().unwrap(),
```
jiangliu, line 1177: Is it safe to unwrap() here? Or should we check and return error code instead?

jiangliu:
Otherwise LGTM!

imeoer:
Fixed in master.

PR #1347, opened 2023-07-06
fix: merge io from same blob panic

ccx1024cc:
## Details

When merging io from same blob with different id, assertion breaks. The images without blob deduplication suffer from it.

This pr removes the assertion that requires merging in same blob index. By design, it makes sense, because different blob layer may share same blob file. A continuous read from same blob for different layer is helpful for performance.

## Types of changes

## Checklist


adam3q (approved):
LGTM, thanks for your work.

jiangliu (approved):

PR #1311, opened 2023-06-05
fix: amplify io is too large to hold in fuse buffer

ccx1024cc:
Fuse request buffer is fixed by `FUSE_KERN_BUF_SIZE * pagesize() + FUSE_HEADER_ SIZE`. When amplify io is larger than it, FuseDevWriter suffers from smaller buffer. As a result, invalid data error is returned.

##  Reproduction
    run nydusd with 3MB amplify_io
    error from random io:
        reply error header OutHeader { len: 16, error: -5, unique: 108 }, error Custom { kind: InvalidData, error: "data out of range, available 1052656 requested 1250066" }

## Details
    size of fuse buffer = 1052656 + 16 (size of inner header) = 256(FUSE_KERN_BUF_SIZE) * 4096(page size) + 4096(fuse header)
    let amplify_io = min(user_specified, fuseWriter.available_bytes())

## Resolution
   Evalucation of amplify_io will be replaced with [ZeroCopyWriter.available_bytes()]("https://github.com/cloud-hypervisor/fuse-backend-rs/pull/135").

## Relevant Issue
#1270
ci: https://github.com/dragonflyoss/image-service/actions/runs/4931748472/jobs/8814099799
#1312

PR #1348, opened 2023-07-06
fix: large files broke prefetch

ccx1024cc:
## Details

Files larger than 4G leads to prefetch panic, because the max blob io range is smaller than 4G. This pr changes blob io max size from u32 to u64.

## Types of changes

## Checklist

### DIFF
--- a/rafs/src/fs.rs
+++ b/rafs/src/fs.rs
@@ -617,12 +617,12 @@ impl FileSystem for Rafs {
 
         let real_size = cmp::min(size as u64, inode_size - offset);
         let mut result = 0;
-        let mut descs = inode.alloc_bio_vecs(&self.device, offset, real_size as usize, true)?;
-        assert!(!descs.is_empty() && !descs[0].is_empty());
+        let mut io_vecs = inode.alloc_bio_vecs(&self.device, offset, real_size as usize, true)?;
+        assert!(!io_vecs.is_empty() && !io_vecs[0].is_empty());
 
         // Try to amplify user io for Rafs v5, to improve performance.
         if self.sb.meta.is_v5() && size < self.amplify_io {
-            let all_chunks_ready = self.device.all_chunks_ready(&descs);
+            let all_chunks_ready = self.device.all_chunks_ready(&io_vecs);
             if !all_chunks_ready {
                 let chunk_mask = self.metadata().chunk_size as u64 - 1;
                 let next_chunk_base = (offset + (size as u64) + chunk_mask) & !chunk_mask;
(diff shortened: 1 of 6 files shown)


smoke/tests/tool/layer.go
```diff
@@ -44,15 +44,20 @@ func (l *Layer) CreateFile(t *testing.T, name string, data []byte) {
...
 		f.Close()
 	}()
 
-	for b := 1; b <= sizeMB; b++ {
-		_, err := f.Write(bytes.Repeat([]byte{byte(b)}, 1024*1024))
+	buf := make([]byte, 0x1000)
+	for i := 0; i < len(buf); i++ {
+		buf[i] = byte(i)
+	}
+
```
jiangliu, line 58: How about using `trunc()` to set file size instead of writing the whole file?

jiangliu: Creating a 5G file for tests is challenging:)

ccx1024cc: > Creating a 5G file for tests is challenging:)

In practice, 5G file is very slow in pack, merge and verify. By `LargerFileMaker`, texture provides the ability to create large file. No large file default. 
So far, it's only used in this case: large file in prefetch.

smoke/tests/texture/layer.go
```diff
@@ -24,14 +32,18 @@ func MakeChunkDictLayer(t *testing.T, workDir string) *tool.Layer {
 	layer.CreateFile(t, "chunk-dict-file-5", []byte("dir-1/file-2"))
 	layer.CreateFile(t, "chunk-dict-file-6", []byte("This is poetry"))
 	layer.CreateFile(t, "chunk-dict-file-7", []byte("My name is long"))
-	layer.CreateLargeFile(t, "chunk-dict-file-8", 13)
 	layer.CreateHoledFile(t, "chunk-dict-file-9", []byte("hello world"), 1024, 1024*1024)
 	layer.CreateFile(t, "chunk-dict-file-10", []byte(""))
 
+	// customized files
```
imeoer, line 38: `Customized`

smoke/tests/tool/layer.go
imeoer: Can we use `rand.Reader` and `io.CopyN` here?

smoke/tests/tool/layer.go
```diff
@@ -44,17 +46,15 @@ func (l *Layer) CreateFile(t *testing.T, name string, data []byte) {
...
 	require.NoError(t, err)
 	defer func() {
 		f.Close()
 	}()
 
-	for b := 1; b <= sizeMB; b++ {
-		_, err := f.Write(bytes.Repeat([]byte{byte(b)}, 1024*1024))
-		require.NoError(t, err)
-	}
+	_, err = io.CopyN(f, rand.Reader, 1024*1024*1024*int64(sizeGB))
```
imeoer, line 56: Maybe use bit operation here.

imeoer (commented):
others LGTM!

rafs/src/fs.rs
```diff
@@ -658,7 +658,7 @@ impl FileSystem for Rafs {
             let r = self.device.read_to(w, desc)?;
             result += r;
             recorder.mark_success(r);
-            if r as u32 != desc.size() {
+            if r as u64 != desc.size() {
```
bergwolf, line 661: desc.size() is always `u32`. So this means that somehow we end up having `self.device.read_to` to return more than 4G. Kernel never sends such large read requests. So it means that we have expanded the read IO size to larger than 4G in the above `amplify_io` logic, which is definitely wrong.... Are you testing with an `amplify_io` config larger than 4G?

Also, I wonder if this gets fixed by #1311 as well.

imeoer (approved):

PR #1349, opened 2023-07-06
fix: deprecated docker field leads to failure of nydusify check

ccx1024cc:
## Details

`NydusImage.Config.Config.ArgsEscaped` is present only for legacy compatibility with Docker and should not be used by new image builders. Nydusify (1.6 and above) ignores it, which is an expected behavior.

This pr ignores comparision of it in nydusify checking, which leads to failure.

## Types of changes

## Checklist

## Spec

[image-spec](https://github.com/opencontainers/image-spec/pull/892)

### DIFF
--- a/contrib/nydusify/pkg/checker/rule/manifest.go
+++ b/contrib/nydusify/pkg/checker/rule/manifest.go
@@ -80,6 +80,20 @@ func (rule *ManifestRule) Validate() error {
 
 	// Check Nydus image config with OCI image
 	if rule.SourceParsed.OCIImage != nil {
+
+		//nolint:staticcheck
+		// ignore static check SA1019 here. We have to assign deprecated field.
+		//
+		// Skip ArgsEscaped's Check
+		//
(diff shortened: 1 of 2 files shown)


contrib/nydusify/pkg/checker/rule/manifest.go
```diff
@@ -80,6 +80,18 @@ func (rule *ManifestRule) Validate() error {
 
 	// Check Nydus image config with OCI image
 	if rule.SourceParsed.OCIImage != nil {
+
+		//nolint:staticcheck
+		// ignore static check SA1019 here. We have to assign deprecated field.
+		//
+		// Skip ArgsEscaped's Check
+		//
+		//   This field is present only for legacy compatibility with Docker and
```
imeoer, line 89: Can we put the link https://github.com/opencontainers/image-spec/pull/892 here?

imeoer (approved):

PR #1367, opened 2023-07-11
enable log rotation by default and expose backups parameter to user

sctb512:

## Details

## Types of changes

## Checklist

### DIFF
--- a/src/bin/nydusd/main.rs
+++ b/src/bin/nydusd/main.rs
@@ -248,7 +248,7 @@ fn prepare_commandline_options() -> Command {
             Arg::new("log-rotation-size")
                 .long("log-rotation-size")
                 .help("Specify log rotation size(MB), 0 to disable")
-                .default_value("0")
+                .default_value("100")
                 .required(false)
                 .global(true),
         )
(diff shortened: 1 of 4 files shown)


src/bin/nydusd/main.rs
```diff
@@ -248,7 +248,15 @@ fn prepare_commandline_options() -> Command {
             Arg::new("log-rotation-size")
                 .long("log-rotation-size")
                 .help("Specify log rotation size(MB), 0 to disable")
-                .default_value("0")
+                .default_value("100")
```
changweige, line 251: We should change the default value of nydusd which is somewhat of a UI. Changing the usual behavior is not good.

src/bin/nydusd/main.rs
```diff
@@ -248,7 +248,15 @@ fn prepare_commandline_options() -> Command {
...
+                .required(false)
+                .global(true),
+        )
+        .arg(
+            Arg::new("log-rotation-backups")
```
changweige, line 256: At this stage, giving the ability to configure `log-rotation-size` at nydus-snapshotter side is acceptable with the hardcoded 10 backups

src/bin/nydusd/main.rs
sctb512: Got it.

src/bin/nydusd/main.rs
sctb512: If we use the hardcoded 10 backups, there no need to expose `log-rotation-backups` to users. I will close this PR later. 😄

PR #1368, opened 2023-07-12
U2.2 backport - #1229 #1311 # 1348

ccx1024cc:
## Backport

feat: add more types of file to smoke #1229
fix: amplify io is too large to hold in fuse buffer #1311
fix: large files broke prefetch #1348

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -479,9 +479,9 @@ dependencies = [
 
 [[package]]
 name = "fuse-backend-rs"
-version = "0.10.1"
+version = "0.10.4"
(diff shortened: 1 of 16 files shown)


sctb512:
These bugfix is important. Can we backport it to v2.1.0 as well?

ccx1024cc:
> These bugfix is important. Can we backport it to v2.1.0 as well?

Maybe 2.2 is a better choice. It's a LTS.

bergwolf (approved):
lgtm. I agree that we should backport it to v2.1 as well. We may decide on an EOL support date for v2.1 once v2.2 is stable enough.

PR #1375, opened 2023-07-17
storage: simplify mirror implementation

imeoer:
```
Nydusd uses a registry backend which generates a surge of blob requests without
auth tokens on initial startup. This caused mirror backends (e.g. dragonfly)
to process very slowly, the commit fixes this problem.

It implements waiting for the first blob request to complete before making other
blob requests, this ensures the first request caches a valid registry auth token,
and subsequent concurrent blob requests can reuse the cached token.

This change is worthwhile to reduce concurrent token requests, it also makes the
behavior consistent with containerd, which first requests the image manifest and
caches the token before concurrently requesting blobs.
```

## Details

## Types of changes

## Checklist

### DIFF
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -7,11 +7,11 @@ use std::collections::HashMap;
 use std::error::Error;
 use std::io::{Read, Result};
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Once, RwLock};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use std::{fmt, thread};
 
-use arc_swap::ArcSwapOption;
+use arc_swap::{ArcSwap, ArcSwapOption};
 use base64::Engine;
 use reqwest::blocking::Response;
 pub use reqwest::header::HeaderMap;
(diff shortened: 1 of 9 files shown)


storage/src/backend/registry.rs
```diff
@@ -637,18 +631,52 @@ impl RegistryReader {
...
+    /// Nydusd uses a registry backend which generates a surge of blob requests without
+    /// auth tokens on initial startup, this caused mirror backends (e.g. dragonfly)
+    /// to process very slowly. The method implements waiting for the first blob request
+    /// to complete before making other blob requests, this ensures the first request
+    /// caches a valid registry auth token, and subsequent concurrent blob requests can
+    /// reuse the cached token.
+    fn handle<F, T>(&self, handle: &mut F) -> T
+    where
+        F: FnMut() -> T,
+    {
```
sctb512, line 647: It seems this can be replaced by:
```ruet
let mut ret = None;
self.first.call_once(|| {
    // FIXME: when handle() returns Err, we need to reset the Once instance,
    // so that the retry logic still be executed in the Once.
    ret = Some(handle());
});
ret.unwrap_or_else(handle)
```

imeoer: Good point!

storage/src/backend/registry.rs
```diff
@@ -637,18 +627,50 @@ impl RegistryReader {
...
+    /// When invoking concurrently, only one of the handle methods will be executed first,
+    /// then subsequent handle methods will be allowed to execute concurrently.
+    ///
+    /// Nydusd uses a registry backend which generates a surge of blob requests without
+    /// auth tokens on initial startup, this caused mirror backends (e.g. dragonfly)
+    /// to process very slowly. The method implements waiting for the first blob request
+    /// to complete before making other blob requests, this ensures the first request
+    /// caches a valid registry auth token, and subsequent concurrent blob requests can
+    /// reuse the cached token.
+    fn handle<F, T>(&self, handle: &mut F) -> T
```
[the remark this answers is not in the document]
imeoer: I marked it with `FIXME`, but will try to resolve it.

storage/src/backend/mirror.js
sctb512, line 1: Is this file for testing purposes only?

sctb512 (approved):
LGTM

imeoer:
@jiangliu PTAL.

jiangliu (approved):

PR #1378, opened 2023-07-18
Use http/2 and keepalives 

technillogue:
## Details
We noticed connections were being dropped frequently. This adds rustls, so that http/2 ALPN works. To benefit from this more, it also sets tcp_keepalive and a connection pool. I'm not attached to the specific values for those.

It also adds a little bit of logging on throughput, which I'm happy to discard if it doesn't fit. 

## Types of changes

## Checklist

### DIFF
--- a/storage/src/backend/connection.rs
+++ b/storage/src/backend/connection.rs
@@ -10,7 +10,7 @@ use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, AtomicI16, AtomicU8, Ordering};
 use std::sync::Arc;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
-use std::{fmt, thread};
+use std::{env, fmt, thread};
 
 use log::{max_level, Level};
 
@@ -607,6 +607,11 @@ impl Connection {
         } else {
             None
         };
+        // get pool size from envvar
+        let pool_max_idle_per_host = match env::var("REGISTRY_CLIENT_POOL_MAX_IDLE_PER_HOST") {
+            Ok(val) => val.parse::<usize>().unwrap_or(20),
+            Err(_) => 20,
+        };
 
         let mut cb = Client::builder()
             .timeout(timeout)
@@ -614,7 +619,7 @@ impl Connection {
             .redirect(Policy::none())
             .use_rustls_tls()
             .tcp_keepalive(Some(Duration::from_secs(5 * 60)))
-            .pool_max_idle_per_host(20);
+            .pool_max_idle_per_host(pool_max_idle_per_host);
 
         if config.skip_verify {
             cb = cb.danger_accept_invalid_certs(true);
--- a/storage/src/backend/connection.rs
+++ b/storage/src/backend/connection.rs
@@ -611,7 +611,10 @@ impl Connection {
         let mut cb = Client::builder()
             .timeout(timeout)
             .connect_timeout(connect_timeout)
-            .redirect(Policy::none());
+            .redirect(Policy::none())
+            .use_rustls_tls()
+            .tcp_keepalive(Some(Duration::from_secs(5 * 60)))
+            .pool_max_idle_per_host(20);
 
         if config.skip_verify {
             cb = cb.danger_accept_invalid_certs(true);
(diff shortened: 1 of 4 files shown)


adam3q (commented):
Please fix broken CI first, thanks.

storage/src/backend/connection.rs
```diff
@@ -626,14 +626,19 @@ impl Connection {
         } else {
             None
         };
+        // get pool size from envvar
+        let pool_max_idle_per_host = match env::var("REGISTRY_CLIENT_POOL_MAX_IDLE_PER_HOST") {
```
imeoer, line 630: Do these idle connections take pressure on the registry server? Especially when the container is started and prefetch work is done. Maybe we can make the value as the one option of ConnectionConfig.

imeoer:
@technillogue Thanks for the PR! It seems to have some compilation errors that still need to be solved.
Does this change improve your bandwidth utilization?

storage/src/backend/connection.rs
technillogue: I don't think it would make a significant impact for reasonable values. Most servers should close idle connections if it becomes a problem.

technillogue:
> Does this change improve your bandwidth utilization?

It was a small improvement. It made less of a difference than changing merging_size, but still helped.

storage/src/backend/connection.rs
```diff
@@ -626,11 +626,19 @@ impl Connection {
...
+            Ok(val) => val.parse::<usize>().unwrap_or(20),
+            Err(_) => 20,
+        };
 
         let mut cb = Client::builder()
             .timeout(timeout)
             .connect_timeout(connect_timeout)
-            .redirect(Policy::none());
+            .redirect(Policy::none())
+            .use_rustls_tls()
```
jiangliu, line 639: why `use_rustls_tls()` here? It introduces heavy dependencies.

jiangliu: How about using `native-tls` instead? We already have vendored openssl-sys crate.
```
openssl = { version = "0.10.55", features = ["vendored"] }
```

technillogue: it's required for http/2, as far as I know native-tls does not provide for the ALPN required for http/2

storage/src/backend/connection.rs
imeoer: How about making the value as the one option of ConnectionConfig?

storage/src/backend/connection.rs
technillogue: would this be acceptable?

imeoer: It should be acceptable I think.

technillogue: Should I make any other changes to this then?

adam3q (commented):
The current commit log looks a bit messy, it would be nice if you could reorganize them.

technillogue:
@adamqqqplay, I've reorganized the commit log

storage/src/backend/connection.rs
imeoer: How about making the env `REGISTRY_CLIENT_POOL_MAX_IDLE_PER_HOST` value as the one option of ConnectionConfig?

technillogue:
Any update on this?

imeoer:
> Any update on this?

How about making the env `REGISTRY_CLIENT_POOL_MAX_IDLE_PER_HOST` value as the one option of ConnectionConfig?

storage/src/backend/connection.rs
ccx1024cc: It's better to fetch `REGISTRY_CLIENT_POOL_MAX_IDLE_PER_HOST`  from configuration.

storage/src/backend/connection.rs
```diff
@@ -607,11 +607,19 @@ impl Connection {
...
+            .tcp_keepalive(Some(Duration::from_secs(5 * 60)))
```
ccx1024cc, line 621: Should be configurable.

ccx1024cc (changes requested):

PR #1379, opened 2023-07-20
Implement nydusd and nydus-snapshotter to deliver config through the socket API

DarkMountain-wyz:
## Relevant Issue (if applicable)
https://github.com/containerd/nydus-snapshotter/issues/388

## Details
- The previous PR has been closed, so a new PR has been submitted
- Command line parameter --config now represents the URL to request daemon config from nydus-snapshotter.
- New command line parameter --config-source, representing the unix socket path for communication between nydus snapshot and nydusd.
- Nydusd is now requesting config information from nydus-snapshotter instead of obtaining config by reading the config dump file.

## Types of changes

_What types of changes does your PullRequest introduce? Put an `x` in all the boxes that apply:_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

_Go over all the following points, and put an `x` in all the boxes that apply._

- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.

PR #1382, opened 2023-07-24
parse image pull auth from env

sctb512:
## Details
This commit attempts to avoid loading auth from the nydusd configuration file, which is insecure for users. The auth loaded from env will overwrite the auth loaded from file if it is not null.

## Types of changes

## Checklist

### DIFF
--- a/api/src/config.rs
+++ b/api/src/config.rs
@@ -207,6 +207,17 @@ impl ConfigV2 {
             false
         }
     }
+
+    /// Fill authorization for registry backend.
+    pub fn update_registry_auth_info(&mut self, auth: &Option<String>) {
+        if let Some(auth) = auth {
+            if let Some(backend) = self.backend.as_mut() {
+                if let Some(registry) = backend.registry.as_mut() {
+                    registry.auth = Some(auth.to_string());
+                }
+            }
+        }
+    }
 }
 
 impl FromStr for ConfigV2 {
--- a/api/src/config.rs
+++ b/api/src/config.rs
@@ -2134,4 +2134,48 @@ mod tests {
         assert_eq!(&config.id, "id1");
         assert_eq!(config.backend.as_ref().unwrap().backend_type, "localfs");
     }
+
+    #[test]
+    fn test_update_registry_auth_info() {
+        let config = r#"
+        {
+            "device": {
+              "id": "test",
+              "backend": {
+                "type": "registry",
+                "config": {
+                    "readahead": false,
+                    "host": "docker.io",
+                    "repo": "library/nginx",
+                    "scheme": "https",
+                    "proxy": {
+                        "fallback": false
+                    },
+                    "timeout": 5,
+                    "connect_timeout": 5,
+                    "retry_limit": 8
+                }
+              }
+            },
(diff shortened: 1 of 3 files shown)


src/bin/nydusd/main.rs
```diff
@@ -375,13 +375,18 @@ fn process_fs_service(
     let bootstrap = args.value_of("bootstrap");
     // safe as virtual_mountpoint default to "/"
     let virtual_mnt = args.value_of("virtual-mountpoint").unwrap();
+    // auth for registry
+    let auth = std::env::var("IMAGE_PULL_AUTH").ok();
+
+    info!("[abin] auth: {auth:?}");
```
jiangliu, line 381: Please help to remove this info message, which may leak secret information through log files.

jiangliu:
There's a PR to enable passing nydusd configuration information through pipe instead of a file, seems these two PRs are both to enhance the security of auth info. So what's the relationship between the two PRs?

sctb512:
> There's a PR to enable passing nydusd configuration information through pipe instead of a file, seems these two PRs are both to enhance the security of auth info. So what's the relationship between the two PRs?
> 

The PR https://github.com/dragonflyoss/image-service/pull/1379 attempts to propose an interface for nydus-snapshotter to update configuration (including auth). This PR avoids loading auth from disk when starting `nydusd`. I think they are not in conflict with each other.

In short, we can parse auth from the environment variable and update it via pipe.

src/bin/nydusd/main.rs
sctb512: Done. This message is for testing purposes only.

api/src/config.rs
```diff
@@ -207,6 +207,19 @@ impl ConfigV2 {
             false
         }
     }
+
+    /// Fill authorization for registry backend.
+    pub fn fill_auth(&self, auth: &Option<String>) -> Self {
```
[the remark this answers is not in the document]
jiangliu: Rename to `update_auth()`?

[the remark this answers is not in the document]
sctb512: Good point. Fixed.

api/src/http.rs
```diff
@@ -31,6 +31,8 @@ pub struct ApiMountCmd {
     pub fs_type: String,
     /// Configuration for the filesystem.
     pub config: String,
+    /// Authorization for private registry.
+    pub auth: Option<String>,
```
jiangliu, line 35: Why is this extension needed? I think callers should merge the auth info into `config`.

sctb512: Yes, you are correct. Fixed it now.

jiangliu (approved):

adam3q (commented):
Maybe we can prefix the first commit log with nydusd? `nydusd: parse image pull auth from env `

sctb512:
> Maybe we can prefix the first commit log with nydusd? `nydusd: parse image pull auth from env `

Done.

adam3q (approved):
LGTM, Thanks!

changweige (commented):
LGTM, thanks

changweige (approved):

PR #1387, opened 2023-08-02
deps: update tar-rs to handle very large uid/gid in image unpack

adam3q:
https://github.com/alexcrichton/tar-rs/issues/332

## Details
update tar-rs to support read large uid/gid from PAX extensions to fix very large UIDs/GIDs (>=2097151, limit of USTAR tar) lost in PAX style tar during unpack.

## Types of changes

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,3 +107,7 @@ backend-s3 = ["nydus-storage/backend-s3"]
 
 [workspace]
 members = ["api", "builder", "clib", "rafs", "storage", "service", "utils"]
+
+[patch.crates-io]
+# Due to the slow processing of PR by the upstream, temporarily fork tar-rs for this project
+tar = { git = "https://github.com/nydusaccelerator/tar-rs.git" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1759,9 +1759,8 @@ dependencies = [
 
 [[package]]
 name = "tar"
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -107,3 +107,6 @@ backend-s3 = ["nydus-storage/backend-s3"]
 
 [workspace]
 members = ["api", "builder", "clib", "rafs", "storage", "service", "utils"]
+
+[patch.crates-io]
+tar = { git = "https://github.com/nydusaccelerator/tar-rs.git" }
```
imeoer, line 112: Maybe add a comment here explaining why we need to use the forked code.

adamqqqplay: OK, comment added.

liubogithub:
It's important to have a regression test for a tricky issue like this.

But I think we can go without it for now.

adamqqqplay:
> It's important to have a regression test for a tricky issue like this.
> 
> But I think we can go without it for now.

@liubogithub I added a test image at: ghcr.io/dragonflyoss/image-service/pax-uid-test to verify this is fixed.

adamqqqplay:
E2E is successfully tested here: https://github.com/adamqqqplay/image-service/actions/runs/5746313149/job/15575642277

imeoer (approved):
LGTM

PR #1395, opened 2023-08-07
Fix image-create with ACLs. Fixes #1394.

### LINKED ISSUE #1394: Bug: nydus-image create does not work when ACLs are set
### Version of nydus being used (nydusd --version)

```
nydus-image --version
# >>>
Version:   v2.2.2
Git Commit:     19d5b12bb0ca58d0474861416e91961169235114
Build Time:     2023-07-17T04:12:08.369576349Z
Profile:   release
Rustc:     rustc 1.66.1 (90743e729 2023-01-10)
```

## Details about issue

To reproduce: 

```
mkdir blobs
install -D /dev/null files/file_with_acl
setfacl -m user:root:rwx files/file_with_acl
xattr files/file_with_acl  # >>> system.posix_acl_access
nydus-image create -t dir-rafs -D blobs files
# >>>
# [2023-08-07 16:06:23.603097 -04:00] ERROR [/src/error.rs:22] Error:
#  "invalid xattr key"
#  at rafs/src/metadata/layout/mod.rs:268
#  note: enable `RUST_BACKTRACE=1` env to display a backtrace
# Error: build failed
#
# Caused by:
#     0: failed to create node "files/file_with_acl"
#     1: failed to build inode
#     2: Invalid argument (os error 22)
```

dhbaird:

Fixes:
https://github.com/dragonflyoss/image-service/issues/1394

## Types of changes

## Checklist

### DIFF
--- a/rafs/src/metadata/layout/mod.rs
+++ b/rafs/src/metadata/layout/mod.rs
@@ -267,7 +267,7 @@ impl RafsXAttrs {
             return Err(einval!("xattr key/value is too big"));
         }
         for p in RAFS_XATTR_PREFIXES {
-            if buf.len() > p.as_bytes().len() && &buf[..p.as_bytes().len()] == p.as_bytes() {
+            if buf.len() >= p.as_bytes().len() && &buf[..p.as_bytes().len()] == p.as_bytes() {
                 self.pairs.insert(name, value);
                 return Ok(());
             }
--- a/rafs/src/metadata/layout/v6.rs
+++ b/rafs/src/metadata/layout/v6.rs
@@ -2085,7 +2085,7 @@ impl RafsXAttrs {
             for (key, value) in self.pairs.iter() {
                 let (index, prefix_len) = Self::match_prefix(key)
                     .map_err(|_| einval!(format!("invalid xattr key {:?}", key)))?;
-                if key.len() <= prefix_len {
+                if key.len() < prefix_len {
                     return Err(einval!(format!("invalid xattr key {:?}", key)));
                 }
                 if value.len() > u16::MAX as usize {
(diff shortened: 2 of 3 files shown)


rafs/src/metadata/layout/mod.rs
```diff
@@ -267,7 +267,7 @@ impl RafsXAttrs {
             return Err(einval!("xattr key/value is too big"));
         }
         for p in RAFS_XATTR_PREFIXES {
-            if buf.len() > p.as_bytes().len() && &buf[..p.as_bytes().len()] == p.as_bytes() {
+            if buf.len() >= p.as_bytes().len() && &buf[..p.as_bytes().len()] == p.as_bytes() {
```
imeoer, line 270: Thanks for the fixup, it seems we need also to handle this?
https://github.com/dragonflyoss/image-service/blob/34ee8255dea73819f4d634f83260d1fa4fe68c97/rafs/src/metadata/layout/v6.rs#L2088

dhbaird: Perhaps so; I am not 100% sure. I apologize that I have yet produced a test in bats to correspond to testing this. Thank you for the time and consideration.

dhbaird: *That I have not yet produced, I mean!

imeoer: Hi @dhbaird, we can add a related test file here https://github.com/dragonflyoss/image-service/blob/34ee8255dea73819f4d634f83260d1fa4fe68c97/smoke/tests/texture/layer.go#L89

imeoer:
@dhbaird Please also add a signed-off-by by following https://github.com/dragonflyoss/image-service/pull/1395/checks?check_run_id=15694892551

dhbaird:
Thanks for the guidance @imeoer! I confirmed that the line you indicated above should change: https://github.com/dragonflyoss/image-service/blob/34ee8255dea73819f4d634f83260d1fa4fe68c97/rafs/src/metadata/layout/v6.rs#L2088

 I am working on the tests now and double checking the changes, and will also added signed-off-by.

dhbaird:
Hi again @imeoer, When I tried to update the test indicated in previous comments, I got blocked by a new issue: `toOCITar()` drops all xattrs except for `security.capability`:

https://github.com/dragonflyoss/image-service/blob/34ee8255dea73819f4d634f83260d1fa4fe68c97/smoke/tests/tool/layer.go#L119

I read the code and traced it to containerd/archive, where it appears to ignore all xattrs except for `security.capability`:

https://github.com/containerd/containerd/blob/70a2c95ae8c02d7a4e448f0e4fb8bb0e6344b5c7/archive/tar.go#L644

Does this look like an issue in containerd/archive now too? Would you recommend that I create an issue over there too?

Thanks!

imeoer:
> Does this look like an issue in containerd/archive now too? Would you recommend that I create an issue over there too?

@dhbaird It seems only the `security.capability` xattr is supported in OCI tar, so we can't add the test case into the build of `oci tar -> nydus rafs`, but still can support it in the build of `directory -> nydus rafs`.

But to keep it simple, I think we just need to add some unit tests for `RafsXAttrs`, what do you think?

imeoer (approved):
Thanks @dhbaird for the excellent tests and fixup!

PR #1399, opened 2023-08-09
local cas: achieve chunk deduplication for nydus.

xwb1136021767:

## Details
Original version from https://github.com/dragonflyoss/image-service/pull/956.
The previous version of local cas was static dedup, which only modified the chunk information in bootstrap. There is a serious problem: it may reuse chunks that cannot be obtained by the backend of the current image, resulting in the container being unable to load the corresponding chunk data on demand during runtime. To address this issue, dynamic dedup was introduced. When nydusd initializes the blob cache, it reads the corresponding backend configuration information of the blob from the CAS database, enabling the blob cache to read chunk data from other backend.

What's more, the mismatch between dynamic dedup and nydus' chunk amplification can result in a larger cache size after dedup than without dedup. Because chunk amplification can cause reused chunks to be pulled multiple times, resulting in a larger cache size after dedup is enabled than when dedup is not enabled. To address this issue, a dedup_bitmap was introduced. When initializing rafs, dedup_bitmap is generated based on the chunk information in blob. The determination of whether a chunk in a blob is ready requires both the chunk map and deduplication bitmap to make a joint decision.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)
- [ x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist


Cargo.lock
```diff
@@ -2422,15 +2424,6 @@ dependencies = [
...
 
-[[package]]
-name = "xattr"
-version = "0.2.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d1526bbe5aaeb5eb06885f4d987bcdfa5e23187055de9b83fe00156a821fabc"
-dependencies = [
- "libc",
-]
-
```
adamqqqplay: Maybe xattr changes in Cargo.lock could merge to first commit?

imeoer:
@xwb1136021767 Sorry for the delayed reply, please help to add some docs about background and usage for users.

api/src/config.rs
```diff
@@ -29,6 +29,8 @@ pub struct ConfigV2 {
     pub cache: Option<CacheConfigV2>,
     /// Configuration information for RAFS filesystem.
     pub rafs: Option<RafsConfigV2>,
+    /// Configuration information for image deduplication.
+    pub deduplication: Option<DeduplicationConfigV2>,
```
ccx1024cc, line 33: Please add [smoke test](https://github.com/dragonflyoss/nydus/tree/master/smoke/tests) and update [doc](https://github.com/dragonflyoss/nydus/tree/master/docs)

builder/src/core/bootstrap_dedup.rs
```diff
@@ -0,0 +1,166 @@
...
+        let mut blob_mgr = BlobManager::new(digest::Algorithm::Sha256);
+        blob_mgr
+            .extend_from_blob_table(&build_ctx, self.rs.superblock.get_blob_infos())
+            .unwrap();
+
+        // Dump bootstrap
+        timing_tracer!(
+            {
+                tree.walk_bfs(true, &mut |n| {
+                    n.lock_node().dedup_chunk_for_node(
```
ccx1024cc, line 127: `get_chunk_ofs` seems not implemented with v5. Is `dedup_chunk_for_node` able to work with v5? If not please use `is_v6` to avoid it.

src/bin/nydus-image/main.rs
```diff
@@ -705,8 +706,33 @@ fn prepare_cmd_args(bti_string: &'static str) -> App {
                     .long("output")
                     .help("path for output tar file")
                     .required(true),
-            ),
-    )
+                )
+        )
+        .subcommand(
+            App::new("dedup")
```
ccx1024cc, line 712: Smoke test is required. Please also update test doc.

storage/src/device.rs
```diff
@@ -1611,4 +1750,89 @@ mod tests {
         assert_eq!(size, iovec.size());
         assert_eq!(chunk_count, iovec.len() as u32);
     }
+
+    #[test]
+    fn test_serialize_cipher_object() {
```
ccx1024cc, line 1755: Assertions is necessary that ensures cipher ok.

storage/src/device.rs
```diff
@@ -1611,4 +1750,89 @@ mod tests {
...
+                }
+            };
+            blob_info.set_cipher_info(algo, Arc::new(cipher_object), cipher_ctx);
+
+            let content = serde_json::to_string(&blob_info).unwrap();
+            let blob_info_from_deserialize: BlobInfo = serde_json::from_str(&content).unwrap();
+            println!("{}", algo);
+            println!("origin blob_info: {:?}", blob_info);
+
+            println!(
```
ccx1024cc, line 1811: Please remove these debug information for merge

ccx1024cc (changes requested):

isidentical:
This is a particularly cool feature that I think would benefit everyone who deals with multiple random images that essentially share the same stuff thats used at runtime (e.g. `libpython3.so`) but built by different methodologies (different base images etc) and would optimize cold starts in a very intriguing way. Any chance on knowing what are the next steps for this PR to get streamlined aside from the existing reviews (which all seems easily addressable, so am not sure if they are a blocker)? Is the concept itself is OK for nydus maintainers? Are there any help needed which I could contribute to (am not super knowledgeable on nydus side, but would be happy to dive deep into fixing the easy stuff like the existing reviews) to push this forward?

imeoer:
@xwb1136021767 Can we also give a performance test result in the doc to give users confidence?

utils/Cargo.toml
```diff
@@ -17,6 +17,7 @@ libc = "0.2"
 log = "0.4"
 lz4-sys = "1.9.4"
 lz4 = "1.24.0"
+rusqlite = { version = "0.29.0", features = ["bundled"] }
```
jiangliu, line 20: please update to latest "0.30.0".

utils/src/cas.rs
```diff
@@ -0,0 +1,450 @@
...
+        conn.execute(
+            "CREATE TABLE IF NOT EXISTS BlobInfos_V5 (
+            BlobId     TEXT NOT NULL PRIMARY KEY,
+            BlobInfo   TEXT NOT NULL,
+            Backend    TEXT NOT NULL
+        )",
+            (),
+        )?;
+        conn.execute(
+            "CREATE INDEX IF NOT EXISTS BlobIndex_V5 ON BlobInfos_V5(BlobId)",
```
jiangliu, line 63: `BlobInfos_V5(BlobId)` is a primary key, should it have already been indexed?

utils/src/cas.rs
```diff
@@ -0,0 +1,450 @@
...
+            (),
+        )?;
+
+        conn.execute(
+            "CREATE TABLE IF NOT EXISTS ChunkInfos_V5 (
+            ChunkId           INTEGER PRIMARY KEY,
+            ChunkDigestValue  TEXT NOT NULL,
+            ChunkInfo         TEXT NOT NULL,
+            BlobId            TEXT NOT NULL,
```
jiangliu, line 72: Should we use an auto-incr integer as primary key for BlobInfos_V5, so we can store an integer instead of a text here.

utils/src/cas.rs
```diff
@@ -0,0 +1,450 @@
...
+    fn test_get_all_blobs() {
+        let path = PathBuf::from("/tmp/local-cas");
+        match fs::create_dir_all(path.clone()) {
+            Ok(()) => println!("Directory created!"),
+            Err(e) => println!("Error: {:?}", e),
+        };
+        let cas_mgr = CasMgr::new(path).unwrap();
+        let vec = cas_mgr.get_all_blobs(false).unwrap();
+
+        println!("v5 blobs");
```
jiangliu, line 379: The `println!` lines should be removed

jiangliu:
> This is a particularly cool feature that I think would benefit everyone who deals with multiple random images that essentially share the same stuff thats used at runtime (e.g. `libpython3.so`) but built by different methodologies (different base images etc) and would optimize cold starts in a very intriguing way. Any chance on knowing what are the next steps for this PR to get it streamlined aside from the existing reviews (which all seems easily addressable, so am not sure if they are a blocker)? Is the concept itself is OK for nydus maintainers? Are there any help needed which I could contribute to (am not super knowledgeable on nydus side, but would be happy to dive deep into fixing the easy stuff like the existing reviews) to push this forward?

This idea is warmly welcomed:) 
The only issue is a little bigger for review, will put more effort on this.

jiangliu:
@xwb1136021767 When reviewing the PR, I have worked out some enhancements/fixes over your work. How about holding on for while, then I will try to submit my patches to your repo?

xwb1136021767:
> @xwb1136021767 When reviewing the PR, I have worked out some enhancements/fixes over your work. How about holding on for while, then I will try to submit my patches to your repo?

Thank you so much for your help!

api/src/config.rs
```diff
@@ -1024,11 +1041,59 @@ impl From<&BlobCacheEntryConfigV2> for ConfigV2 {
...
+    pub work_dir: String,
+}
+
+impl DeduplicationConfigV2 {
+    /// Validate image deduplication configuration.
+    pub fn validate(&self) -> bool {
+        true
+    }
+
+    pub fn get_enable(&self) -> bool {
```
imeoer, line 1068: It's a pub field, still need this? Also better be `is_enable`.

api/src/config.rs
```diff
@@ -1325,6 +1390,28 @@ impl TryFrom<&CacheConfig> for CacheConfigV2 {
     }
 }
 
+/// Configuration information for image deduplication.
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+struct DeduplicationConfig {
+    /// Whether to enable image dedup
+    #[serde(default)]
+    pub enable: bool,
+    /// Work fir for image dedup
```
imeoer, line 1402: typo `dir`.

builder/src/core/bootstrap.rs
```diff
@@ -2,16 +2,34 @@
 // Copyright (C) 2023 Alibaba Cloud. All rights reserved.
 //
 // SPDX-License-Identifier: Apache-2.0
-
-use anyhow::{Context, Error, Result};
+#![allow(unused_variables, unused_imports)]
```
imeoer, line 5: Why we need this.

builder/src/core/bootstrap.rs
```diff
@@ -209,4 +227,215 @@ impl Bootstrap {
...
+            (old_blob_table_offset, bootstrap_end)
+        };
+        //write rs
+        sb.set_blob_table_offset(blob_table_offset as u64);
+        sb.set_blob_table_size(blob_table_size as u32);
+        sb.set_extended_blob_table_offset(ext_blob_table_offset as u64);
+        sb.set_extended_blob_table_entries(u32::try_from(blob_table.extended.entries())?);
+        writer.seek(SeekFrom::Start(0))?;
+        sb.store(writer).context("failed to store superblock")?;
+        //rewrite blob table
```
imeoer, line 287: Ditto.

imeoer:
I can't understand why we need to rebuild a new bootstrap, can we do runtime de-duplication by enhancing the blob cache? i.e. for example using a global `(chunk_digest, chunk_info)` db records.

imeoer (changes requested):
I can't understand why we need to rebuild a new bootstrap, can we do runtime de-duplication by enhancing the blob cache? i.e. for example using a global (chunk_digest, chunk_info) db records.

xwb1136021767:
> I can't understand why we need to rebuild a new bootstrap, can we do runtime de-duplication by enhancing the blob cache? i.e. for example using a global (chunk_digest, chunk_info) db records.

I have also considered the solution you mentioned before. My previous concern was whether runtime deduplication would make the CAS database a performance bottleneck in IO intensive scenarios. There are indeed problems with the current solution, and personally, I think the biggest problem is how to combine it with the GC mechanism of the image, especially in localfs or localdisk mode, because nydus snapshot cannot prevent the deletion operation of the image.

imeoer:
> I have also considered the solution you mentioned before. My previous concern was whether runtime deduplication would make the CAS database a performance bottleneck in IO intensive scenarios. There are indeed problems with the current solution, and personally, I think the biggest problem is how to combine it with the GC mechanism of the image, especially in localfs or localdisk mode, because nydus snapshot cannot prevent the deletion operation of the image.

This may require some testing, full dedup and rebuilding the whole bootstrap (not only chunk infos?) on booting may also affect cold start performance. Either full dedup or on-demand dedup, there will be GC issues as containerd checks if an image is referencing a blob based on the image manifest, references:

https://github.com/containerd/nydus-snapshotter/blob/5009c522df583cdf76ee37ae2a3c6440d5d79797/snapshot/snapshot.go#L583

https://github.com/containerd/nydus-snapshotter/blob/5009c522df583cdf76ee37ae2a3c6440d5d79797/pkg/cache/manager.go#L70

jiangliu:
A simplified version to implement chunk deduplication at runtime instead of at build time.
#1507

Issue #1410, opened 2023-08-29
failed to build image for targz-ref when use regular targz file


zyfjeff:
## Additional Information
```
nydus-image create -t
                "targz-ref",
                "--features",
                "blob-toc",
                "--blob-dir",
                "/tmp/nydus/blobs/",
                "--bootstrap",
                "/tmp/nydus/bootstrap",
/d19f32bd9e4106d487f1a703fc2f09c8edadd92db4405d477978e8e466ab290d",
            ],
```

```
[2023-08-29 10:43:19.592151 +08:00] ERROR [/src/error.rs:23] Error:
        "seek() not supported!"
        at builder/src/tarball.rs:84
        note: enable `RUST_BACKTRACE=1` env to display a backtrace
```

### Version of nydus being used (nydusd --version)

Version:        v2.2.0-302-g1abf0aeb
Git Commit:     1abf0aeb84b703b21becd4fba858f1d7284bb5cf
Build Time:     2023-08-29T02:47:13.121837300Z
Profile:        debug
Rustc:          rustc 1.68.2 (9eb3afe9e 2023-03-27)

imeoer:
It seems we only tested the FIFO file.

PR #1411, opened 2023-08-29
 Add seekable method for TarReader use to determine whether the current reader supports seek

zyfjeff:
#1410 

## Details
 Add seekable method for TarReader use to determine whether the current reader supports seek

## Types of changes

## Checklist

### DIFF
--- a/builder/src/tarball.rs
+++ b/builder/src/tarball.rs
@@ -62,6 +62,15 @@ enum TarReader {
     ZranReader(ZranReader<File>),
 }
 
+impl TarReader {
+    fn seekable(&self) -> bool {
+        matches!(
+            self,
+            TarReader::File(_) | TarReader::BufReaderInfoSeekable(_)
+        )
+    }
+}
+
 impl Read for TarReader {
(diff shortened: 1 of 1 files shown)


builder/src/tarball.rs
```diff
@@ -78,10 +87,20 @@ impl Read for TarReader {
 
 impl Seek for TarReader {
     fn seek(&mut self, pos: SeekFrom) -> std::io::Result<u64> {
+        let seekable = self.seekable();
         match self {
-            TarReader::File(f) => f.seek(pos),
-            TarReader::BufReaderInfoSeekable(b) => b.seek(pos),
-            _ => Err(enosys!("seek() not supported!")),
+            TarReader::File(f) => {
+                debug_assert!(seekable);
```
imeoer, line 93: We do not recommend using `debug_assert!`, as this may cause inconsistent binary behavior between debug and release builds.

zyfjeff: fixed, use assert!

zyfjeff:
retest

imeoer (approved):

PR #1412, opened 2023-08-29
bugfix: do not fill 0 buffer, and skip validate features

zyfjeff:
1. Buffer reset to 0 will cause race during concurrency.

2. Previously, the second validate_header did not actually take effect. Now it is repaired, and it is found that the features of blob info do not set the --inline-bootstrap position to true, so the check of features is temporarily skipped. Essentially needs to be fixed from nydus-image from upstream.

## Details

## Types of changes

## Checklist


imeoer (approved):
LGTM

PR #1417, opened 2023-09-04
storage: fix chunk map compatibility

imeoer:
The blob cache file of nydusd v2.2 and <=v2.1 are in different
formats, which are not compatible. Should use different chunk map
files for them, in order to upgrade or downgrade smoothly.

For the nydusd <=v2.1, the files in blob cache directory:

```
$blob_id
$blob_id.chunk_map
```

For the nydusd =v2.2, the files in blob cache directory:

```
$blob_id.blob.data
$blob_id.chunk_map
```

NOTE: nydusd (v2.2) maybe use the chunk map file of nydusd(<=v2.1),
it will cause the corrupted blob cache data to be read.

For the nydusd of current patch, the files in blob cache directory:

```
$blob_id.blob.data
$blob_id.blob.data.chunk_map
```

NOTE: this will discard the old blob cache data and chunk map files.

## Types of changes

## Checklist

### DIFF
--- a/storage/src/cache/filecache/mod.rs
+++ b/storage/src/cache/filecache/mod.rs
@@ -25,6 +25,9 @@ use crate::cache::{BlobCache, BlobCacheMgr};
 use crate::device::{BlobFeatures, BlobInfo};
 use crate::RAFS_DEFAULT_CHUNK_SIZE;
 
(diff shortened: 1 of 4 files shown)


imeoer:
The compatibility issue is found by @ccx1024cc and @power-more, @ccx1024cc provided a repair idea.

jiangliu (approved):

PR #1418, opened 2023-09-06
nydus-overlayfs: filter option "io.katacontainers.volume"

jiangliu:
Filter mount option "io.katacontainers.volume", which is a superset of "extraoptions".

## Types of changes

## Checklist

### DIFF
--- a/contrib/nydus-overlayfs/cmd/main.go
+++ b/contrib/nydus-overlayfs/cmd/main.go
@@ -13,7 +13,11 @@ import (
 )
 
 const (
+	// Extra mount option to pass Nydus specific information from snapshotter to runtime through containerd.
 	extraOptionKey = "extraoption="
+	// Kata virtual volume infmation passed from snapshotter to runtime through containerd, superset of `extraOptionKey`.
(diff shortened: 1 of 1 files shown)


contrib/nydus-overlayfs/cmd/main.go
```diff
@@ -13,7 +13,8 @@ import (
 )
 
 const (
-	extraOptionKey = "extraoption="
+	extraOptionKey      = "extraoption="
+	kataVolumeOptionKey = "io.katacontainers.volume="
```
imeoer, line 17: Can add some comments to explain why we ignore the option key here?

adamqqqplay: Yeah, I'm confused too.

jiangliu: This is to make use of the `KataVirtualVolume` abstraction from
https://github.com/kata-containers/kata-containers/blob/main/src/libs/kata-types/src/mount.rs#L233

imeoer (approved):

PR #1424, opened 2023-09-13
feat: notifying  incosistent fs version problem with exit code

PerseidMeteor:
If acceld converts with different fs version cache, leading to an inconsistent fs version problem when merging into boostrap layer. So we need to notify acceld that an inconsistent version occured and handle this error.

## Details
Define `InconsistentFsVersion` with `thiserror` and check error type after merge, if is fs version inconsistent problem, just return status code 2.

## Types of changes

## Checklist

### DIFF
--- a/rafs/src/metadata/mod.rs
+++ b/rafs/src/metadata/mod.rs
@@ -17,8 +17,9 @@ use std::path::{Component, Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
+use thiserror::Error;
 
-use anyhow::bail;
+use anyhow::{bail, ensure};
 use fuse_backend_rs::abi::fuse_abi::Attr;
 use fuse_backend_rs::api::filesystem::Entry;
 use nydus_api::{ConfigV2, RafsConfigV2};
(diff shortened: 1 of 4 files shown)


rafs/src/metadata/mod.rs
```diff
@@ -397,34 +399,44 @@ pub struct RafsSuperConfig {
     pub is_tarfs_mode: bool,
 }
 
+#[derive(Error, Debug)]
+pub enum MergeError {
```
imeoer, line 403: How about just using this:

```
pub enum MergeError {
    #[error("Inconsistent RAFS Filesystem: {0}")]
    InconsisentFilesystem(String),
    #[error(transparent)]
    Other(#[from] Error),
}
```

builder/src/merge.rs
```diff
@@ -83,6 +83,7 @@ impl Merger {
         target: ArtifactStorage,
         chunk_dict: Option<PathBuf>,
         config_v2: Arc<ConfigV2>,
+        expected_fs_version: Option<u32>,
```
imeoer, line 86: It seems we need to also consider other options like `chunk_size`, `digester`, I think maybe it's unnecessary to check this.

rafs/src/metadata/mod.rs
PerseidMeteor: ok, I droped the `DifferentDigestAlgorithm(String)`

src/bin/nydus-image/main.rs
```diff
@@ -759,7 +759,13 @@ fn main() -> Result<()> {
             }
         }
     } else if let Some(matches) = cmd.subcommand_matches("merge") {
-        Command::merge(matches, &build_info)
+        let result = Command::merge(matches, &build_info);
+        if let Err(e) = result {
```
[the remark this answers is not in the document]
PerseidMeteor: we may can do it `if let MergeError::InconsisentFsVersion(_) = result.unwrap_err()`

builder/src/merge.rs
PerseidMeteor: Dropped this commit temporarily

rafs/src/metadata/mod.rs
```diff
@@ -397,34 +398,42 @@ pub struct RafsSuperConfig {
...
+    #[error("Inconsistent RAFS version:{0}")]
+    InconsisentFsVersion(String),
```
imeoer, line 404: Better be

```
#[error("Inconsistent RAFS Filesystem: {0}")]
nconsisentFilesystem(String),
```

src/bin/nydus-image/main.rs
```diff
@@ -759,7 +759,11 @@ fn main() -> Result<()> {
             }
         }
     } else if let Some(matches) = cmd.subcommand_matches("merge") {
-        Command::merge(matches, &build_info)
+        let result = Command::merge(matches, &build_info);
+        if let Err(MergeError::InconsisentFsVersion(_)) = result {
+            std::process::exit(2);
+        }
+        Ok(())
```
imeoer, line 766: Other errors should also be handled here?

rafs/src/metadata/mod.rs
PerseidMeteor: There will be a warning if doesn't have an upper camel case name.May be follows better.
```
#[error("Inconsistent RAFS Filesystem: {0}")]
NconsisentFilesystem(String),
```

src/bin/nydus-image/main.rs
PerseidMeteor: yep

rafs/src/metadata/mod.rs
```diff
@@ -397,34 +398,44 @@ pub struct RafsSuperConfig {
     pub is_tarfs_mode: bool,
 }
 
+#[derive(Error, Debug)]
+pub enum MergeError {
+    #[error("Inconsistent RAFS Filesystem: {0}")]
+    NconsisentFilesystem(String),
```
imeoer, line 404: `InconsistentFilesystem(String),` ?

rafs/src/metadata/mod.rs
```diff
@@ -397,34 +398,44 @@ pub struct RafsSuperConfig {
...
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
+    #[error(transparent)]
+    IOError(#[from] std::io::Error),
```
imeoer, line 408: Is it necessary?

PerseidMeteor: yep, because these two func return an `std::io::Error`. If we don't have an `IOError`  from, they can not be converted to an MergeError automaticly.

builder/src/merge.rs
```diff
@@ -83,49 +83,56 @@ impl Merger {
...
         }
         if let Some(digests) = blob_digests.as_ref() {
-            ensure!(
-                digests.len() == sources.len(),
-                "number of blob digest entries {} doesn't match number of sources {}",
-                digests.len(),
-                sources.len(),
-            );
+            if digests.len() != sources.len() {
+                Err(MergeError::Other(anyhow!(
```
imeoer, line 94: Does this let the `merge` func return the error immediately?

rafs/src/metadata/mod.rs
imeoer: But it seems also to be converted to `MergeError::Other` in `OutputSerializer::dump(matches, output, build_info).map_err(MergeError::Other)`.

builder/src/merge.rs
PerseidMeteor: maybe we can do it use `return`

PerseidMeteor: Maybe we can use `return`.

rafs/src/metadata/mod.rs
PerseidMeteor: > yep, because these two func return an `std::io::Error`. If we don't have an `IOError` from, they can not be converted to an MergeError automaticly.

The func is `blob_ctx.blob_id =
                                BlobInfo::get_blob_id_from_meta_path(bootstrap_path)?;`
We can manually convert it to `MergeError.`

builder/src/merge.rs
```diff
@@ -222,8 +234,11 @@ impl Merger {
                         {
                             blob_ctx.blob_id = original_id;
                         } else {
-                            blob_ctx.blob_id =
-                                BlobInfo::get_blob_id_from_meta_path(bootstrap_path)?;
+                            let blob_id = BlobInfo::get_blob_id_from_meta_path(bootstrap_path);
+                            if let Err(err) = blob_id {
```
imeoer, line 238: Better to use `map_err` ?

src/bin/nydus-image/main.rs
```diff
@@ -758,7 +758,11 @@ fn main() -> Result<()> {
             }
         }
     } else if let Some(matches) = cmd.subcommand_matches("merge") {
-        Command::merge(matches, &build_info)
+        let result = Command::merge(matches, &build_info);
+        if let Err(MergeError::InconsistentFilesystem(_)) = result {
+            std::process::exit(2);
```
imeoer, line 763: Does this will print the error message to stderr before exit?

PerseidMeteor: it seems not,may be use `eprintln`?

imeoer (approved):
LGTM! Thanks!

PR #1425, opened 2023-09-14
storage: fix auth compatibility for registry backend

lihuahua123:
## Details
fix auth compatibility for registry backend:
Due to the different implementations of various image registries, auth requests may use the GET or POST methods.
In the previous way, it just use 'GET' method, and in the pr, we fix it advanced, it can use and cache the GET or POST methods.

## Types of changes

## Checklist

### DIFF
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -93,19 +93,22 @@ impl Cache {
 }
 
 #[derive(Default)]
-struct HashCache(RwLock<HashMap<String, String>>);
+struct HashCache<T>(RwLock<HashMap<String, T>>);
 
-impl HashCache {
+impl<T> HashCache<T> {
     fn new() -> Self {
         HashCache(RwLock::new(HashMap::new()))
     }
 
-    fn get(&self, key: &str) -> Option<String> {
+    fn get(&self, key: &str) -> Option<T>
+    where
+        T: Clone,
+    {
         let cached_guard = self.0.read().unwrap();
         cached_guard.get(key).cloned()
     }
 
-    fn set(&self, key: String, value: String) {
+    fn set(&self, key: String, value: T) {
         let mut cached_guard = self.0.write().unwrap();
         cached_guard.insert(key, value);
     }
@@ -136,6 +139,7 @@ struct BasicAuth {
 }
 
 #[derive(Debug, Clone)]
+#[allow(dead_code)]
 struct BearerAuth {
     realm: String,
     service: String,
@@ -189,10 +193,14 @@ struct RegistryState {
     // Example: RwLock<"Bearer <token>">
     //          RwLock<"Basic base64(<username:password>)">
     cached_auth: Cache,
+    // Cache for the HTTP method when getting auth, it is "true" when using "GET" method.
+    // Due to the different implementations of various image registries, auth requests
+    // may use the GET or POST methods, we need to cache the method after the
+    // fallback, so it can be reused next time and reduce an unnecessary request.
+    cached_auth_using_http_get: HashCache<bool>,
     // Cache 30X redirect url
     // Example: RwLock<HashMap<"<blob_id>", "<redirected_url>">>
-    cached_redirect: HashCache,
-
+    cached_redirect: HashCache<String>,
     // The epoch timestamp of token expiration, which is obtained from the registry server.
     token_expired_at: ArcSwapOption<u64>,
     // Cache bearer auth for refreshing token.
@@ -238,47 +246,34 @@ impl RegistryState {
         }
     }
 
-    /// Request registry authentication server to get bearer token
+    // Request registry authentication server to get bearer token
     fn get_token(&self, auth: BearerAuth, connection: &Arc<Connection>) -> Result<TokenResponse> {
-        // The information needed for getting token needs to be placed both in
-        // the query and in the body to be compatible with different registry
-        // implementations, which have been tested on these platforms:
-        // docker hub, harbor, github ghcr, aliyun acr.
-        let query = [
-            ("service", auth.service.as_str()),
-            ("scope", auth.scope.as_str()),
-            ("grant_type", "password"),
-            ("username", self.username.as_str()),
-            ("password", self.password.as_str()),
-            ("client_id", REGISTRY_CLIENT_ID),
-        ];
+        let http_get = self
+            .cached_auth_using_http_get
+            .get(&self.host)
+            .unwrap_or_default();
+        let resp = if http_get {
+            self.get_token_with_get(&auth, connection)?
+        } else {
+            match self.get_token_with_post(&auth, connection) {
+                Ok(resp) => resp,
+                Err(err) => {
+                    warn!("retry http GET method to get auth token: {}", err);
+                    let resp = self.get_token_with_get(&auth, connection)?;
+                    // Cache http method for next use.
+                    self.cached_auth_using_http_get.set(self.host.clone(), true);
+                    resp
+                }
+            }
+        };
 
-        let mut form = HashMap::new();
-        for (k, v) in &query {
-            form.insert(k.to_string(), v.to_string());
-        }
-
-        let mut headers = HeaderMap::new();
-        if let Some(auth_header) = &auth.header {
-            headers.insert(HEADER_AUTHORIZATION, auth_header.clone());
-        }
-
-        let token_resp = connection
-            .call::<&[u8]>(
-                Method::GET,
-                auth.realm.as_str(),
-                Some(&query),
-                Some(ReqBody::Form(form)),
-                &mut headers,
-                true,
-            )
-            .map_err(|e| einval!(format!("registry auth server request failed {:?}", e)))?;
-        let ret: TokenResponse = token_resp.json().map_err(|e| {
+        let ret: TokenResponse = resp.json().map_err(|e| {
             einval!(format!(
                 "registry auth server response decode failed: {:?}",
                 e
             ))
         })?;
+
         if let Ok(now_timestamp) = SystemTime::now().duration_since(UNIX_EPOCH) {
             self.token_expired_at
                 .store(Some(Arc::new(now_timestamp.as_secs() + ret.expires_in)));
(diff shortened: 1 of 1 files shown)


imeoer:
Hi @lihuahua123 please add the detailed commit message, and the commit title better be like `storage: fix auth compatibility for registry backend`.

storage/src/backend/registry.rs
```diff
@@ -188,6 +189,10 @@ struct RegistryState {
     // Use RwLock here to avoid using mut backend trait object.
     // Example: RwLock<"Bearer <token>">
     //          RwLock<"Basic base64(<username:password>)">
+
+    // Cache the method of getting auth, get | post
+    cached_get_auth_method: HashCache,
```
imeoer, line 194: How about:

``` rust
// Cache for the HTTP method when getting auth, possible values: "get", "post".
// Due to the different implementations of various image registries, auth requests
// may use the GET or POST methods, we need to cache the method after the
// fallback, so it can be reused next time and reduce an unnecessary request.
cached_auth_using_http_get: HashCache,
```

And can we make it as `HashCache<bool>` ?

```
struct HashCache(RwLock<HashMap<String, T>>);
```

storage/src/backend/registry.rs
```diff
@@ -136,6 +136,7 @@ struct BasicAuth {
 }
 
 #[derive(Debug, Clone)]
+#[allow(dead_code)]
```
imeoer, line 139: Why mark `dead_code` here.

storage/src/backend/registry.rs
```diff
@@ -259,39 +307,49 @@ impl RegistryState {
...
+    }
 
-        Ok(ret)
+    // Get bearer token using a GET request
+    fn get_token_with_get(
+        &self,
+        auth: &BearerAuth,
+        connection: &Arc<Connection>,
+    ) -> Result<Response> {
+        let query = [
```
imeoer, line 331: Can we wrap this as a `fn`? So that we can reuse it in `get_token_with_post`.

storage/src/backend/registry.rs
lihuahua123: In the context ,HashCache is implemented as 
`struct HashCache(RwLock<HashMap<String, String>>);`
if  we make it as' HashCache<bool> ', it need implement another logic

imeoer: I mean the generic type https://doc.rust-lang.org/rust-by-example/generics.html

storage/src/backend/registry.rs
```diff
@@ -253,45 +331,20 @@ impl RegistryState {
...
         ];
 
-        let mut form = HashMap::new();
-        for (k, v) in &query {
-            form.insert(k.to_string(), v.to_string());
-        }
-
         let mut headers = HeaderMap::new();
-        if let Some(auth_header) = &auth.header {
-            headers.insert(HEADER_AUTHORIZATION, auth_header.clone());
```
imeoer, line 263: Why remove this? Please take a look the commit [069818c](https://github.com/dragonflyoss/image-service/commit/069818c6958ad2c9cfe84ecf361b3993fdc2c795#diff-f7ce8f265a570c66eae48c85e0f5b6f29fdaec9cf2ee2eded95810fe320d80e1R346) .

lihuahua123: When we use `get` method, it just need to add paramaters into request headers, and when use `post` method, it just need to add paramaters into request form. This can  refer to contained as following.
```
// FetchTokenWithOAuth fetches a token using a POST request
func FetchTokenWithOAuth(ctx context.Context, client *http.Client, headers http.Header, clientID string, to TokenOptions) (*OAuthTokenResponse, error) {
	form := url.Values{}
	if len(to.Scopes) > 0 {
		form.Set("scope", strings.Join(to.Scopes, " "))
	}
	form.Set("service", to.Service)
	form.Set("client_id", clientID)

	if to.Username == "" {
		form.Set("grant_type", "refresh_token")
		form.Set("refresh_token", to.Secret)
	} else {
		form.Set("grant_type", "password")
		form.Set("username", to.Username)
		form.Set("password", to.Secret)
	}
	if to.FetchRefreshToken {
		form.Set("access_type", "offline")
	}

	req, err := http.NewRequestWithContext(ctx, http.MethodPost, to.Realm, strings.NewReader(form.Encode()))
	if err != nil {
		return nil, err
	}
	req.Header.Set("Content-Type", "application/x-www-form-urlencoded; charset=utf-8")
	for k, v := range headers {
		req.Header[k] = append(req.Header[k], v...)
	}
	if len(req.Header.Get("User-Agent")) == 0 {
		req.Header.Set("User-Agent", "containerd/"+version.Version)
	}

	resp, err := client.Do(req)
	if err != nil {
		return nil, err
	}
	defer resp.Body.Close()

	if resp.StatusCode < 200 || resp.StatusCode >= 400 {
		return nil, remoteserrors.NewUnexpectedStatusErr(resp)
	}

	decoder := json.NewDecoder(resp.Body)

	var tr OAuthTokenResponse
	if err = decoder.Decode(&tr); err != nil {
		return nil, fmt.Errorf("unable to decode token response: %w", err)
	}

	if tr.AccessToken == "" {
		return nil, ErrNoToken
	}

	return &tr, nil
}




// FetchToken fetches a token using a GET request
func FetchToken(ctx context.Context, client *http.Client, headers http.Header, to TokenOptions) (*FetchTokenResponse, error) {
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, to.Realm, nil)
	if err != nil {
		return nil, err
	}

	for k, v := range headers {
		req.Header[k] = append(req.Header[k], v...)
	}
	if len(req.Header.Get("User-Agent")) == 0 {
		req.Header.Set("User-Agent", "containerd/"+version.Version)
	}

	reqParams := req.URL.Query()

	if to.Service != "" {
		reqParams.Add("service", to.Service)
	}

	for _, scope := range to.Scopes {
		reqParams.Add("scope", scope)
	}

	if to.Secret != "" {
		req.SetBasicAuth(to.Username, to.Secret)
	}

	if to.FetchRefreshToken {
		reqParams.Add("offline_token", "true")
	}

	req.URL.RawQuery = reqParams.Encode()

	resp, err := client.Do(req)
	if err != nil {
		return nil, err
	}
	defer resp.Body.Close()

	if resp.StatusCode < 200 || resp.StatusCode >= 400 {
		return nil, remoteserrors.NewUnexpectedStatusErr(resp)
	}

	decoder := json.NewDecoder(resp.Body)

	var tr FetchTokenResponse
	if err = decoder.Decode(&tr); err != nil {
		return nil, fmt.Errorf("unable to decode token response: %w", err)
	}

	// `access_token` is equivalent to `token` and if both are specified
	// the choice is undefined.  Canonicalize `access_token` by sticking
	// things in `token`.
	if tr.AccessToken != "" {
		tr.Token = tr.AccessToken
	}

	if tr.Token == "" {
		return nil, ErrNoToken
	}

	return &tr, nil
}


```

storage/src/backend/registry.rs
lihuahua123: I have resolve it with changing the implemention of `struct HashCache(RwLock<HashMap<String, String>>);` to `struct HashCache<T>(RwLock<HashMap<String, T>>);`.

storage/src/backend/registry.rs
```diff
@@ -189,10 +193,14 @@ struct RegistryState {
     // Example: RwLock<"Bearer <token>">
     //          RwLock<"Basic base64(<username:password>)">
     cached_auth: Cache,
+    // Cache for the HTTP method when getting auth, possible values: "get", "post".
```
imeoer, line 196: Should also update the comment.

storage/src/backend/registry.rs
```diff
@@ -238,12 +246,86 @@ impl RegistryState {
...
 
-    /// Request registry authentication server to get bearer token
+    // Request registry authentication server to get bearer token
     fn get_token(&self, auth: BearerAuth, connection: &Arc<Connection>) -> Result<TokenResponse> {
-        // The information needed for getting token needs to be placed both in
-        // the query and in the body to be compatible with different registry
-        // implementations, which have been tested on these platforms:
-        // docker hub, harbor, github ghcr, aliyun acr.
+        let mut resp_token: Response;
+        if let Some(method) = self.cached_auth_using_http_get.get(&self.host) {
```
imeoer, line 252: `method` -> `using_get` ?

storage/src/backend/registry.rs
imeoer: Ok, let's move forward, thanks.

storage/src/backend/registry.rs
```diff
@@ -238,12 +246,86 @@ impl RegistryState {
...
+            if resp_token.status().is_success() {
+                self.cached_auth_using_http_get
+                    .set(self.host.clone(), false);
+            } else {
+                resp_token = self.get_token_with_get(&auth, connection)?;
+                self.cached_auth_using_http_get
+                    // .set(self.host.clone(), "get".to_string());
+                    .set(self.host.clone(), true);
+            }
+        }
```
imeoer, line 269: How about?

``` rust
        let http_get = self
            .cached_auth_using_http_get
            .get(&self.host)
            .unwrap_or_default();
        let resp = if http_get {
            self.get_token_with_get(&auth, connection)?
        } else {
            match self.get_token_with_post(&auth, connection) {
                Ok(resp) => resp,
                Err(err) => {
                    warn!("retry http GET method to get auth token: {}", err);
                    let resp = self.get_token_with_get(&auth, connection)?;
                    // Cache http method for next use.
                    self.cached_auth_using_http_get.set(self.host.clone(), true);
                    resp
                }
            }
        };
```

imeoer (approved):
LGTM, thanks!

storage/src/backend/registry.rs
lihuahua123: I have changed it .

storage/src/backend/registry.rs
lihuahua123: Ok！

Issue #1415, opened 2023-08-30
add a new args for nydus-image use to generate aligned blobcache file


zyfjeff:
## Details about issue

At present, the nydus blob we generate is not directly used by nydusd, but first downloaded, decompressed and finally written to the blobcache file, if we can directly generate the blobcache file when the image is built, nydusd directly uses this file to start. The process of writing data from Nydus blob copy data to blobcache files is avoided. This is an optimization for localfs mode.

PR #1433, opened 2023-09-25
Add --blob-cache-dir arg use to generate raw blob cache and meta

zyfjeff:
#1415 

## Details
generate blob cache and blob meta through the —-blob-cache-dir parameters,
so that nydusd can be started directly from these two files without
going to the backend to download. this can improve the performance
of data loading in localfs mode.

## Types of changes

## Checklist

### DIFF
--- a/builder/src/core/blob.rs
+++ b/builder/src/core/blob.rs
@@ -3,7 +3,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::borrow::Cow;
-use std::io::Write;
 use std::slice;
 
 use anyhow::{Context, Result};
@@ -16,9 +15,8 @@ use sha2::digest::Digest;
 
 use super::layout::BlobLayout;
 use super::node::Node;
-use crate::{
-    ArtifactWriter, BlobContext, BlobManager, BuildContext, ConversionType, Feature, Tree,
-};
+use crate::core::context::Artifact;
+use crate::{BlobContext, BlobManager, BuildContext, ConversionType, Feature, Tree};
 
 /// Generator for RAFS data blob.
 pub(crate) struct Blob {}
@@ -29,7 +27,7 @@ impl Blob {
         ctx: &BuildContext,
         tree: &Tree,
         blob_mgr: &mut BlobManager,
-        blob_writer: &mut ArtifactWriter,
+        blob_writer: &mut dyn Artifact,
     ) -> Result<()> {
         match ctx.conversion_type {
             ConversionType::DirectoryToRafs => {
@@ -101,7 +99,7 @@ impl Blob {
     fn finalize_blob_data(
         ctx: &BuildContext,
         blob_mgr: &mut BlobManager,
-        blob_writer: &mut ArtifactWriter,
+        blob_writer: &mut dyn Artifact,
     ) -> Result<()> {
         // Dump buffered batch chunk data if exists.
         if let Some(ref batch) = ctx.blob_batch_generator {
@@ -159,7 +157,7 @@ impl Blob {
     pub(crate) fn dump_meta_data(
         ctx: &BuildContext,
         blob_ctx: &mut BlobContext,
-        blob_writer: &mut ArtifactWriter,
+        blob_writer: &mut dyn Artifact,
     ) -> Result<()> {
         // Dump blob meta for v6 when it has chunks or bootstrap is to be inlined.
         if !blob_ctx.blob_meta_info_enabled || blob_ctx.uncompressed_blob_size == 0 {
@@ -194,7 +192,6 @@ impl Blob {
         } else if ctx.blob_tar_reader.is_some() {
             header.set_separate_blob(true);
         };
-
         let mut compressor = Self::get_compression_algorithm_for_meta(ctx);
         let (compressed_data, compressed) = compress::compress(ci_data, compressor)
             .with_context(|| "failed to compress blob chunk info array".to_string())?;
@@ -223,6 +220,9 @@ impl Blob {
         }
 
         blob_ctx.blob_meta_header = header;
+        if let Some(blob_cache) = ctx.blob_cache_generator.as_ref() {
+            blob_cache.write_blob_meta(ci_data, &header)?;
+        }
         let encrypted_header =
             crypt::encrypt_with_context(header.as_bytes(), cipher_obj, cipher_ctx, encrypt)?;
         let header_size = encrypted_header.len();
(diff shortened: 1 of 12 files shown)


src/bin/nydus-image/main.rs
```diff
@@ -1479,6 +1498,52 @@ impl Command {
...
+    ) -> Result<Option<Mutex<ArtifactFileWriter>>> {
+        if conversion_type == ConversionType::EStargzIndexToRef {
+            Ok(None)
+        } else if let Some(p) = matches
+            .get_one::<PathBuf>("blob-meta")
+            .map(|b| ArtifactStorage::SingleFile(b.clone()))
+        {
+            if conversion_type == ConversionType::TarToTarfs {
+                bail!(
+                    "conversion type `{}` conflicts with `--blob-meta`",
```
imeoer, line 1513: What other conversion types don't support the option? Should also treat `ConversionType::EStargzIndexToRef` like this?

zyfjeff: Yes, the types that cannot get the original file content at build time are not supported, TarToTarfs, EStargzIndexToRef, etc

builder/src/core/blob.rs
```diff
@@ -223,6 +223,17 @@ impl Blob {
         }
 
         blob_ctx.blob_meta_header = header;
+        if let Some(blob_cache) = ctx.blob_cache_generator.as_ref() {
+            let mut guard = blob_cache.lock().unwrap();
+            let meta_write = guard.get_blob_meta_writer_mut();
+            let aligned_uncompressed_size = try_round_up_4k(uncompressed_size as u64).unwrap();
```
jiangliu, line 229: Please return error instead of `unwrap()` when possible.

src/bin/nydus-image/main.rs
```diff
@@ -1041,6 +1050,12 @@ impl Command {
         build_ctx.set_chunk_size(chunk_size);
         build_ctx.set_batch_size(batch_size);
 
+        let blob_cache_generator = match blob_cache_storage {
+            Some(storage) => Some(Mutex::new(BlobCacheGenerator::new(storage)?)),
+            None => None,
+        };
+        build_ctx.blob_cache_generator = blob_cache_generator;
```
jiangliu, line 1057:
```
build_ctx.blob_cache_generator = blob_cache_storage.map(|v| {Mutex::new(BlobCacheGeneratorNew(v))});
```

src/bin/nydus-image/main.rs
```diff
@@ -1479,6 +1494,31 @@ impl Command {
         }
     }
 
+    fn get_blob_cache_storage(
+        matches: &ArgMatches,
+        conversion_type: ConversionType,
+    ) -> Result<Option<ArtifactStorage>> {
+        if let Some(p) = matches.get_one::<PathBuf>("blob-cache-dir") {
+            if conversion_type == ConversionType::TarToTarfs
```
jiangliu, line 1502: How about using whitelist instead of blacklist here?

builder/src/lib.rs
```diff
@@ -237,6 +237,11 @@ fn finalize_blob(
         // blob file.
         if !is_tarfs {
             blob_writer.finalize(Some(blob_meta_id))?;
+            if let Some(blob_cache) = ctx.blob_cache_generator.as_ref() {
```
jiangliu, line 240: How about moving the code piece out of `!is_tarfs`?
We should finalize the storage if ctx.blob_cache_generator is valid.

builder/src/core/blob.rs
```diff
@@ -223,6 +223,17 @@ impl Blob {
         }
 
         blob_ctx.blob_meta_header = header;
+        if let Some(blob_cache) = ctx.blob_cache_generator.as_ref() {
```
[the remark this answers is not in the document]
zyfjeff: Ok, I will put these code into the blob_cache_generator

src/bin/nydus-image/main.rs
zyfjeff: BlobCacheGenerator::new(storage) returns a Result, it can't return the error directly inside the closure

src/bin/nydus-image/main.rs
zyfjeff: There are more whitelists than Blacklist, we support xxxToRef and xxxToRafs

smoke/tests/blobcache_test.go
```diff
@@ -0,0 +1,120 @@
...
+	require.NoError(t, err)
+
+	a.compareTwoFiles(t, filepath.Join(blobcacheDir,fmt.Sprintf("%s.blob.data", ociBlobDigest.Hex())), filepath.Join(ctx.Env.CacheDir, fmt.Sprintf("%s.blob.data", ociBlobDigest.Hex())))
+	a.compareTwoFiles(t, filepath.Join(blobcacheDir, fmt.Sprintf("%s.blob.meta", ociBlobDigest.Hex())), filepath.Join(ctx.Env.CacheDir, fmt.Sprintf("%s.blob.meta", ociBlobDigest.Hex())))
+}
+
+
+func TestBlobCache(t *testing.T) {
+	test.Run(t, &BlobCacheTestSuite{T: t})
+}
```
imeoer, line 120: Should add a blank line here.

src/bin/nydus-image/main.rs
```diff
@@ -356,6 +357,13 @@ fn prepare_cmd_args(bti_string: &'static str) -> App {
                         .action(ArgAction::SetTrue)
                         .required(false)
                 )
+                .arg(
+                    Arg::new("blob-cache-dir")
+                        .long("blob-cache-dir")
+                        .help("Directory path to save generated blob cache files(blob meta and blob data)")
```
imeoer, line 363: How about `Directory path to generate blob cache files ($id.blob.meta and $id.blob.data)`.

builder/src/core/context.rs
```diff
@@ -193,7 +194,13 @@ impl Write for ArtifactMemoryWriter {
     }
 }
 
-struct ArtifactFileWriter(ArtifactWriter);
+pub struct ArtifactFileWriter(pub ArtifactWriter);
```
jiangliu, line 197: Is this change unnecessary now？

src/bin/nydus-image/main.rs
```diff
@@ -867,8 +888,8 @@ impl Command {
                 }
                 compressor = compress::Algorithm::GZip;
                 digester = digest::Algorithm::Sha256;
-                if blob_storage.is_none() {
-                    bail!("both --blob and --blob-dir are missing");
+                if blob_storage.is_none() && blob_cache_storage.is_none() {
+                    bail!("both --blob and --blob-dir or --blob-cache-dir are missing");
```
jiangliu, line 892: `all of --blob, --blob-dir and --blob-cache-dir are missing`?

imeoer (approved):

jiangliu (approved):

zyfjeff:
nydus-unit-test-coverage failed due to network issues, and it had nothing to do with my changes.

Issue #1421, opened 2023-09-11
nydusd failover and live upgrade in fuse dev


imeoer:
- Failover: in the event of a nydusd crash or other failures, the nydusd process can be restarted from a snapshot. This will ensure that the container filesystem served by nydusd is still available.

- Live upgrade: nydusd can be upgraded without disrupting the container filesystem it serves. This is done by creating a snapshot of the running nydusd process, upgrading the binary, and then restoring the snapshot.

To achieve these capabilities, nydusd needs to store the following information in a persistent store:

- The fuse file descriptor for the fuse connection;
- The fuse state in [fuse-backend-rs](https://github.com/cloud-hypervisor/fuse-backend-rs);

The [nydus-snapshotter](https://github.com/containerd/nydus-snapshotter) will be responsible for the workflow.

imeoer:
cc @loheagn

loheagn:
https://github.com/dragonflyoss/nydus/pull/1487 and https://github.com/dragonflyoss/nydus/pull/1529 may fix this issue.

PR #1434, opened 2023-09-26
feat: implement `takeover` for nydusd fusedev daemon

loheagn:

#1421 

## Details

This patch implements the `save` and `restore` functions in the `fusedev_upgrade` in the service crate. 

To do this,
- This patch adds a new crate named `nydus-upgrade` into the workspace. The `nydus-upgrade` crate has some util functions help to do serialization and deserialization for the rust structs using the [versionize](https://github.com/firecracker-microvm/versionize) and [snapshot](https://github.com/firecracker-microvm/firecracker/tree/main/src/snapshot) crates. The crate also has a trait named `StorageBackend` which can be used to store and restore fuse session fds and state data for the upgrade action, and there's also an implementation named `UdsStorageBackend` which uses unix domain socket to do this.
- as we have to use the same fuse session connection, backend file system mount commands, Vfs to re-mount the rafs for the new daemon (created for "hot upgrade" or failover), this patch add a new struct named `FusedevState` to hold these information. The `FusedevState` is serialized and stored into the `UdsStorageBackend` (which happens in the `save` function in the `fusedev_upgrade` module) before the new daemon is created, and the `FusedevState` is deserialized and restored from the `UdsStorageBackend` (which happens in the `restore` function in the `fusedev_upgrade` module) when the new daemon is triggered by `takeover`.

## Types of changes

## Checklist

### DIFF
--- a/service/src/upgrade.rs
+++ b/service/src/upgrade.rs
@@ -4,15 +4,40 @@
 
 //! Online upgrade manager for Nydus daemons and filesystems.
 
+use std::any::TypeId;
+use std::collections::HashMap;
 use std::convert::{TryFrom, TryInto};
+use std::io;
+use std::os::fd::RawFd;
 use std::path::PathBuf;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+use nydus_upgrade::backend::unix_domain_socket::UdsStorageBackend;
+use nydus_upgrade::backend::{StorageBackend, StorageBackendErr};
 
 use crate::fs_service::{FsBackendMountCmd, FsBackendUmountCmd};
-use crate::Result;
+use crate::{Error, Result};
 
 /// Error codes related to upgrade manager.
 #[derive(thiserror::Error, Debug)]
-pub enum UpgradeMgrError {}
+pub enum UpgradeMgrError {
+    #[error("missing supervisor path")]
+    MissingSupervisorPath,
+
+    #[error("failed to save/restore data via the backend, {0}")]
+    StorageBackendError(StorageBackendErr),
+
+    #[error("failed to serialize, {0}")]
+    Serialize(io::Error),
+    #[error("failed to deserialize, {0}")]
+    Deserialize(io::Error),
+}
+
+impl From<UpgradeMgrError> for Error {
+    fn from(e: UpgradeMgrError) -> Self {
+        Error::UpgradeManager(e)
+    }
+}
 
 /// FUSE fail-over policies.
 #[derive(PartialEq, Eq)]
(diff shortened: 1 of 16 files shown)


loheagn:
As this patch depend on the [PR](https://github.com/cloud-hypervisor/fuse-backend-rs/pull/149),    this pr may keep draft until the [PR](https://github.com/cloud-hypervisor/fuse-backend-rs/pull/149) merged.

kevinXYin:
Thanks for the nice job! Can this work with the nydus-snapshotter ? I also have plans to support takeover feature for fscache mode,  I think there are some process that can be reused.

service/src/upgrade.rs
```diff
@@ -43,46 +68,255 @@ impl TryFrom<&String> for FailoverPolicy {
...
+        // set fd
+        let fd = svc
+            .as_ref()
+            .as_any()
+            .downcast_ref::<FusedevFsService>()
+            .unwrap()
+            .session
+            .lock()
+            .unwrap()
+            .get_fuse_file()
```
kevinXYin, line 226: If nydusd exit when there are inflight fuse requests, IO hung occurs because the mainline kernel does not support resending inflight fuse requests. If we want to support takeover for fusedev, how about store the fuse fd created by ioctl FUSE_DEV_IOC_CLONE, then when nydusd exit, the inflight requests will be dropped instead of blocking IO.

loheagn:
> Thanks for the nice job! Can this work with the nydus-snapshotter ? I also have plans to support takeover feature for fscache mode, I think there are some process that can be reused.

Here is a pr for nydus-snapshotter to do failover and hot upgrade. This pr adds some check code to ensure the upgrade process be done for all the old version nydus daemons, especially the deamons created during the upgrade stage.

I have tested the failover feature and more test cases should be added.

service/src/upgrade.rs
kevinXYin: I'm not sure if macos can support clone,  maybe we need to keep the API more generic. 
Any suggestions? @imeoer @jiangliu

kevinXYin:
> > Thanks for the nice job! Can this work with the nydus-snapshotter ? I also have plans to support takeover feature for fscache mode, I think there are some process that can be reused.
> 
> Here is a pr [containerd/nydus-snapshotter#520](https://github.com/containerd/nydus-snapshotter/pull/520) for nydus-snapshotter to do failover and hot upgrade. This pr adds some check code to ensure the upgrade process be done for all the old version nydus daemons, especially the deamons created during the upgrade stage.
> 
> I have tested the failover feature and more test cases should be added.

As I know , the nydus-snapshotter already support failover and hot upgrade workflow.

service/src/upgrade.rs
imeoer: Maybe make it as a built-in feature in `fuse-backend-rs`? WDYT cc @jiangliu ?

kevinXYin:
I will send another pr also support takeover for fscache , which fold your patch , then we can start reviewing full takeover feature.

service/src/upgrade.rs
jiangliu: Sure, let's get fuse-backend-rs ready for this:)

loheagn: @imeoer @jiangliu 

> Sure, let's get fuse-backend-rs ready for this:)

@imeoer @jiangliu I might be able to help with this :)

loheagn: https://github.com/cloud-hypervisor/fuse-backend-rs/pull/177 may fix this

bergwolf:
@kevinXYin The feature has been pending for too long. Let's do it step by step, by merging the fuse takeover first, and then you can proceed with fscache takeover.

PR #1475, opened 2023-11-07
Add some unit test for utils, storage, rafs, api and builder.

ZouTaooo:

## Details

## Types of changes

## Checklist

### DIFF
--- a/storage/Cargo.toml
+++ b/storage/Cargo.toml
@@ -35,6 +35,7 @@ url = { version = "2.1.1", optional = true }
 vm-memory = "0.10"
 fuse-backend-rs = "^0.10.3"
 gpt = { version = "3.0.0", optional = true }
+toml = "0.5"
 
 nydus-api = { version = "0.3", path = "../api" }
 nydus-utils = { version = "0.4", path = "../utils", features = ["encryption", "zran"] }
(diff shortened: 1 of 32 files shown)


imeoer:
Great work, thanks! @ZouTaooo 
cc @jiangliu @adamqqqplay Let's take a look together.

adamqqqplay:
@ZouTaooo Thanks for your contribution! It looks really good.

storage/Cargo.toml
```diff
@@ -35,6 +35,7 @@ url = { version = "2.1.1", optional = true }
 vm-memory = "0.10"
 fuse-backend-rs = "^0.10.3"
 gpt = { version = "3.0.0", optional = true }
+toml = "0.5"
```
imeoer, line 38: Maybe move it to `[dev-dependencies]`.

ZouTaooo: Thanks for your review. Now, I have fixed it with new commit.

ZouTaooo (commented):
The Smoke-Test found a test-case error. And the commit fix it.
Two seconds is an edge-case, but there is a resolution problem in linux timer. So the data will be updated with different index sometimes. So the new test case with two changes, first using the normal argument rather than edge case, secondly considering the timer delta error.

jiangliu (approved):

PR #1491, opened 2023-11-22
go work: save go.work.sum locally

Desiki-high:
## Details
`go.work.sum` always changes too large. We only need it to work well locally. See https://github.com/dragonflyoss/nydus/pull/1490#discussion_r1400376065.

## Types of changes

## Checklist


smoke/go.sum
```diff
@@ -1,17 +1,23 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
imeoer, line 1: Does other */go.sum need also be updated?

Desiki-high: I have checked other `*/go.sum`, and only found it in smoke module.

imeoer (approved):
Thanks!

PR #1499, opened 2023-11-30
docs: nydus-sandbox.yaml add uid

### LINKED ISSUE #1496: The sandbox.yaml reference in the containerd-env-setup.md document has issues
Related issues: https://github.com/kubernetes-sigs/cri-tools/issues/1273

The `nydus-sandbox.yaml` does not specify a uid which causes the `crictl rmp` to fail
```
metadata:
  attempt: 1
  name: nydus-sandbox
  namespace: default
log_directory: /tmp
linux:
  security_context:
    namespace_options:
      network: 2
```

As a test example, uid can be specified as an arbitrary string, like: 
```
metadata:
  attempt: 1
  name: nydus-sandbox
  namespace: default
  uid: nydus-sandbox-test
```

This issue will be fixed in the new version, but it remains a problem for those using slightly earlier versions of crictl.

whygyc:
Fix #1496 

## Details
To avoid the inability to delete pods created by crictl, add the UID to the example yaml.

## Types of changes

### DIFF
--- a/docs/containerd-env-setup.md
+++ b/docs/containerd-env-setup.md
@@ -189,6 +189,7 @@ metadata:
   attempt: 1
   name: nydus-sandbox
   namespace: default
+  uid: nydus-sandbox-test
 log_directory: /tmp
 linux:
   security_context:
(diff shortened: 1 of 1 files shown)


adam3q (approved):
Thanks for your help!

PR #1561, opened 2024-03-22
Morgan/upstream fix

ccx1024cc:
## Types of changes

1. fix residual fuse mountpoint after graceful shutdown
2. fix compatibility to image without ext table for blob cache
3. fix not reuse blob cache of old version
4. fix: bad read by wrong data region


adamqqqplay:
Unit test failed, PTAL.

ccx1024cc:
> Unit test failed, PTAL.

This pr is in no contact with block device. Please help seek for problems of ut. @jiangliu

PR #1595, opened 2024-07-03
feat: skip proxy health check if connection is not active

### LINKED ISSUE #1588: should close health check connection after image data request done
The nydusd healthy check connection is 90s idle timeout (https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.pool_idle_timeout), as the number of images increases in k8s cluster, the dragonfly server load will be stressful.

BruceAko:

Fixes #1588 

## Details

1. Add `last_active` field for `Connection`. When `Connection.call()` is called, `last_active`  is updated to current timestamp.
2. Add `check_pause_elapsed` field for `ProxyConfig` and `MirrorConfig`. Connection is considered to be inactive if the current time to the `last_active` time exceeds `check_pause_elapsed`.
3. In proxy and mirror's health checking thread's loop, if the connection is not active (exceeds `check_pause_elapsed`), this round of health check is skipped.
4. Update the document.

## Types of changes

## Checklist

### DIFF
--- a/api/src/config.rs
+++ b/api/src/config.rs
@@ -907,6 +907,9 @@ pub struct ProxyConfig {
     /// Replace URL to http to request source registry with proxy, and allow fallback to https if the proxy is unhealthy.
     #[serde(default)]
     pub use_http: bool,
+    /// Elapsed time to pause proxy health check when the request is inactive, in seconds.
+    #[serde(default = "default_check_pause_elapsed")]
+    pub check_pause_elapsed: u64,
 }
 
 impl Default for ProxyConfig {
@@ -917,6 +920,7 @@ impl Default for ProxyConfig {
             fallback: true,
             check_interval: 5,
             use_http: false,
+            check_pause_elapsed: 300,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


imeoer:
Also need to handle the `start_mirrors_health_thread` func.

imeoer:
Please also update the [doc](https://github.com/dragonflyoss/nydus/blob/master/docs/nydusd.md) for these sections:

- Enable Mirrors for Storage Backend
- Enable P2P Proxy for Storage Backend

BruceAko:
The above has been revised, PTAL, thanks. @imeoer

api/src/config.rs
```diff
@@ -938,6 +942,9 @@ pub struct MirrorConfig {
     /// Maximum number of failures before marking a mirror as unusable.
     #[serde(default = "default_failure_limit")]
     pub failure_limit: u8,
+    /// Elapsed time to pause mirror health check when the request is inactive, in seconds.
+    #[serde(default = "default_check_pause_elapsed")]
+    pub check_pause_elapsed: u64,
```
imeoer, line 947: We have `health_check_interval`, so maybe name it `health_check_pause_elapsed` ?

imeoer (approved):
others LGTM, thanks!

api/src/config.rs
BruceAko: Ok

BruceAko: Fixed, PTAL :) @imeoer

imeoer (approved):

PR #1597, opened 2024-07-06
fix: add namespace flag for nydusify commit.

PerseidMeteor:
https://github.com/dragonflyoss/nydus/issues/1596

## Details
Add `namespace` flag for `nydusify commit`.

## Types of changes

## Checklist

### DIFF
--- a/contrib/nydusify/cmd/nydusify.go
+++ b/contrib/nydusify/cmd/nydusify.go
@@ -1180,6 +1180,13 @@ func main() {
 					EnvVars: []string{"CONTAINERD_ADDR"},
 				},
 				&cli.StringFlag{
+					Name:    "namespace",
+					Aliases: []string{"n"},
+					Value:   "default",
+					Usage:   "Container namespace, default with \"default\" namespace",
+					EnvVars: []string{"NAMESPACE"},
+				},
+				&cli.StringFlag{
 					Name:     "container",
 					Required: true,
 					Usage:    "Target container id",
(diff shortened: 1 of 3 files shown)


PerseidMeteor:
PTAL :) @imeoer

contrib/nydusify/cmd/nydusify.go
```diff
@@ -1179,6 +1179,12 @@ func main() {
 					Usage:   "Containerd address, optionally with \"unix://\" prefix [$CONTAINERD_ADDRESS] (default \"/run/containerd/containerd.sock\")",
 					EnvVars: []string{"CONTAINERD_ADDR"},
 				},
+				&cli.StringFlag{
+					Name:    "namespace",
```
imeoer, line 1183: Maybe also add an aliases `n`. :)

contrib/nydusify/pkg/committer/commiter.go
```diff
@@ -84,7 +85,7 @@ func (cm *Committer) Commit(ctx context.Context, opt Opt) error {
 		return errors.Wrap(err, "parse target image name")
 	}
 
-	inspect, err := cm.manager.Inspect(ctx, opt.ContainerID)
+	inspect, err := cm.manager.Inspect(ctx, opt.ContainerID, opt.Namespace)
```
imeoer, line 88: How about calling `ctx = namespaces.WithNamespace(ctx, namespace)` before here.

contrib/nydusify/cmd/nydusify.go
PerseidMeteor: Add `aliases`.

contrib/nydusify/pkg/committer/commiter.go
PerseidMeteor: How about calling in `nydusify.go` `main` func?
Just in `return cm.Commit(c.Context, opt)`

imeoer: Better to put it in `committer.go`, so we can import it as a package in other  project. :)

imeoer (approved):
thanks, LGTM!

shusley244:
thanks~

PR #1606, opened 2024-07-30
nydus-image: refactor unpack/compact cli interface

### LINKED ISSUE #1602: Weird cli interface of nydus-image unpack (and also compact)
The following code snippet, when determining the location of the blob file, will first look for the blob file based on the locations specified by `--blob-dir` and `--blob`, and finally attempt to use `--backend-config`. 

https://github.com/dragonflyoss/nydus/blob/36844742543f69ec5ef32e1f416d38a4fed680db/src/bin/nydus-image/main.rs#L1479-L1514

However, the parameters accepted by `--backend-config` are still the configuration files for nydusd, from which the backend configuration is then obtained. This logic seems strange and redundant, because the --config parameter is also used to read the nydusd configuration file.

https://github.com/dragonflyoss/nydus/blob/36844742543f69ec5ef32e1f416d38a4fed680db/src/bin/nydus-image/main.rs#L1778-L1790

According to the following code snippet it seems that `--backend-config` is deprecated and only maintained for backward compatibility.

https://github.com/dragonflyoss/nydus/blob/36844742543f69ec5ef32e1f416d38a4fed680db/src/bin/nydus-image/main.rs#L1702-L1711

I doubt whether it is necessary for us to introduce the `--backend-config` option in 081c8f968bcb9304c7a4caaa5ce038e54434cd69, especially considering that nydus-image only interacts with local files. Could we just remove it to keep it consistent with the other subcommand in nydus-image?

SToPire:
Since unpack and compact subcommands does not need the entire nydusd configuration file, let's refactor their cli interface and directly take backend configuration file.

Specifically, we introduce `--backend-type`, `--backend-config` and `--backend-config-file` options to specify the backend type and remove `--config` option.

Fixes: #1602

#1602 

## Details
Introduce  `--backend-type`, `--backend-config` and `--backend-config-file` options for `unpack` and `compact` subcommands in nydus-image.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

### DIFF
--- a/smoke/tests/main_test.go
+++ b/smoke/tests/main_test.go
@@ -22,10 +22,16 @@ func TestMain(m *testing.M) {
 		registryPort = "5077"
 		os.Setenv("REGISTRY_PORT", registryPort)
 	}
+
+	var reg *tool.Registry = nil
 	if os.Getenv("DISABLE_REGISTRY") == "" {
-		reg := tool.NewRegistry()
-		defer reg.Destroy()
+		reg = tool.NewRegistry()
 	}
+
 	code := m.Run()
+
+	if reg != nil {
+		reg.Destroy()
+	}
 	os.Exit(code)
 }
--- a/smoke/tests/main_test.go
+++ b/smoke/tests/main_test.go
@@ -23,7 +23,7 @@ func TestMain(m *testing.M) {
 		os.Setenv("REGISTRY_PORT", registryPort)
 	}
 
-	var reg *tool.Registry = nil
+	var reg *tool.Registry
 	if os.Getenv("DISABLE_REGISTRY") == "" {
 		reg = tool.NewRegistry()
 	}
--- a/src/bin/nydus-image/main.rs
+++ b/src/bin/nydus-image/main.rs
@@ -671,17 +671,48 @@ fn prepare_cmd_args(bti_string: &'static str) -> App {
                         .required(true),
                 )
                 .arg(
-                    Arg::new("config")
-                        .long("config")
-                        .short('C')
-                        .help("config to compactor")
-                        .required(true),
+                    Arg::new("backend-type")
+                        .long("backend-type")
+                        .help(format!(
+                            "Type of backend [possible values: {}]",
+                            BlobFactory::supported_backends()
+                                .into_iter()
+                                .filter(|x| x != "localfs")
+                                .collect::<Vec<_>>()
+                                .join(", ")
+                        ))
+                        .required(false)
+                        .group("backend"),
                 )
                 .arg(
                     Arg::new("backend-config")
                         .long("backend-config")
-                        .help("config file of backend")
-                        .required(true),
+                        .help("Config string of backend")
+                        .required(false),
+                )
+                .arg(
+                    Arg::new("backend-config-file")
+                        .long("backend-config-file")
+                        .help("Config file of backend")
+                        .conflicts_with("backend-config")
+                        .required(false),
+                )
+                .arg(
+                    Arg::new("blob")
+                        .long("blob")
+                        .short('b')
+                        .help("Path to RAFS data blob file")
+                        .required(false)
+                        .group("backend"),
+                )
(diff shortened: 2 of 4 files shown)


SToPire:
As it may break the existing usage of `nydus-image unpack`, I hope to take more advices for this PR. Then I will add the corresponding tests and update the doc.

src/bin/nydus-image/main.rs
```diff
@@ -1778,15 +1797,80 @@ impl Command {
...
+                    fs::read_to_string(backend_file).with_context(|| {
+                        format!("fail to read backend config file {:?}", backend_file)
+                    })?
+                } else if let Some(backend_config) = matches.get_one::<String>("backend-config") {
+                    backend_config.clone()
+                } else {
+                    bail!("--backend-config or --backend-config-file must be specified");
+                };
+
+            backend = if backend_type == "localfs" {
```
imeoer, line 1832: Maybe refactor `BlobFactory::new_backend` or add a new `BlobFactory::new_backend_from` to create backend from config string, so that we can't miss the handle of the new backend types in future.

imeoer (commented):
others LGMT, thanks!

storage/src/factory.rs
```diff
@@ -245,6 +248,43 @@ impl BlobFactory {
         }
     }
 
+    pub fn new_backend_from_json(
```
imeoer, line 251: Can we reuse `new_backend` here? we can also reuse the handle of `#[cfg(feature = "backend-xxx")]`.

smoke/tests/main_test.go
```diff
@@ -22,10 +22,16 @@ func TestMain(m *testing.M) {
...
 	if os.Getenv("DISABLE_REGISTRY") == "" {
-		reg := tool.NewRegistry()
-		defer reg.Destroy()
+		reg = tool.NewRegistry()
 	}
+
 	code := m.Run()
+
+	if reg != nil {
+		reg.Destroy()
```
imeoer, line 34: Why are we taking it out?

src/bin/nydus-image/main.rs
```diff
@@ -671,17 +671,41 @@ fn prepare_cmd_args(bti_string: &'static str) -> App {
...
                 )
                 .arg(
-                    Arg::new("config")
-                        .long("config")
-                        .short('C')
-                        .help("config to compactor")
-                        .required(true),
+                    Arg::new("backend-type")
+                        .long("backend-type")
+                        .help("Type of backend [possible values: localdisk, s3, oss, registry, http-proxy]")
```
imeoer, line 676: Maybe provide a list method of backend types in `storage/src/factory.rs` and use it here.

src/bin/nydus-image/main.rs
```diff
@@ -703,10 +732,24 @@ fn prepare_cmd_args(bti_string: &'static str) -> App {
                     .help("File path of RAFS metadata")
                     .required_unless_present("bootstrap"),
             )
+            .arg(
+                Arg::new("backend-type")
+                    .long("backend-type")
+                    .help("Type of backend [possible values: localdisk, s3, oss, registry, http-proxy]")
```
imeoer, line 738: Ditto.

smoke/tests/main_test.go
SToPire: As I mentioned in the commit message, the cleanup logic registered with `defer` will not be executed when `os.Exit()` is called.

src/bin/nydus-image/main.rs
SToPire: Yes, I will try it.

storage/src/factory.rs
imeoer: Okay, let's only add the `feature` flag. :)

imeoer (commented):
others LGTM!

imeoer (approved):
LGTM, thanks!

PR #1609, opened 2024-08-10
Fix: Correct compression algorithm case for Zstd in from_str implemen…

daiyongxuan:


daiyongxuan:
It seems remains bugs, I will check it.

PR #1612, opened 2024-08-21
feat: support local tarball for nydusify copy

BruceAko:
## Relevant Issue
https://github.com/dragonflyoss/nydus/issues/1592

## Details
proposal:
https://www.yuque.com/chongzhi-rrdux/qn22ch/rolbzm0xe00g1a3b

implement:
https://www.yuque.com/chongzhi-rrdux/qn22ch/tgg8w0rcazmyoo8f

## Types of changes

## Checklist

### DIFF
--- a/contrib/nydusify/pkg/converter/provider/provider.go
+++ b/contrib/nydusify/pkg/converter/provider/provider.go
@@ -170,7 +170,7 @@ func (pvd *Provider) Import(ctx context.Context, reader io.Reader) (string, erro
 	}
 
 	if len(images) != 1 {
-		return "", errors.New("importing multiple images")
+		return "", errors.New("incorrect tarball format")
 	}
 	image := images[0]
 
--- a/contrib/nydusify/pkg/converter/provider/provider.go
+++ b/contrib/nydusify/pkg/converter/provider/provider.go
@@ -7,6 +7,7 @@ package provider
 import (
 	"context"
 	"crypto/tls"
+	"io"
 	"net"
 	"net/http"
 	"os"
@@ -17,13 +18,16 @@ import (
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/content"
 	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/containerd/images/archive"
 	"github.com/containerd/containerd/platforms"
 	"github.com/containerd/containerd/remotes"
 	"github.com/containerd/containerd/remotes/docker"
 	"github.com/goharbor/acceleration-service/pkg/cache"
 	accelcontent "github.com/goharbor/acceleration-service/pkg/content"
 	"github.com/goharbor/acceleration-service/pkg/remote"
+	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
 )
 
 var LayerConcurrentLimit = 5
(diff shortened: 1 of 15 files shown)


contrib/nydusify/pkg/copier/copier.go
```diff
@@ -285,41 +286,78 @@ func Copy(ctx context.Context, opt Opt) error {
...
-		return errors.Wrap(err, "parse source reference")
-	}
-	targetNamed, err := docker.ParseDockerRef(opt.Target)
-	if err != nil {
-		return errors.Wrap(err, "parse target reference")
-	}
-	source := sourceNamed.String()
-	target := targetNamed.String()
+	var source string
+	if strings.HasPrefix(opt.Source, "file:///") {
```
imeoer, line 290: Can we use `url.Parse` here?

contrib/nydusify/pkg/copier/copier.go
```diff
@@ -285,41 +286,78 @@ func Copy(ctx context.Context, opt Opt) error {
...
+		inputPath := strings.TrimPrefix(opt.Source, "file:///")
+
+		logrus.Infof("importing source image from %s", inputPath)
+		f, err := os.Open(inputPath)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		decompressor, err := compression.DecompressStream(f)
```
imeoer, line 299: `decompressor` should be closed.

contrib/nydusify/pkg/converter/provider/provider.go
```diff
@@ -152,6 +156,36 @@ func (pvd *Provider) Push(ctx context.Context, desc ocispec.Descriptor, ref stri
...
+		skipDgstRef:     func(name string) bool { return name != "" },
+		platformMatcher: pvd.platformMC,
+	}
+	images, err := load(ctx, reader, pvd.store, iopts)
+	if err != nil {
+		return "", err
+	}
+
+	if len(images) != 1 {
+		return "", errors.New("importing multiple images")
```
imeoer, line 173: We should support multi platforms image.

imeoer:
Please add test case in `smoke/tests/nydusify_copy_test.go`, thanks!

imeoer:
Also add a doc about `nydusify copy` in `docs/nydusify.md`. :)

contrib/nydusify/pkg/copier/copier.go
BruceAko: When given an absolute path, if we use `url.Parse`,  it incorrectly resolves it into `URL.host` instead of `URL.path`. So as an alternative, I've referenced [helm](https://sourcegraph.com/github.com/helm/helm/-/blob/internal/resolver/resolver.go?L256-260) and added a function getLocalPath(). A detailed description of this function is in the latest commit, PTAL, thanks!

contrib/nydusify/pkg/copier/copier.go
BruceAko: Resolved.

contrib/nydusify/pkg/converter/provider/provider.go
BruceAko: image

The length of `images` array represents the number of images in the tarball, not manifests. As shown in the figure, there are two images `airpine` and `nginx` in the tarball, each of which is multi platform image storing multiple manifests(`linux/amd64` and `linux/arm64/v8`).

Since all of our exported tarball will only contain 1 image and it's manifests, it can be assumed that if the length of `images` in the tarball is greater than 1, the tarball is not formatted correctly. 

So maybe this code doesn't need to be changed?

BruceAko:
> Also add a doc about `nydusify copy` in `docs/nydusify.md`. :)

Done.

BruceAko:
> Please add test case in `smoke/tests/nydusify_copy_test.go`, thanks!

The E2E test process is as follows:
1. `source` is registry repository, `target` is local tarball, run `nydusify copy` to determine if the tarball was successfully exported;
2. `source` is the just-exported tarball, `target` is remote reposiroty, perform another `nydusify copy` to determine whether the tarball was successfully imported;
3. Run `nydusify check` to determine if the image which was just imported into reposiroty is correct.

Is there anything else I should add into test cases? For example, do I need to test the case where the `backend` is `oss` or `s3`?

contrib/nydusify/pkg/converter/provider/provider.go
imeoer: Thanks for the detail, LGTM.

BruceAko:
> Please add test case in `smoke/tests/nydusify_copy_test.go`, thanks!

Since `smoke/tests/image_test.go` already contained several test cases about `nydusify copy`, instead of creating a new file, I added a function `testNydusifyCopy()` to the `image_test.go` to refine the checking of the correctness of `copy` by different Nydus images.

imeoer (approved):
LGTM! Thanks for the work.

PR #1616, opened 2024-09-11
refactor: fixed the unnecessary mutex lock operation

Chasing1020:

## Details

This PR is a preliminary proposal, and I welcome friendly discussions on this issue.

Due to `Rc` is neither `Send` or `Sync`, it cannot be sent between threads safely. Rust cannot send an `Rc` to another thread, then you have to construct the value on the thread that will use it, or send the value to that thread before wrapping it in an `Rc`. So rustc's mechanism will not exploit the `Mutex`'s protection.

IMHO, for the most part, we should use `Rc<RefCell<T>>` for single thread or `Arc<Mutex<T>>` for exclusive access (until compiler complain about our code need Send or Sync).

Thank you for your time!

## Types of changes

## Checklist

### DIFF
--- a/builder/src/chunkdict_generator.rs
+++ b/builder/src/chunkdict_generator.rs
@@ -93,7 +93,7 @@ impl Generator {
     /// Validate tree.
     fn validate_tree(tree: &Tree) -> Result<()> {
         let pre = &mut |t: &Tree| -> Result<()> {
-            let node = t.lock_node();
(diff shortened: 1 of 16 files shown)


imeoer (approved):
Thanks for the changes, let's add SOB for the commit by following https://github.com/dragonflyoss/nydus/pull/1616/checks?check_run_id=29996970070

PR #1617, opened 2024-09-16
feat: add retry mechanism in read blob metadata

Desiki-high:
https://github.com/dragonflyoss/nydus/issues/1613

## Details
When reading blob size from blob metadata, we should retry reading from the remote if an error occurs.
- Handle the BackendError.
- Set the max retry times to 3.

## Types of changes

## Checklist


imeoer (approved):
LGTM!

PR #1689, opened 2025-04-25
[registry] Fix auth token parsing for access_token

Fricounet:

## Details

Extend auth token parsing to support token in different json fields. 
There is no real consensus on Oauth2 token response format, which means that each registry can implement their own. In particular, Azure ACR uses `access_token` as described here https://github.com/Azure/acr/blob/main/docs/Token-BasicAuth.md#get-a-pull-access-token-for-the-user. As such, when attempting to parse the JSON response containing the authorization token, we should attempt to deserialize using either `token` or `access_token` (and potentially more fields in the future if needed). 

To not break the integration with existing registry, the behavior is to fallback to `access_token` only if `token` does not exist in the response. The requested behavior is implemented by an intermediary struct `TokenRegistryResponse` which allows both `token` and `access_token` as Options (to not fail the JSON parsing if one is missing). Afterwards, the struct is cast to `TokenRegistry`, selecting at the same time the right token field by order of priority.

## Types of changes

## Checklist

### DIFF
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -120,20 +120,13 @@ impl<T> HashCache<T> {
 }
 
 #[derive(Clone, serde::Deserialize)]
-struct TokenRegistryResponse {
-    /// Registry token string.
-    /// This field might vary depending on the registry server.
-    token: Option<String>,
-    access_token: Option<String>,
-    /// Registry token period of validity, in seconds.
-    #[serde(default = "default_expires_in")]
-    expires_in: u64,
-}
-
-#[derive(Clone, serde::Deserialize)]
 struct TokenResponse {
     /// Registry token string.
+    /// This field might vary depending on the registry server.
+    #[serde(default)]
     token: String,
+    #[serde(default)]
+    access_token: String,
     /// Registry token period of validity, in seconds.
     #[serde(default = "default_expires_in")]
     expires_in: u64,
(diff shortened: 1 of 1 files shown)


storage/src/backend/registry.rs
```diff
@@ -119,6 +119,17 @@ impl<T> HashCache<T> {
     }
 }
 
+#[derive(Clone, serde::Deserialize)]
+struct TokenRegistryResponse {
```
imeoer, line 123: Thanks, very nice PR!

How about reusing the `TokenResponse` struct but:

```
struct TokenResponse {
    #[serde(default)]
    token: String,
    #[serde(default)]
    access_token: String,
    #[serde(default = "default_expires_in")]
    expires_in: u64,
}
```

Then we can implement:

```
impl TokenResponse {
    fn from_resp(resp: Response) -> Result<Self> {
        // TODO: the codes in previous `extract_token`.
    }
}
```

And use it:

```
let ret = TokenResponse::from_resp(resp)?;
```

Fricounet: Thanks a lot for your comment. I'm very new to Rust so I do appreciate a lot your feedback to improve my code!

I didn't know about `serde(default)` so that's indeed a good way to not have to add an additional struct. 
I've updated my code in a2ca336 with this change. If that works for you, I'll squash the commit properly afterwards.

imeoer (commented):
very nice updates, LGTM!

imeoer (approved):

PR #1690, opened 2025-04-27
readme: add deepwiki reference

bergwolf:
Add deepwiki reference which is indeed a good place for nydus documentation.


BraveY (approved):
LGTM!

PR #1703, opened 2025-06-04
feat(nydusify): After converting the image, if the push operation fails, increase the number of retries.

ymygogogo:
## Details
When the image is too large, during the push process, there might be some issues in the middle, causing a significant amount of time to be spent on conversion each time.  To address this problem, a retry mechanism was added during the push process.

## Types of changes

## Checklist

### DIFF
--- a/contrib/nydusify/cmd/nydusify.go
+++ b/contrib/nydusify/cmd/nydusify.go
@@ -432,6 +432,18 @@ func main() {
 					Usage:   "Enable plain http for Nydus image push",
 					EnvVars: []string{"PLAIN_HTTP"},
 				},
+				&cli.IntFlag{
+					Name:    "push-retry-count",
+					Value:   3,
+					Usage:   "Number of retries when pushing to registry fails",
+					EnvVars: []string{"PUSH_RETRY_COUNT"},
+				},
+				&cli.StringFlag{
+					Name:    "push-retry-delay",
+					Value:   "5s",
+					Usage:   "Delay between push retries (e.g. 5s, 1m, 1h)",
+					EnvVars: []string{"PUSH_RETRY_DELAY"},
+				},
 			},
 			Action: func(c *cli.Context) error {
 				setupLogLevel(c)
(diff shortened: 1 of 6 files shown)


contrib/nydusify/cmd/nydusify.go
```diff
@@ -432,6 +432,18 @@ func main() {
 					Usage:   "Enable plain http for Nydus image push",
 					EnvVars: []string{"PLAIN_HTTP"},
 				},
+				&cli.IntFlag{
+					Name:    "push-retry-count",
```
imeoer, line 436: Perhaps the flag related to push retry is unnecessary, can we simply provide the fixed value?

contrib/nydusify/pkg/converter/provider/provider.go
```diff
@@ -153,7 +166,28 @@ func (pvd *Provider) Push(ctx context.Context, desc ocispec.Descriptor, ref stri
 		MaxConcurrentUploadedLayers: LayerConcurrentLimit,
 	}
 
-	return push(ctx, pvd.store, rc, desc, ref)
+	// Add retry logic
+	var lastErr error
+	for i := 0; i < pvd.pushRetryCount; i++ {
```
imeoer: How about reusing the `utils.WithRetry` method? We can improve it rather than rewrite it.

BraveY: https://github.com/avast/retry-go Maybe this lib could be used in refactor.

contrib/nydusify/cmd/nydusify.go
ymygogogo: I think this parameter is necessary and a default value has been added. If you think it needs to be modified, you can also make the modification

contrib/nydusify/pkg/converter/provider/provider.go
ymygogogo: The modifications have been completed here. Please take another look for me @imeoer

imeoer (commented):
We added two retry options, are these necessary for the use case? They might be adding unnecessary complexity to the CLI. :)

ymygogogo:
> We added two retry options, are these necessary for the use case? They might be adding unnecessary complexity to the CLI. :)

I think this parameter is still necessary. When using it, you can decide whether to add it based on your own needs.

ymygogogo:
These two parameters should be very important to me.

> We added two retry options, are these necessary for the use case? They might be adding unnecessary complexity to the CLI. :)我们添加了两个重试选项，这些对于用例来说是必要的吗？它们可能会给 CLI 增加不必要的复杂性。:)
These two parameters should be very important to me.

ymygogogo:
Please take a look for me again. smoke test doesn't seem to be my problem @imeoer

imeoer (approved):
Thanks, LGTM!

PR #1717, opened 2025-07-10
fix: Add filesystem sync in both container and host namespaces before commit

jackowfish:

_No specific issue tracked - discovered during testing that committed container changes weren't persisting._

### Problem
Found a bug where `nydusify commit` would succeed and show the right blob sizes, but when you pull the new image, recent file changes are missing. Believe the issue is that filesystem changes were stuck in memory buffers and not actually written to disk before the commit snapshot happened. 

### Solution
Added a filesystem sync step before committing:
- Run `sync` inside the container namespace to flush buffers
- Run `sync` on the host to make sure overlay filesystem changes hit disk

Uses the existing `nsenter` code to do this properly. If sync fails, it warns but continues the commit so existing workflows don't break.

### Files Changed
- `contrib/nydusify/pkg/committer/commiter.go`: Added sync step to commit flow

## Types of changes

## Checklist

### DIFF
--- a/contrib/nydusify/pkg/committer/commiter.go
+++ b/contrib/nydusify/pkg/committer/commiter.go
@@ -230,6 +230,14 @@ func (cm *Committer) Commit(ctx context.Context, opt Opt) error {
 		return appendedEg.Wait()
 	}
 
+	// Ensure filesystem changes are written to disk before committing
+	// This prevents issues where changes are still in memory buffers
+	// and not yet visible in the overlay filesystem's upper directory
+	logrus.Infof("syncing filesystem before commit")
+	if err := cm.syncFilesystem(ctx, opt.ContainerID); err != nil {
+		logrus.Warnf("failed to sync filesystem, continuing anyway: %v", err)
+	}
+
 	if err := cm.pause(ctx, opt.ContainerID, commit); err != nil {
 		return errors.Wrap(err, "pause container to commit")
 	}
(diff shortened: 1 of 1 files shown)


contrib/nydusify/pkg/committer/commiter.go
```diff
@@ -230,6 +230,14 @@ func (cm *Committer) Commit(ctx context.Context, opt Opt) error {
 		return appendedEg.Wait()
 	}
 
+	// Ensure filesystem changes are written to disk before committing
+	// This prevents issues where changes are still in memory buffers
+	// and not yet visible in the overlay filesystem's upper directory
+	logrus.Infof("syncing filesystem before commit")
+	if err := cm.syncFilesystem(ctx, opt.ContainerID); err != nil {
+		logrus.Warnf("failed to sync filesystem, continuing anyway: %v", err)
```
Zephyrcf: if sync fails, the image can be inconsistent due to unflushed data. A WARN isn't enough here. To ensure image integrity, sync failures should abort the commit. Can we make this a fatal error?

imeoer:
Hi @jackowfish Thanks for the PR, it's LGTM. This change is based on the reason that overlayfs / kernel only guarantees that the merge dir reflects the latest data, but does not guarantee the read consistency from upper dir, right?

contrib/nydusify/pkg/committer/commiter.go
jackowfish: @Zephyrcf Done 👍

jackowfish:
> Hi @jackowfish Thanks for the PR, it's LGTM. This change is based on the reason that overlayfs / kernel only guarantees that the merge dir reflects the latest data, but does not guarantee the read consistency from upper dir, right?

@imeoer That was my thinking - my understanding is that overlayfs guarantees consistency in the merged directory but not immediate read consistency from the upper directory due to buffering. Since nydusify reads directly from the upper directory for commits, we need the sync to flush everything first.
Thanks for the review!

imeoer (approved):
Thanks!

imeoer:
Let’s merge first, the broken CI (from TestNativeLayer) isn’t related to this change, will investigate later.

PR #1733, opened 2025-07-31
[registry] Accept empty scope during token auth challenge

Fricounet:

## Details

The distribution spec (https://distribution.github.io/distribution/spec/auth/scope/#authorization-server-use) mentions that the access token provided during auth challenge "may include a scope" which means that it's not necessary to have one either to comply with the spec. Additionally, this is something that is already accepted by containerd which will simply log a warning when no scope is specified: https://github.com/containerd/containerd/blob/main/core/remotes/docker/auth/fetch.go#L64 To match with what containerd and the spec suggest, the commit modifies the `parse_auth` logic to accept an empty `scope` field. It also logs the same warning as containerd.

## Types of changes

## Checklist

### DIFF
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -437,17 +437,21 @@ impl RegistryState {
                 Some(Auth::Basic(BasicAuth { realm }))
             }
             "Bearer" => {
-                if !paras.contains_key("realm")
-                    || !paras.contains_key("service")
-                    || !paras.contains_key("scope")
-                {
+                if !paras.contains_key("realm") || !paras.contains_key("service") {
                     return None;
                 }
+
+                let scope = if let Some(scope) = paras.get("scope") {
+                    (*scope).to_string()
+                } else {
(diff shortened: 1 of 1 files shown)


storage/src/backend/registry.rs
```diff
@@ -437,17 +437,21 @@ impl RegistryState {
...
-                    || !paras.contains_key("scope")
-                {
+                if !paras.contains_key("realm") || !paras.contains_key("service") {
                     return None;
                 }
 
+                let scope = if let Some(scope) = paras.get("scope") {
+                    (*scope).to_string()
+                } else {
+                    warn!("no scope specified for token auth challenge");
```
imeoer: Thanks for the PR! How about also using `debug!` level log like containerd? Otherwise this may print a large amount of logs. :)

Fricounet: yes good point, I don't know why but I thought containerd was logging this as a warning.
Changed in the latest commit

imeoer (approved):
Thanks!

PR #1739, opened 2025-08-13
[Improvement] unify token fetching and improve validation

### LINKED ISSUE #1729: registry backend: why are username and password in query params when getting token
Hello folks, 

I'm trying to understand why does the `get_token_with_get` function need to have the `username` and `password` field in its[ query params](https://github.com/DataDog/nydus/blob/master/storage/src/backend/registry.rs#L356-L357) when attempting to fetch the token. Especially since they are already present in the `Authorization` [header](https://github.com/DataDog/nydus/blob/master/storage/src/backend/registry.rs#L369) and containerd [implementation](https://github.com/containerd/containerd/blob/main/core/remotes/docker/auth/fetch.go#L196) seem to be content with only the `Authrozation` header.

I found this [commit](https://github.com/DataDog/nydus/commit/069818c6958ad2c9cfe84ecf361b3993fdc2c795) mentioning needing the Authorization header for harbor and ghcr, however it doesn't mention why the query params are needed and since containerd doesn't use it, I kind of think that this must be enough for all registries no?

Zephyrcf:
close https://github.com/dragonflyoss/nydus/issues/1729

## Details
unify token fetching and improve validation.

## Types of changes

## Checklist

### DIFF
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -198,8 +198,6 @@ struct RegistryState {
     repo: String,
     // Base64 encoded registry auth
     auth: Option<String>,
-    username: String,
-    password: String,
     // Retry limit for read operation
     retry_limit: u8,
     // Scheme specified for blob server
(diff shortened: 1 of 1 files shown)


storage/src/backend/registry.rs
```diff
@@ -305,80 +303,59 @@ impl RegistryState {
...
-    fn get_token_with_post(
-        &self,
-        auth: &BearerAuth,
-        connection: &Arc<Connection>,
-    ) -> Result<Response> {
-        let mut form = HashMap::new();
-        form.insert("service".to_string(), auth.service.clone());
-        form.insert("scope".to_string(), auth.scope.clone());
-        form.insert("grant_type".to_string(), "password".to_string());
-        form.insert("username".to_string(), self.username.clone());
```
BraveY, line 318: Considering deleting `passwd` and `username` here.

BraveY:
Thanks. Do we have some test results to support this change?  It would be better to have some unit tests or integration tests.

Zephyrcf:
> Thanks. Do we have some test results to support this change? It would be better to have some unit tests or integration tests.

I think this fluctuation did not destroy the original logic, so just use the current test.
harbor and tencent tcr have been tested locally.
image

BraveY (approved):
LGTM!

Issue #1697, opened 2025-05-26
support fuse resend API for live upgrade and failover


imeoer:
This issue aims to implement request resend for the fuse-backend-rs lib and the nydusd live upgrade and failover, ensuring the fuse requests are resent after nydusd(fuse daemon) recovery, thereby avoiding interruptions in fuse io request processing.

X1aoZEOuO:
Hello, I would like to participate in this GLCC project. Could you please provide some detailed scenarios?

Currently, my understanding of the project is that it requires persisting requests received by the FUSE Daemon (e.g., logging them), and then, when reloading the FUSE Daemon, continuing the previous FUSE requests based on the logs. However, I feel this might reduce the throughput of FUSE requests. Is this approach correct? Additionally, where in the code for `fuse-backend-rs` and `Nydus` should I focus my reading? I would like to study this further.

imeoer:
Hi @X1aoZEOuO Plz take a look the new FUSE notify API for request resend [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=760eac73f9f69aa28fcb3050b4946c2dcc656d12).

X1aoZEOuO:
> Hi [@X1aoZEOuO](https://github.com/X1aoZEOuO) Plz take a look the new FUSE notify API for request resend [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=760eac73f9f69aa28fcb3050b4946c2dcc656d12).

Thank you, I will take a deep look at it.

Zephyrcf:
@lsc2001 He is dealing with this issue.

lsc2001:
> [@lsc2001](https://github.com/lsc2001) He is dealing with this issue.

Yes，I‘ll updating my code.

imeoer:
Fixed in https://github.com/dragonflyoss/nydus/pull/1740

PR #1740, opened 2025-08-16
[Feature] Support resending fuse requests when restoring nydusd.

lsc2001:
## Relevant Issue
#1697 

## Details
Support resending the fuse request when nydusd undergoes a hot upgrade or during a fault recovery process.
Requests can be resend by fuse device and sysfs.

## Types of changes

### DIFF
--- a/service/src/fusedev.rs
+++ b/service/src/fusedev.rs
@@ -8,8 +8,9 @@
 use std::any::Any;
 use std::ffi::{CStr, CString};
 use std::fs::metadata;
-use std::io::{Error, ErrorKind, Result};
+use std::io::{Error, ErrorKind, Result, Write};
 use std::ops::Deref;
+use std::os::fd::AsRawFd;
 #[cfg(target_os = "linux")]
 use std::os::linux::fs::MetadataExt;
 #[cfg(target_os = "linux")]
@@ -17,7 +18,7 @@ use std::os::unix::ffi::OsStrExt;
 #[cfg(target_os = "macos")]
 use std::os::unix::fs::MetadataExt;
 use std::os::unix::net::UnixStream;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::{
     atomic::{AtomicI32, AtomicU64, Ordering},
     mpsc::{channel, Receiver, Sender},
@@ -29,7 +30,7 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use fuse_backend_rs::abi::fuse_abi::{InHeader, OutHeader};
 use fuse_backend_rs::api::server::{MetricsHook, Server};
 use fuse_backend_rs::api::Vfs;
-use fuse_backend_rs::transport::{FuseChannel, FuseSession};
+use fuse_backend_rs::transport::{FuseChannel, FuseDevWriter, FuseSession};
 use mio::Waker;
 #[cfg(target_os = "linux")]
 use nix::sys::stat::{major, minor};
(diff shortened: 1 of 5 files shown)


service/src/fusedev.rs
```diff
@@ -205,6 +206,66 @@ impl FusedevFsService {
...
+
+    fn notify_resend_by_fusedev(&self) -> NydusResult<()> {
+        let session = self.session.lock().unwrap();
+        let file = session
+            .get_fuse_file()
+            .ok_or(NydusError::FuseFileNotFound)?;
+
+        let fd = file.as_raw_fd();
+        let mut buf = vec![0u8; session.bufsize()];
+        let writer: FuseDevWriter = FuseDevWriter::new(fd, &mut buf).unwrap();
```
Zephyrcf:
```suggestion
        let writer: FuseDevWriter = FuseDevWriter::new(fd, &mut buf)?;
```
It is more appropriate to use the ? operator.

lsc2001: Thanks for this suggestion. I'll fix it.

service/src/fusedev.rs
```diff
@@ -205,6 +206,66 @@ impl FusedevFsService {
...
+
+        self.server
+            .notify_resend(writer)
+            .map_err(NydusError::FuseWriteError)?;
+        Ok(())
+    }
+
+    fn notify_resend_by_sysfs(&self) -> NydusResult<()> {
+        // Notify the kernel to resend the request by writing to `/sys/fs/fuse/connections/<cid>/resend`
```
imeoer, line 227: It is better to check if the file exists first, and then write to it, otherwise kernels that do not support this file will throw an error.

service/src/fusedev.rs
```diff
@@ -205,6 +206,66 @@ impl FusedevFsService {
...
+        let mut file = std::fs::OpenOptions::new()
+            .write(true)
+            .open(path)
+            .map_err(NydusError::SysfsOpenError)?;
+
+        file.write_all(b"1").map_err(NydusError::SysfsWriteError)?;
+        Ok(())
+    }
+
+    pub fn do_failover(&self) -> NydusResult<()> {
```
imeoer: May be `do_failover` -> `drain_fuse_requests`

src/bin/nydusd/main.rs
```diff
@@ -113,9 +113,9 @@ fn append_fuse_options(app: Command) -> Command {
     .arg(
         Arg::new("failover-policy")
             .long("failover-policy")
-            .default_value("resend")
+            .default_value("none")
```
imeoer: Maybe keep resend as default value.

service/src/upgrade.rs
```diff
@@ -477,13 +480,10 @@ pub mod fusedev_upgrade {
...
-                .downcast_ref::<FusedevFsService>()
-                .unwrap()
-                .session
-                .lock()
-                .unwrap()
-                .set_fuse_file(f);
+            let fuse_svc = svc.as_any().downcast_ref::<FusedevFsService>().unwrap();
+            fuse_svc.session.lock().unwrap().set_fuse_file(f);
+            // do failover policy
+            fuse_svc.do_failover()?;
```
imeoer: How about printing a warning instead of throwing an error here? Lower kernel versions do not support resend or flush, compatibility must be ensured.

service/src/lib.rs
```diff
@@ -106,6 +106,14 @@ pub enum Error {
     // Fuse session has been shutdown.
     #[error("FUSE session has been shut down, {0}")]
     SessionShutdown(FuseTransportError),
+    #[error("FUSE file not found")]
+    FuseFileNotFound,
```
imeoer: FuseDeviceNotFound?

service/src/fusedev.rs
xujihui1985: se.with_writer(|writer| {
            self.server
                .notify_resend(writer)
                .unwrap_or_else(|e| println!("failed to send resend notification {}", e));
        });

I think we have expose a with_writer on session to make it easy to acquire writer from session

service/src/fusedev.rs
```diff
@@ -205,6 +206,66 @@ impl FusedevFsService {
...
     }
+
+    fn notify_resend_by_fusedev(&self) -> NydusResult<()> {
+        let session = self.session.lock().unwrap();
+        let file = session
+            .get_fuse_file()
+            .ok_or(NydusError::FuseFileNotFound)?;
+
+        let fd = file.as_raw_fd();
+        let mut buf = vec![0u8; session.bufsize()];
```
aftersnow: Maybe it's better to put buf allocation out of the lock

service/src/fusedev.rs
```diff
@@ -205,6 +206,66 @@ impl FusedevFsService {
...
+        self.server
+            .notify_resend(writer)
+            .map_err(NydusError::FuseWriteError)?;
+        Ok(())
+    }
+
+    fn notify_resend_by_sysfs(&self) -> NydusResult<()> {
+        // Notify the kernel to resend the request by writing to `/sys/fs/fuse/connections/<cid>/resend`
+        let path = format!(
+            "/sys/fs/fuse/connections/{}/resend",
```
aftersnow: try to avoid fusectl path hardcode: /sys/fs/fuse/connections/{}/resend

service/src/fusedev.rs
```diff
@@ -205,6 +206,66 @@ impl FusedevFsService {
...
+            .map_err(NydusError::SysfsOpenError)?;
+
+        file.write_all(b"1").map_err(NydusError::SysfsWriteError)?;
+        Ok(())
+    }
+
+    fn notify_flush_by_sysfs(&self) -> NydusResult<()> {
+        // Notify the kernel to flush the request by writing to `/sys/fs/fuse/connections/<cid>/flush`
+        let path = format!(
+            "/sys/fs/fuse/connections/{}/flush",
```
aftersnow: the same

service/src/fusedev.rs
lsc2001: > se.with_writer(|writer| { self.server .notify_resend(writer) .unwrap_or_else(|e| println!("failed to send resend notification {}", e)); });
> 
> I think we have expose a with_writer on session to make it easy to acquire writer from session

I noticed `with_writer`, but I need the error returned by  `notify_resend`.

xujihui1985: ok, I see.

service/src/lib.rs
lsc2001: image
OK, this error is returned by `get_fuse_file`, but fuse device not found is better actually.

service/src/upgrade.rs
lsc2001: OK, I'll fix it.

src/bin/nydusd/main.rs
lsc2001: OK, I'll fix it.

service/src/fusedev.rs
lsc2001: OK, i'll fix it.

service/src/fusedev.rs
lsc2001: OK, I'll fix it.

service/src/fusedev.rs
lsc2001: OK, I'll fix it.

service/src/fusedev.rs
lsc2001: OK, I'll fix it.

service/src/fusedev.rs
lsc2001: > Maybe it's better to put buf allocation out of the lock

But we maybe have to lock session before we get the bufsize.

service/src/fusedev.rs
xujihui1985: https://github.com/cloud-hypervisor/fuse-backend-rs/pull/216 
I make a PR to update the with_writer to try_with_writer to handle error properly, let me know if that satisfy your expectation

lsc2001: > [cloud-hypervisor/fuse-backend-rs#216](https://github.com/cloud-hypervisor/fuse-backend-rs/pull/216) I make a PR to update the with_writer to try_with_writer to handle error properly, let me know if that satisfy your expectation

Yes, I need it. I'll update my code after fuse-backend-rs version of nydusd getting updated.

src/bin/nydusd/main.rs
imeoer: Any updates?

service/src/fusedev.rs
```diff
@@ -145,6 +146,88 @@ impl FuseServer {
...
+                Err(e) => {
+                    if !matches!(e, NydusError::SysfsOpenError(_)) {
+                        return Err(e);
+                    }
+                    open_error = Some(e);
+                }
+            }
+        }
+
+        Err(open_error.unwrap())
```
imeoer: How about:

```
for (idx, path) in paths.iter().enumerate() {
    match self.try_notify_with_path(base_path, event) {
        Ok(()) => return Ok(()),
        Err(e) => {
            if !matches!(e, NydusError::SysfsOpenError(_)) ||
                idx == paths.len() - 1 {
                return Err(e);
            }
        }
    }
}

Ok(())
```

lsc2001: OK, this looks better.

imeoer (approved):
Thanks, LGTM!

PR #1745, opened 2025-08-28
Fix: add recursive inode/dentry invalidation and API smoke test

### LINKED ISSUE #1694: invalidate dentry/inode cache after 255 nydusd mounts
After performing more than 255 sub-mount operations on nydusd (even if unmounting follows mounting), there is a probability that file system corruption will occur when users attempt to mount again.

This issue manifests as accessing files or content from previous mount points. The root cause lies in the fuse-backend-rs library, which uses a u8 type for pseudo inode indexing. Additionally, after unmounting sub-mount points, the library does not call FUSE's invalidate dentry cache or invalidate inode cache. As a result, after exceeding 255 mounts, new mountpoints may reuse the pseudo inode of sub-mount points due to kernel cache not being cleared, leading to file system inconsistencies.

sidneychang:
## Overview
Performing more than 255 sub-mount operations on nydusd can lead to file system corruption, even if each mount is properly unmounted. The corruption occurs because fuse-backend-rs uses a u8 type for pseudo inode indexing, and after unmounting, it does not invalidate the FUSE dentry or inode caches. As a result, kernel caches may retain stale pseudo inodes. Once the 255 pseudo inode limit is exceeded, new mount points can reuse old pseudo inodes, causing users to access files or content from previous mounts, leading to inconsistencies.
## Related Issues
Fix #1694 

## Change Details
- Before unmounting a RAFS filesystem, recursively invalidate all inodes and dentries to prevent stale kernel caches when using shared mounts.
- Added a new smoke test `TestSubMountCache` to mount/unmount sub-mounts multiple times (300 iterations) and verify file, directory, symlink.

## Test Results

In the smoke test, the first mount uses a “thin” filesystem containing fewer files, while the next 255 mounts use a full filesystem with more files.

Before this patch, on the 256th mount, accessing files that exist in the full filesystem but not in the thin filesystem could fail:
```
[2025-08-28 13:39:12.597203 +08:00] INFO RAFS filesystem imported
[2025-08-28 13:39:12.597339 +08:00] INFO Rafs filesystem mounted at /mount-255
[2025-08-28 13:39:12.597602 +08:00] ERROR fuse: reply error header OutHeader { len: 16, error: -5, unique: 72202 }, error Custom { kind: NotADirectory, error: "Not a directory (os error 20)" }
api_test.go:263: 
        Error Trace:    /root/nydus/smoke/tests/api_test.go:263
        Error:          Received unexpected error:
                        stat /tmp/nydus-smoke-1919054383/mnt/mount-255/dir-1/file-1: input/output error
        Test:           TestAPI/TestSubMountCache
        Messages:       stat file-1 failed


```
This happens because the pseudo inode cache from the first mount (thin filesystem) is reused in the 256th mount, and the kernel cache has not been invalidated. As a result, the kernel sees stale pseudo inodes, causing read errors when accessing files that exist in the full filesystem but not in the thin one.

After applying this patch, the test passes completely, as recursive inode/dentry invalidation ensures that kernel caches are cleared and pseudo inode reuse no longer causes inconsistencies.
```
[2025-08-28 13:59:08.047650 +08:00] INFO State machine(pid=1116461): from Running to Ready, input [Stop], output [Some(TerminateService)]
[2025-08-28 13:59:08.047787 +08:00] INFO State machine(pid=1116461): from Ready to Die, input [Stop], output [Some(Umount)]
[2025-08-28 13:59:08.047835 +08:00] INFO state_machine thread exits
--- PASS: TestAPI (0.00s)
    --- PASS: TestAPI/TestSubMountCache (9.30s)
PASS
```

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/service/src/fs_service.rs
+++ b/service/src/fs_service.rs
@@ -13,6 +13,7 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::{Arc, MutexGuard};
 
+use fuse_backend_rs::abi::fuse_abi::ROOT_ID;
 #[cfg(target_os = "linux")]
 use fuse_backend_rs::api::filesystem::{FileSystem, FsOptions, Layer};
 use fuse_backend_rs::api::vfs::VfsError;
@@ -23,6 +24,7 @@ use fuse_backend_rs::overlayfs::{config::Config as overlay_config, OverlayFs};
 use fuse_backend_rs::passthrough::{CachePolicy, Config as passthrough_config, PassthroughFs};
 use nydus_api::ConfigV2;
 use nydus_rafs::fs::Rafs;
+use nydus_rafs::metadata::RafsInode;
 use nydus_rafs::{RafsError, RafsIoRead};
 use nydus_storage::factory::BLOB_FACTORY;
 use serde::{Deserialize, Serialize};
@@ -86,7 +88,7 @@ impl FsBackendCollection {
         Ok(())
     }
 
-    fn del(&mut self, id: &str) {
+    pub fn del(&mut self, id: &str) {
         self.0.remove(id);
     }
 }
@@ -99,7 +101,7 @@ pub trait FsService: Send + Sync {
 
     /// Get the [BackFileSystem](https://docs.rs/fuse-backend-rs/latest/fuse_backend_rs/api/vfs/type.BackFileSystem.html)
     /// object associated with a mount point.
-    fn backend_from_mountpoint(&self, mp: &str) -> Result<Option<Arc<BackFileSystem>>> {
+    fn backend_from_mountpoint(&self, mp: &str) -> Result<Option<(Arc<BackFileSystem>, u8)>> {
         self.get_vfs().get_rootfs(mp).map_err(|e| e.into())
     }
 
@@ -133,7 +135,7 @@ pub trait FsService: Send + Sync {
 
     /// Remount a filesystem instance.
     fn remount(&self, cmd: FsBackendMountCmd) -> Result<()> {
-        let rootfs = self
+        let (rootfs, _) = self
             .backend_from_mountpoint(&cmd.mountpoint)?
             .ok_or(Error::NotFound)?;
         let mut bootstrap = <dyn RafsIoRead>::from_file(&cmd.source)?;
@@ -178,9 +180,23 @@ pub trait FsService: Send + Sync {
 
     /// Umount a filesystem instance.
     fn umount(&self, cmd: FsBackendUmountCmd) -> Result<()> {
-        let _ = self
+        let (fs, fs_idx) = self
             .backend_from_mountpoint(&cmd.mountpoint)?
             .ok_or(Error::NotFound)?;
+
+        if self.is_fuse() {
+            if let Some(rafs) = fs.deref().as_any().downcast_ref::<Rafs>() {
+                let root_ino = rafs.get_root_inode().unwrap();
+                self.walk_and_notify_invalidation(
+                    ROOT_ID,
+                    cmd.mountpoint.trim_start_matches('/'),
+                    root_ino,
+                    fs_idx,
+                )?;
+            }
+        }
+
+        drop(fs);
 
         self.get_vfs().umount(&cmd.mountpoint)?;
         self.backend_collection().del(&cmd.mountpoint);
@@ -201,7 +217,7 @@ pub trait FsService: Send + Sync {
 
     /// Export information about the filesystem service.
     fn export_backend_info(&self, mountpoint: &str) -> Result<String> {
-        let fs = self
+        let (fs, _) = self
             .backend_from_mountpoint(mountpoint)?
             .ok_or(Error::NotFound)?;
         let any_fs = fs.deref().as_any();
@@ -215,6 +231,21 @@ pub trait FsService: Send + Sync {
     /// Export metrics about in-flight operations.
     fn export_inflight_ops(&self) -> Result<Option<String>>;
 
+    /// Recursively walk the inode tree and send cache invalidation notifications.
+    fn walk_and_notify_invalidation(
+        &self,
+        _parent_kernel_ino: u64,
+        _cur_name: &str,
+        _cur_inode: Arc<dyn RafsInode>,
+        _fs_idx: u8,
+    ) -> Result<()> {
+        Ok(())
+    }
+
+    /// Check whether the filesystem service is a FUSE service.
+    fn is_fuse(&self) -> bool {
+        false
+    }
     /// Cast `self` to trait object of [Any] to support object downcast.
     fn as_any(&self) -> &dyn Any;
 }
--- a/service/src/fusedev.rs
+++ b/service/src/fusedev.rs
@@ -5,8 +5,10 @@
 
 //! Nydus FUSE filesystem daemon.
 
+use core::option::Option::None;
+use nydus_rafs::metadata::{RafsInode, RafsInodeWalkAction};
 use std::any::Any;
-use std::ffi::{CStr, CString};
+use std::ffi::{CStr, CString, OsStr, OsString};
 use std::fs::metadata;
 use std::io::{Error, ErrorKind, Result, Write};
 use std::ops::Deref;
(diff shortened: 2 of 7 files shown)


sidneychang:
Due to modifications in fuse-backend-rs, the project is now using version v0.13.0. The code constructs FuseDevWriter directly from a file descriptor (raw_fd) to work around current limitations.
There are two known issues in fuse-backend-rs:
**with_writer**: Switching to with_writer could simplify the code, but it requires additional macOS support in the library. Without that, CI builds on macOS would fail.
**unsafe usage**: There are two places in the library where certain calls need to be wrapped in unsafe blocks.
For now, we continue using the raw_fd approach, accepting that CI may fail, and will revisit these issues when we decide on the proper modifications to the library.

sidneychang:
I am currently using my own fork of fuse-backend-rs to verify whether the CI can pass. Once the corresponding commits are merged into the upstream fuse-backend-rs repository and a release is published, I will switch back.

service/src/fusedev.rs
```diff
@@ -198,6 +205,93 @@ impl FusedevFsService {
...
+        stack.push((
+            ROOT_PARENT_INO,
+            path.trim_start_matches('/').to_string(),
+            inode.clone(),
+            false,
+        ));
+
+        while let Some((parent_kernel_ino, cur_name, cur_inode, visited)) = stack.pop() {
+            // Convert rafs inode to inode id used by kernel
+            let cur_kernel_ino = ((fs_idx as u64) << FS_IDX_SHIFT) | cur_inode.ino();
```
imeoer: May be use `Vfs::convert_inode` here (not require for this PR).

service/src/fusedev.rs
```diff
@@ -198,6 +205,93 @@ impl FusedevFsService {
 
         inflight_op
     }
+    fn walk_and_notify_invalidation(
```
imeoer: Can we reuse `RafsInode::walk_children_inodes` here?

service/src/fusedev.rs
```diff
@@ -237,6 +331,31 @@ impl FsService for FusedevFsService {
         }
     }
 
+    fn umount(&self, cmd: FsBackendUmountCmd) -> NydusResult<()> {
```
imeoer: Can we call `service/src/fusedev.rs -> walk_and_notify_invalidation` in the `service/src/fs_service.rs -> umount`? So that we can reuse the `umount` codes.

smoke/tests/api_test.go
```diff
@@ -193,6 +193,109 @@ func (a *APIV1TestSuite) TestPrefetch(t *testing.T) {
...
+		EnablePrefetch:  ctx.Runtime.EnablePrefetch,
+		DigestValidate:  false,
+		AmplifyIO:       ctx.Runtime.AmplifyIO,
+	}
+
+	// mount prefix where the daemon actually mounts sub-mounts
+	mountPrefix := ctx.Env.WorkDir + "/mnt"
+
+	// iterate 256 times; first mount uses thinlower (without some files), subsequent mounts use Lower.
+	for i := 0; i < 256; i++ {
```
imeoer: It's better to generate different file names and content (e.g., the `texture.MakeMatrixLayer` method), mount exceed 255 count, and verify the fs.

service/src/fusedev.rs
sidneychang: I’m not sure whether the kernel cache should always be invalidated in all scenarios. For virtiofs, the upstream PR [fuse-backend-rs#152](https://github.com/cloud-hypervisor/fuse-backend-rs/pull/152)
 notes that VFS keeps pseudo inodes on umount to support hot upgrade and quick recovery, which suggests that retaining the kernel dentry/inode cache might also be intentional. For normal fuse, however, not clearing the cache exposes invalid directories on the host, so explicit invalidation is required. If we decide that invalidating the cache is necessary in all cases, I can move the logic into fs_service::umount so it’s handled consistently.

> Can we call `service/src/fusedev.rs -> walk_and_notify_invalidation` in the `service/src/fs_service.rs -> umount`? So that we can reuse the `umount` codes.

service/src/fusedev.rs
sidneychang, line 208: walk_children_inodes is already used here to collect the children. The reason I implemented a manual DFS stack is that in the v5 metadata implementation, the handler doesn’t provide an Arc<dyn RafsInode>, so recursion can’t be done directly. If we only need to support v6 (which passes the inode to the handler), then yes, I can simplify this and reuse walk_children_inodes directly for recursion.

smoke/tests/api_test.go
sidneychang: Thanks for pointing this out — the current test only repeats with fixed file names and content. I’ll switch to MakeMatrixLayer to generate varied files and mount more than 255 times to better cover cache and boundary cases.

service/src/fusedev.rs
sidneychang: walk_children_inodes is already used here to collect the children. The reason I implemented a manual DFS stack is that in the v5 metadata implementation, the handler doesn’t provide an Arc, so recursion can’t be done directly. If we only need to support v6 (which passes the inode to the handler), then yes, I can simplify this and reuse walk_children_inodes directly for recursion.

service/src/fs_service.rs
```diff
@@ -32,6 +33,8 @@ use versionize_derive::Versionize;
 use crate::upgrade::UpgradeManager;
 use crate::{Error, FsBackendDescriptor, FsBackendType, Result};
 
+const ROOT_PARENT_INO: u64 = 1;
```
imeoer: Mey be use `fuse_backend_rs::abi::fuse_abi::ROOT_ID`.

service/src/fusedev.rs
```diff
@@ -335,6 +339,65 @@ impl FsService for FusedevFsService {
...
+            cur_inode.walk_children_inodes(0, &mut handler)?;
+        }
+
+        // === Post-order: invalidate cache of the current node ===
+        let cstr_name = CString::new(cur_name).map_err(|_| eother!("invalid file name"))?;
+        // Invalidate inode cache
+        self.session.lock().unwrap().with_writer(|writer| {
+            if let Err(e) = self.server.notify_inval_inode(writer, cur_kernel_ino, 0, 0) {
+                warn!("notify_inval_inode failed: {} {:?}", cur_name, e);
+            }
```
imeoer, line 381: Can we also put `notify_inval_entry` here.

smoke/tests/api_test.go
```diff
@@ -192,6 +192,80 @@ func (a *APIV1TestSuite) TestPrefetch(t *testing.T) {
...
+		curCfg := childTpl
+		curCfg.BootstrapPath = bootstrap
+		curCfg.MountPath = childTpl.MountPath + fmt.Sprintf("-%d", i)
+
+		err = nydusd.MountByAPI(curCfg)
+		require.NoError(t, err, "failed to mount by API at iteration %d", i)
+
+		fullMount := filepath.Join(mountPrefix, curCfg.MountPath)
+
+		// Verify file contents
```
imeoer: It seems the check is unnecessary, since `nydusd.VerifyByPath` will also verify the file's metadata and content.

imeoer (commented):
Thanks, others LGTM!

service/src/fusedev.rs
sidneychang: Since both notify_inval_inode and notify_inval_entry require writer as a mutable borrow, we cannot perform two write operations within the same with_writer closure. Each notify function requires exclusive access to writer, so we must call with_writer twice.
```
pub fn notify_inval_inode<S: BitmapSlice>(
    &self,
    mut w: FuseDevWriter<'_, S>,  // mutable borrow
    ...
) 

pub fn notify_inval_entry<S: BitmapSlice>(
    &self,
    mut w: FuseDevWriter<'_, S>,  // mutable borrow
    ...
) -
```

imeoer (approved):
Thanks, LGTM!

imeoer:
The broken CI will be resolved next PR.

PR #1776, opened 2025-10-02
Use native RAFS in `nydusify commit` instead of tarfs and custom tarball logic

markasoftware-tc:
## Overview
This PR replaces much of the `nydusify commit` logic by using existing features in `nydus-image`. This simplifies the code and improves performance.

## Related Issues
No related issue.

## Change Details
The existing `nydusify commit` logic uses Go to stream a tarball of modified files and whiteouts into `nydus-image create` in tarfs mode. Without historical context I don't know exactly why this was done, but my guess is that `nydus-image` didn't handle overlayfs-format whiteouts at the time `nydusify commit` was written. This logic duplicates some of the work `nydus-image` is about to do anyway, and isn't implemented in the most efficient way (for example, it stats all upper-layer files on the lower layer as well, to check if they are identical).

This PR just calls `nydus-image create --whiteout-spec overlayfs` to do effectively the same thing. In my benchmarks, this typically improves performance by at least 25% on fast disks.

`nydus-image create` has the `--parent-bootstrap` option performs a merge internally, so this PR also removes the explicit calls to `nydus-image merge` to simplify things further.

## Test Results
I have manually tested committing, including with deleted files, for correctness.

## Change Type
- [ ] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [X] Code Refactoring
- [X] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

Remaining items I'm unsure about:
* Haven't done super extensive testing in general.


imeoer:
Hi @markasoftware-tc, thanks for the PR, we initially considered using the diff approach for the lower and upper dirs, is the `fn Changes` in `contrib/nydusify/pkg/committer/diff/overlay_linux.go`, because it implements stricter and complete overlayfs key features, such as `checkRedirect` and `checkOpaque`. But the `nydus-image create --whiteout-spec overlayfs` only consider about the upper dir, it's not a complete implementation for overlayfs diff. :(

Issue #1671, opened 2025-02-24
Flag checks cause some flags introduced in the new version to be unrecognized by the old version of nydusd.


BraveY:
There are two specific checks in the current implementation that are causing compatibility issues between the new and old versions of nydusd:
1. Blob Feature Checks: https://github.com/dragonflyoss/nydus/blob/dd9ba54e332802e46316755314c54b588f3b0889/rafs/src/metadata/layout/v5.rs#L656
2. Chunk Flag Checks:
https://github.com/dragonflyoss/nydus/blob/91931607f8188482315b97b687448024c979f902/storage/src/meta/chunk_info_v2.rs#L232

These two checks collectively prevent the old version of nydusd from recognizing and properly handling new flags introduced in the new version, leading to potential runtime errors or failures during execution.

BraveY:
Closed by #1786

PR #1786, opened 2025-10-15
Fix: fix chunk flag compatibility by replacing error with warning

sidneychang:
## Overview
Fix compatibility issues between old and new versions of nydusd by replacing strict error handling with warnings for invalid chunk flags, allowing old versions to continue running when encountering unknown flags.

## Related Issues
#1671

## Change Details
Replace strict error with warning and use atomic
operations to avoid duplicate warnings for same flag values.

## Test Results

## Change Type
- [ ] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/rafs/src/metadata/layout/v5.rs
+++ b/rafs/src/metadata/layout/v5.rs
@@ -653,7 +653,17 @@ impl RafsV5BlobTable {
                         return Err(einval!());
                     }
                     let entry = &self.extended.entries[index];
-                    let blob_features = BlobFeatures::from_bits(entry.features).ok_or_else(|| einval!("invalid blob feature flags"))?;
+                    let blob_features = match BlobFeatures::try_from(entry.features) {
+                        Ok(f) => f,
+                        Err(_) => {
+                            let known = BlobFeatures::from_bits_truncate(entry.features);
+                            let unknown = entry.features & !known.bits();
+                            if unknown != 0 {
+                                warn!("Unknown blob feature flags 0x{:x}, unknown bits: 0x{:x}, truncating.", entry.features, unknown);
+                            }
+                            known
+                        }
+                    };
                     (entry.chunk_count, entry.uncompressed_size, entry.compressed_size, blob_features)
                 } else {
                     (0, 0, 0, BlobFeatures::_V5_NO_EXT_BLOB_TABLE)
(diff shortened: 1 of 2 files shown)


BraveY:
Thanks for the PR. Please sign the commit. Do we test the compatibility for this PR?

storage/src/meta/chunk_info_v2.rs
```diff
@@ -251,10 +254,14 @@ impl BlobMetaChunkInfo for BlobChunkInfoV2Ondisk {
 
         let invalid_flags = self.check_flags();
         if invalid_flags != 0 {
-            return Err(Error::new(
-                ErrorKind::Other,
-                format!("unknown chunk flags 0x{:x}", invalid_flags),
-            ));
+            let current = LAST_WARNED_FLAGS.load(Ordering::Relaxed);
```
imeoer, line 257: Thanks for the PR, why we need `LAST_WARNED_FLAGS` here?

rafs/src/metadata/layout/v5.rs
```diff
@@ -653,7 +653,7 @@ impl RafsV5BlobTable {
                         return Err(einval!());
                     }
                     let entry = &self.extended.entries[index];
-                    let blob_features = BlobFeatures::from_bits(entry.features).ok_or_else(|| einval!("invalid blob feature flags"))?;
+                    let blob_features = BlobFeatures::from_bits_truncate(entry.features);
```
BraveY: Could we use `BlobFeatures::try_from()` at [device.rs](https://github.com/dragonflyoss/nydus/blob/340b2cb1705a3c7b7b558ff464721c1135976b08/storage/src/device.rs#L102) to check the flag including new flag? 
Or add a warning that detected an unknown flag. I'm a bit worried that ignore unknown flag directly would cause some error. So maybe we could add some warning logs.

storage/src/meta/chunk_info_v2.rs
sidneychang: Thanks for the question!
`LAST_WARNED_FLAGS` is introduced to prevent log flooding caused by repeated warnings for the same unknown flag values.

**Example:**
When reading a  blob with 5 chunks, the code calls `validate()` 5 times (once per chunk).
If all chunks share the same unknown flag (e.g. `0x20`), we would log 5 identical warnings.
When an operation involves a large number of chunks, this repetition can quickly accumulate into a large number of duplicate warning messages.

**Call chain:**
`User IO → get_chunks_compressed() → get_chunk_entry() → validate()` (per chunk)

Since unknown flags are typically consistent within a blob (and stable for its lifetime), warning once per unique flag value globally is sufficient.
Using an `AtomicU8` to track the last warned flags adds negligible overhead compared to the cost of repeated logging.

BraveY (approved):
Thanks for the PR. LGTM

sidneychang:
> Please sign the commit.

Thanks! I've already signed the commit.

PR #1822, opened 2026-01-04
action: fix up macos release

bergwolf:
* macos-13 has been deprecated in github action
* cross does not really work on macos
* add missing nydusctl binary


bergwolf:
cargo-deny error is fixed in https://github.com/dragonflyoss/nydus/pull/1823

BraveY (approved):
Thanks, LGTM

PR #1824, opened 2026-01-06
fix: ensure chunk bitmap changes are synced to disk

gabrielstedman:
## Overview

When containers run with lazy loading and the registry becomes unavailable (e.g., in an offline scenario), the system must rely on the cached chunk state from previous runs. However, bitmap changes marking chunks as ready were not being explicitly synced to disk, causing them to be lost on process restart.

## Root Cause

The issue stems from chunk ready state not being durably persisted to disk:

1. **During First Run (Lazy Loading):**
   - Only accessed file chunks are fetched from registry and cached
   - Chunks marked as ready update the memory-mapped bitmap file
   - Changes are written to the mmap'd memory but NOT explicitly synced to disk

2. **On Process Restart:**
   - Cache subsystem reads chunk bitmap from disk
   - Bitmap shows stale/incomplete state (missing synced updates)
   - System thinks chunks that were actually cached are still missing
   - Attempts to fetch missing chunks from registry
   - In offline scenario: registry unreachable → containers fail to start with "binary not found" errors

## Solution

The fix ensures chunk ready state is durably persisted.

-. **Always open bitmap file with write permission** 
   - Changed from `.write(create)` to `.write(true)`
   - Ensures file can be synced on reopened instances

- **Sync bitmap after each chunk ready state change**

This ensures chunk ready state survives process restarts and enables reliable lazy loading to survive offline scenarios by re-reading from cache.
       -        -        - 
## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/storage/src/cache/state/persist_map.rs
+++ b/storage/src/cache/state/persist_map.rs
@@ -236,8 +236,7 @@ impl PersistMap {
             }
 
             if self.write_u8(index, current) {
-                // Sync the bitmap change to disk immediately to ensure it persists
-                let _ = self.filemap.sync_data();
+                let _ = self.filemap.sync_data_async();
                 if self.not_ready_count.fetch_sub(1, Ordering::AcqRel) == 1 {
                     self.mark_all_ready();
                 }
@@ -249,6 +248,7 @@ impl PersistMap {
     }
 
     fn mark_all_ready(&self) {
+        // Use blocking sync for final completion to ensure all data is written to disk
         let _ = self.filemap.sync_data();
     }
 
--- a/utils/src/filemap.rs
+++ b/utils/src/filemap.rs
@@ -188,6 +188,22 @@ impl FileMapState {
         std::mem::forget(file);
         result
     }
+
+    /// Schedule an asynchronous sync of mapped file data to disk.
+    pub fn sync_data_async(&self) -> Result<()> {
+        let ret = unsafe {
+            libc::msync(
+                self.base as *mut libc::c_void,
+                self.size,
+                libc::MS_ASYNC,
+            )
+        };
+        if ret != 0 {
+            Err(last_error!("msync async failed"))
+        } else {
+            Ok(())
+        }
+    }
 }
 
 /// Duplicate a file object by `libc::dup()`.


imeoer:
Hi @gabrielstedman, thanks for the PR, the kernel should sync the data written via mmap to disk at an appropriate time after the nydusd process exits, is it still necessary to manually call msync?

gabrielstedman:
@imeoer I've changed this to explicitly invoke `let _ = self.filemap.sync_data();` instead of msync.

In my testing I loaded 16 containers using erofs with lazy pulling.
Then restart the node and cut WAN access
Every time data on disk was stale and none of the containers would come up (when registry isn't accessible) unless I explicitly use "prefetch_all" or I re-enable WAN access.

I feel that the bitmap file tracks critical state (which chunks are cached). Unlike application data where eventual consistency is acceptable, this metadata should be durable immediately.
After applying this fix all 16 containers are able to start after a node reboot without WAN access.

bergwolf:
@gabrielstedman Are you testing erofs lazy loading with the fscache driver? The bitmap consistency is supposed to be guaranteed by the file's page cache that is shared between nydusd and the kernel. It looks like erofs is not relying on the file page cache for bitmap file consistency?

What's more, `self.filemap.sync_data` really hurts performance for each and every bitmap change.

gabrielstedman:
I swapped this to use async msync instead per chunk write. In my testing bitmap consistency wasn't being guaranteed and data was stale.
Happy to explore additional testing or alternative mechanisms.

gabrielstedman:
This is what I'm testing:

nydus-config.json:
```json
{
  "type": "bootstrap",
  "domain_id": "shared-nydus-cache",
  "config": {
    "backend_type": "registry",
    "backend_config": {
      "scheme": "https"
    },
    "cache_type": "fscache"
  }
}
```
```bash
~# lsmod | grep cachefiles
cachefiles            172032  56
netfs                 589824  2 erofs,cachefiles
```

nydus-snapshotter config.toml:
```toml
version = 1
root = "/var/lib/containerd/io.containerd.snapshotter.v1.nydus"
address = "/run/containerd-nydus/containerd-nydus-grpc.sock"
daemon_mode = "shared"
cleanup_on_close = true
enable_referrer_detect = true

[daemon]
nydusd_path = "/bin/nydusd"
nydusimage_path = "/usr/bin/nydus-image"

fs_driver = "fscache"
recover_policy = "restart"
threads_number = 0

[snapshot]
enable_nydus_overlayfs = false
sync_remove = true

[cache_manager]
disable = false
gc_period = "24h"
cache_dir = ""

[log]
level = "info"
log_to_stdout = true
```
nydus snapshotter runs as:
```bash
/usr/bin/containerd-nydus-grpc --config/etc/nydus/config.toml --nydusd-config /etc/nydus/nydusd-config.json --log-to-stdout
```
My test:
- With registry up run 17 nydus-format containers started by kubelet. All of them start correctly (lazy pulling)
- Shut down node.
- Block access to registry
- Boot node
Containers will not come up as nydus tries to access registry for pulling chunks (but they should be picked up from cache).

If I use this for nydusd config then the issue doesn't occur:
```json
    "prefetch_config": {
      "enable": true,
      "threads_count": 3,
      "merging_size": 131072,
      "bandwidth_rate": 10485760,
      "prefetch_all": true
    }
```

- With this patch and repeating the original steps the issue doesn't occur and nydusd finds chunks in cache.

imeoer:
The blob cache file is managed by the kernel fscache driver, while the chunk map file is handled by nydusd. The issue likely stems from the "Shut down node" step in your test, which prevented timely msync execution.

But the nydus `prefetch all` can't ensure all chunks are fully ready (for example: network issues caused some chunks to be incompletely prefetched).

gabrielstedman:
> The blob cache file is managed by the kernel fscache driver, while the chunk map file is handled by nydusd. The issue likely stems from the "Shut down node" step in your test, which prevented timely msync execution.
> 
> But the nydus `prefetch all` can't ensure all chunks are fully ready (for example: network issues caused some chunks to be incompletely prefetched).

I will point out that I did allow plenty of time between containers running and node shutting down - so it would seem like automatic msync is not occurring in my case.

Feel free to close this issue if you feel the changes would have a performance impact but I'm currently not so sure the automatic msync is occurring at all.

bergwolf:
@gabrielstedman As I pointed out earlier, the design is to use the bitmap file's page cache to coordinate between the kernel and use space. We do not rely on the bitmap to be synced to disk just to let erofs know that a chunk is locally available.  The issue needs more investigation from the kernel side, rather than trying to workaround it with an extra sync on every bit written, which can hurt general IO performance for both erofs and FUSE cases.

OTOH, given that `erofs` + `fscache` is being deprecated by the kernel, I'm not sure if it is even worth investigating from the kernel developer's view.

PR #1864, opened 2026-01-28
api: support hot reload of registry authentication

imeoer:
## Overview
Add HTTP API endpoints to update and query daemon configuration at
runtime without restarting nydusd. This is particularly useful for
refreshing registry credentials when tokens expire.

Key changes:
- Add GET/PUT /api/v1/config endpoints for configuration management
- Add config module in utils for centralized configuration updates
- Add smoke test cases for hot reload configuration API
- Update documentation with hot reload usage examples

## Related Issues
https://github.com/containerd/nydus-snapshotter/issues/690

## Change Details

## Test Results

## Change Type
- [ ] Bug Fix
- [x] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -191,13 +191,12 @@ impl fmt::Display for Scheme {
 }
 
 struct RegistryState {
+    id: String,
     // HTTP scheme like: https, http
     scheme: Scheme,
     host: String,
     // Image repo name like: library/ubuntu
     repo: String,
-    // Base64 encoded registry auth
-    auth: Option<String>,
     // Retry limit for read operation
     retry_limit: u8,
     // Scheme specified for blob server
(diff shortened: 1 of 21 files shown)


Zephyrcf (approved):
LGTM!

storage/src/backend/registry.rs
BraveY: It looks like the id parameter now represents a mountpoint ID rather than a blob ID.

BraveY, line 893: Because this id is no longer blobdid, the corresponding previous read and write statistics for blobid level no longer exist, this behavior change is very large, I think we need to consider adding a mountpoint id as a parameter, instead of directly changing the original blobid to mountpointid.

BraveY (changes requested):

storage/src/backend/registry.rs
imeoer: Thanks, let me check this, the blob ID seems to be useless, it should be used as a mountpoint.

BraveY (approved):
LGTM

PR #1873, opened 2026-02-04
CI: make codecov include both UT and smoke test code coverage

bergwolf:
Right now codecov only counts UT code coverage. Let's switch to the instrument-coverage so that smoke coverage can also be counted.

CI example can be found at: https://github.com/bergwolf/nydus/actions/runs/21663517010


bergwolf:
Somehow getting `Failing after 1s — 48.36% (-8.28%) compared to e2afc59` while no actual code was introduced. Investigating... Marking WIP in the meanwhile.

bergwolf:
Now the report result looks good:)

image

BraveY (approved):
Great work! LGTM

PR #1877, opened 2026-02-27
Update fuse-backend-rs and related rust-vmm dependencies

bergwolf:
1. Update fuse-backend-rs and related rust-vmm dependencies
2. One key change is that virtiofs DAX is removed.
3. Revert the profraw approach as it constantly causes ENOSPC error on github action


BraveY (approved):
LGTM! Thanks

PR #1894, opened 2026-03-15
Blobcache gc

rohanphadnis-thunder:
## Overview
This PR implements the `GET /api/v2/blobs?domain_id=X&blob_id=Y` API call in `nydusd`.

**The goal of this PR is to access the ref counters, allowing for safe garbage collection in the blobcache.**

## Related Issues
No related issues but this method was marked as a `todo!()`.

## Change Details
We define `BlobCacheInfoEntry` and `BlobCacheInfoList` structs. We implement a method handler and expose it to the `nyduysd` api.

## Test Results
N/A

## Change Type
- [x] Add Unit Tests
- [x] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/api/src/http.rs
+++ b/api/src/http.rs
@@ -68,6 +68,26 @@ pub struct BlobCacheObjectId {
     pub blob_id: String,
 }
 
+/// Information about a cached blob object returned by the GetBlobObject API.
+#[derive(Debug, Default, Serialize)]
+pub struct BlobCacheInfoEntry {
+    /// Type of blob object: "bootstrap" or "datablob".
+    pub blob_type: String,
+    /// Blob identifier.
+    pub blob_id: String,
+    /// Domain identifier.
+    pub domain_id: String,
+    /// Reference count for data blobs (number of active users).
+    pub ref_count: u32,
+}
+
+/// Response for the GetBlobObject API containing a list of blob cache entries.
+#[derive(Debug, Default, Serialize)]
+pub struct BlobCacheInfoList {
+    /// List of blob cache information entries.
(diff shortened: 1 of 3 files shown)


imeoer:
Hi @rohanphadnis-thunder, can we directly use the `underlying_files` field from `GET /api/v1/daemon`? The blob reference counting should be handled by the upper-level manager, such as the nydus-snapshotter, which can call APIs of all managed nydusds to calculate the blob reference counts.

BraveY:
Hi, Thanks for the PR.
The ref_count exposed by this API is **inaccurate for non-singleton modes**.
Currently, `BlobCacheMgr` is shared across all bootstraps only in **singleton mode** (fscache). In this mode, the ref_count correctly reflects how many bootstraps reference a data blob within that single nydusd process.
However, in **fusedev/virtiofs mode**, each container/image gets its own nydusd process, and each process has its own isolated `BlobCacheMgr`. The API returns ref_count from `DAEMON_CONTROLLER.get_blob_cache_mgr()`, which only sees bootstraps managed by that specific daemon process.
This means:
- **Singleton mode**: ref_count = all references within this nydusd ✓
- **Fusedev mode**: ref_count = only references from this container's bootstrap ✗
For accurate GC at the system level, the upper layer (e.g., nydus-snapshotter) would need to aggregate ref_counts across all nydusd instances, which this API doesn't support.

So. I agree with Imeoer. Using the underlying_files field from GET /api/v1/daemon is a good idea.

PR #1784, opened 2025-10-13
[registry] Disable token refresh thread when prefetch is enabled

Fricounet:
## Overview
_Please briefly describe the changes your pull request makes._

Add a new RegistryConfig field called `enable_token_refresh` that will decide whether the registry refresh token thread is started on not. 

## Related Issues
_Please link to the relevant issue. For example: `Fix #123` or `Related #456`._

Fix #1777

## Change Details
_Please describe your changes in detail:_

The new field will be set to true by default but will be disabled when prefetch is enabled. The refresh thread is disabled in this case because it becomes useless once everything has been prefetched and it's unlikely that tokens would expire during the time it takes for the prefetch to run (as it would mean that regular image pulls  would be likely to fail as well because they download the layers in a similar time to the prefetch)

## Test Results
_If you have any relevant screenshots or videos that can help illustrate your changes, please add them here._

Running tests with this basic nydusd config:
```
{
  "device": {
    "backend": {
      "type": "registry",
      "config": {
        "timeout": 15,
        "connect_timeout": 15,
        "retry_limit": 5
      }
    },
    "cache": {
      "type": "blobcache",
      "config": {
        "work_dir": "cache"
      }
    }
  },
  "mode": "direct",
  "digest_validate": false,
  "iostats_files": false,
  "enable_xattr": true,
  "fs_prefetch": {
    "enable": true,
    "threads_count": 4
  }
}
```

The token thread isn't started:
```
INFO[2025-10-13T11:38:56.430857078+02:00] nydusd command: /containerd-local/nydusd fuse --config /containerd-local/io.containerd.snapshotter.v1.nydus/config/d3mchc1jjoq9cr38fp00/config.json --bootstrap /containerd-local/io.containerd.snapshotter.v1.nydus/snapshots/24/fs/image.boot --mountpoint /containerd-local/io.containerd.snapshotter.v1.nydus/snapshots/24/mnt --apisock /containerd-local/io.containerd.snapshotter.v1.nydus/socket/d3mchc1jjoq9cr38fp00/api.sock --log-level debug --log-rotation-size 100
[2025-10-13 11:38:56.437379 +02:00] INFO Program Version: v2.3.7-dd, Git Commit: "c100686d6b0243c451f1a261b900497f827c90aa", Build Time: "2025-10-13T09:38:39.942636660Z", Profile: "debug", Rustc Version: "rustc 1.84.0 (9fc6b4312 2025-01-07)"
[...]

PR #1913, opened 2026-04-03
registry: move config auth refresh to background thread

### LINKED ISSUE #1893: [Bug] Hot reload config API does not proactively refresh cached auth
### Problem Description

https://github.com/dragonflyoss/nydus/pull/1864 added GET/PUT /api/v1/config endpoints to support hot reload of registry authentication. While the global `CONFIG_MAP` is correctly updated on PUT, the new credentials are not always used by subsequent registry requests due to the `cached_auth` layer.

### Expected Behavior

I think a PUT to `/api/v1/config` updating `registry_auth` should cause subsequent registry requests to use the new credentials without waiting for a 401 failure.

### Actual Behavior

`Registry::request()` checks `cached_auth first` (local cache to RegistryState). If it contains a value, it is used directly as the Authorization header without consulting `CONFIG_MAP`. Updated credentials in `CONFIG_MAP` are only picked up when:

1. A request fails with 401, the re-auth flow in `request()` calls `get_auth_header()`, which reads from `CONFIG_MAP`. On success, `cached_auth` is updated.
2. The refresh token thread fires (bearer auth only) -- it calls `get_token() -> fetch_token()`, which reads from `CONFIG_MAP`. However, this thread only activates when `token_expired_at` is set, which requires a prior successful call to get_token().

Scenarios where hot reload does not work as expected
1. Bearer auth, daemon started with invalid/missing credentials: get_token() is never called successfully, so token_expired_at and cached_bearer_auth are never set. The refresh thread loops but does nothing. New credentials via PUT are only used when a FUSE request triggers a 401.
2. Basic auth (no bearer token exchange): The refresh thread is irrelevant. It only handles bearer tokens. cached_auth holds the Basic ... header directly. Updated credentials via PUT are only used after a 401.

### How to reproduce

1. Start nydus with valid creds
2. Call `PUT /api/v1/config` with bogus creds. For instance with foo:bar: `sudo curl --unix-socket /containerd-local/io.containerd.snapshotter.v1.nydus/socket/d6psqh1jjoqbddk18p2g/api.sock -X PUT "http://localhost/api/v1/config?id=%2F" --data '{"registry_auth": "Zm9vOmJhcg=="}'`
3. Try to read a file not yet fetched from the registry. The read will succeed thanks to the cached creds, meaning the bogus creds weren't used

### Environment Details

- Nydus-snapshotter version: N/A
- Nydus version: built from https://github.com/dragonflyoss/nydus/commit/beba83c4973c46a0723989501591367eabbe4671
- Container runtime:
- Operating System:
- Kernel version:

### Additional Information

_No response_

Fricounet:
## Overview

Move `refresh_cached_auth_from_config` from the hot request path to the background refresh token thread, and remove the `disable_token_refresh` workaround that was disabling the thread when prefetch was enabled.

## Related Issues

Fixes #1893

## Change Details

**Move config auth refresh to background thread:** `refresh_cached_auth_from_config` was called at the top of every `request()`. When multiple concurrent requests detect a config change simultaneously, each one independently calls `get_token()`, potentially causing a thundering herd on the auth server. Additionally, performing the refresh in the request path adds latency to the first request after a config update.

The refresh is now performed in the existing background thread, polling every 5 seconds for update in the config.

**Remove `disable_token_refresh`:** This was introduced in #1784 to silence noisy 401/403 errors from the refresh thread when using expired tokens with prefetch enabled. With config auth changes now handled properly by the thread, the workaround is no longer needed. Always starting the thread ensures config auth updates via the API are picked up promptly regardless of prefetch configuration.

## Test Results

Existing unit tests pass. 

Tested manually using an AWS private ECR:
```
DEBU[2026-04-03T17:17:25.689707264+02:00] Trying to get credentials from kubelet        ref="111111111111.dkr.ecr.us-east-1.amazonaws.com/datadog-agent:7.75.0-rc.7-jmx-zstd-nydus"
INFO[2026-04-03T17:17:30.773328562+02:00] Got credentials from provider kubelet         ref="111111111111.dkr.ecr.us-east-1.amazonaws.com/datadog-agent:7.75.0-rc.7-jmx-zstd-nydus"
DEBU[2026-04-03T17:17:30.773347970+02:00] adding credential entry to store              ref="111111111111.dkr.ecr.us-east-1.amazonaws.com/datadog-agent:7.75.0-rc.7-jmx-zstd-nydus"
[2026-04-03 17:17:30.773598 +02:00] DEBUG [/src/http_handler.rs:186] <--- Put Uri { string: "/api/v1/config?id=%2F" }
[2026-04-03 17:17:30.773682 +02:00] DEBUG [/src/http_handler.rs:191] ---> Put Status Code: NoContent, Elapse: Ok(88.86µs), Body Size: 0
[2026-04-03 17:17:33.628709 +02:00] DEBUG [/src/backend/registry.rs:331] refreshed basic registry auth after registry_auth config update
```

## Change Type

- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [x] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/api/src/config.rs
+++ b/api/src/config.rs
@@ -623,9 +623,6 @@ pub struct RegistryConfig {
     /// Enable HTTP proxy for the read request.
     #[serde(default)]
     pub proxy: ProxyConfig,
-    /// Disable background token refresh thread. Defaults to false.
-    #[serde(skip_deserializing)]
-    pub disable_token_refresh: bool,
 }
 
 /// Configuration information for blob cache manager.
@@ -1364,7 +1361,7 @@ impl TryFrom<RafsConfig> for ConfigV2 {
     type Error = std::io::Error;
 
     fn try_from(v: RafsConfig) -> std::result::Result<Self, Self::Error> {
-        let mut backend: BackendConfigV2 = (&v.device.backend).try_into()?;
+        let backend: BackendConfigV2 = (&v.device.backend).try_into()?;
         let mut cache: CacheConfigV2 = (&v.device.cache).try_into()?;
         let rafs = RafsConfigV2 {
             mode: v.mode,
@@ -1378,13 +1375,6 @@ impl TryFrom<RafsConfig> for ConfigV2 {
         };
         if !cache.prefetch.enable && rafs.prefetch.enable {
             cache.prefetch = rafs.prefetch.clone();
-        }
-
-        // If prefetch is enabled, disable token refresh by default
-        if cache.prefetch.enable {
-            if let Some(registry) = backend.registry.as_mut() {
-                registry.disable_token_refresh = true;
-            }
         }
 
         Ok(ConfigV2 {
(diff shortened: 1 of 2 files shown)


imeoer (approved):
Thanks, LGTM!

PR #1915, opened 2026-04-07
Disable HTTPS-to-HTTP fallback unless skip_verify is set

### LINKED ISSUE #1910: [Feature] Add option to disable HTTPS-to-HTTP fallback in registry backend
### Feature Description

Add a `skip_http_fallback` configuration option to `RegistryConfig` that prevents the automatic HTTPS-to-HTTP downgrade on TLS errors. When set to `true`, TLS errors propagate as-is instead of silently falling back to plaintext HTTP. Defaults to `false` to preserve backward compatibility.

Example configuration:

```json
{
  "backend": {
    "type": "registry",
    "config": {
      "scheme": "https",
      "host": "registry.example.com",
      "repo": "library/ubuntu",
      "skip_http_fallback": true
    }
  }
}
```

Additionally, a similar change would need to be made in the nydus-snapshotter:
- to support propagating this option through to nydusd configuration.
- to also disable the fallback at the snapshotter level since it's also making calls to the registry

### Problem and Use Case

When the registry backend encounters TLS errors (wrong version number, connection refused, SSL-related), it automatically downgrades from HTTPS to HTTP via `needs_fallback_http` / `fallback_http`. While convenient for development, this poses a security risk in production environments where plaintext HTTP should never be used. Image pull traffic can be exposed to eavesdropping or tampering.

Operators running hardened deployments need a way to enforce HTTPS-only communication with the registry.

### Related issues

_No response_

- [X] Yes I am willing to submit a PR!

Fricounet:
## Overview

When `skip_verify` is `false` (the default), TLS errors now propagate as-is instead of silently falling back to plaintext HTTP. The fallback is still available when `skip_verify` is `true`, aligning with nerdctl's `--insecure-registry` semantics.

## Related Issues

Fix #1910

## Change Details

- Added a `skip_verify` field to `RegistryState`, populated from `RegistryConfig.skip_verify`.
- `needs_fallback_http()` now returns `false` early when `skip_verify` is `false`, preventing any HTTPS-to-HTTP downgrade.
- When `skip_verify` is `true`, the existing fallback behavior is preserved unchanged.

## Test Results

- `test_no_fallback_http_by_default` -- verifies TLS errors are never swallowed when `skip_verify` is `false`.
- `test_scheme_and_fallback_http` -- verifies fallback still works when `skip_verify` is `true`.
- Full `nydus-storage` test suite passes.

## Change Type

- [x] Feature Addition

## Self-Checklist

### DIFF
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -205,6 +205,9 @@ struct RegistryState {
     repo: String,
     // Retry limit for read operation
     retry_limit: u8,
+    // When true, skip TLS certificate verification AND allow HTTPS-to-HTTP fallback.
+    // When false (default), TLS errors propagate as-is without falling back to HTTP.
+    skip_verify: bool,
     // Scheme specified for blob server
     blob_url_scheme: String,
     // Replace registry redirected url host with the given host
@@ -246,6 +249,9 @@ impl RegistryState {
     }
 
     fn needs_fallback_http(&self, e: &dyn Error) -> bool {
+        if !self.skip_verify {
+            return false;
+        }
         match e.source() {
             Some(err) => match err.source() {
                 Some(err) => {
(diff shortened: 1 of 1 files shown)


imeoer (approved):
Thanks, LGTM!

PR #1922, opened 2026-04-13
Fix token refresh firing every 20s instead of near expiry

Fricounet:
## Overview
Fix the token refresh thread in the registry backend so it correctly refreshes tokens near expiry rather than every 20 seconds.

## Related Issues
Regression introduced by https://github.com/dragonflyoss/nydus/pull/1913.

## Change Details
The `refresh_interval` variable was set to `expires_in - 20` after each refresh, then used in the check `now + refresh_interval >= expired_at`. Since `expired_at = fetch_time + expires_in`, this resolved to "refresh when 20s have passed since last fetch" rather than "refresh 20s before expiry".

Replace the mutable `refresh_interval` with a fixed `REGISTRY_TOKEN_REFRESH_MARGIN` constant so the check correctly triggers only when the token is about to expire within the next 20 seconds.

## Test Results
N/A

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -34,6 +34,9 @@ const HEADER_WWW_AUTHENTICATE: &str = "www-authenticate";
 const REGISTRY_DEFAULT_TOKEN_EXPIRATION: u64 = 10 * 60; // in seconds
 const REGISTRY_CONFIG_POLL_INTERVAL: u64 = 5; // in seconds
 
+// Refresh tokens this many seconds before they expire to avoid using an expired token.
+const REGISTRY_TOKEN_REFRESH_MARGIN: u64 = 20; // in seconds
+
 /// Error codes related to registry storage backend operations.
 #[derive(Debug)]
 pub enum RegistryError {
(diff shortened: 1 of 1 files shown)


storage/src/backend/registry.rs
```diff
@@ -33,6 +33,8 @@ const HEADER_WWW_AUTHENTICATE: &str = "www-authenticate";
 
 const REGISTRY_DEFAULT_TOKEN_EXPIRATION: u64 = 10 * 60; // in seconds
 const REGISTRY_CONFIG_POLL_INTERVAL: u64 = 5; // in seconds
+                                              // Refresh tokens this many seconds before they expire to avoid using an expired token.
```
BraveY: This comment should be moved to the beginning of the line

BraveY (commented):
Please sign off the commit. Others LGTM

storage/src/backend/registry.rs
Fricounet: seems like `cargo fmt` wanted to align that with the `// in seconds` above, fixed

BraveY (approved):
LGTM

PR #1929, opened 2026-04-26
fusedev: fix remount invalidation for nested mountpoints

### LINKED ISSUE #1928: [Bug] Remounting a nested FUSE mountpoint returns ENOENT until its parent is read
### Problem Description

When `nydusd` runs in FUSE mode with a shared mount root, unmounting and then remounting an image at the same mountpoint leaves the remounted path temporarily inaccessible. Direct path access returns `ENOENT` immediately after remount, but the path becomes visible again after reading its direct parent directory.

This breaks remount correctness for nested mountpoints and suggests stale FUSE dentry state across `umount` and `mount`.

### Expected Behavior

After `umount` + `mount` of the same Nydus mountpoint, direct access to the remounted path should succeed immediately, without requiring a parent directory scan or any extra refresh operation.

### Actual Behavior

After remount, `stat(<mountpoint>)` returns `ENOENT` even though the mount request has completed. Running `ls -l <parent>` makes the same `stat(<mountpoint>)` succeed again.

In this repro, `ls -l <parent>` likely makes the path visible again because it triggers `readdirplus` on the parent directory, which refreshes the kernel's view of the `parent/name` entry.

### How to reproduce

1. Start `nydusd` in FUSE mode with a shared mount root, for example `--mountpoint /data/nydus-images`.
2. Mount an image through the HTTP API to a nested mountpoint such as `/repo/ns/project/image_tag`.
3. Access the mounted path once:
   `stat /data/nydus-images/repo/ns/project/image_tag`
4. Unmount that image through the HTTP API.
5. Mount the same image again to the same mountpoint.
6. Immediately run:
   `stat /data/nydus-images/repo/ns/project/image_tag`
   and observe `ENOENT`.
7. Run:
   `ls -l /data/nydus-images/repo/ns/project`
   and then run the same `stat` again; it succeeds.

### Environment Details

- Nydus version: `v2.4.2`
- Operating System: Ubuntu 16.04 LTS
- Kernel version: Linux 5.4.0-48-generic

### Additional Information

After tracing the FUSE invalidation path, I believe the issue is in `FsService::umount()` in `service/src/fs_service.rs`. For the mountpoint entry itself, the invalidation should target the direct parent inode together with the mountpoint basename, rather than `ROOT_ID` together with the full mountpoint path.

libfuse defines entry invalidation in terms of a direct `parent/name` pair:

> ```c
> int fuse_lowlevel_notify_inval_entry(struct fuse_session *se,
>                                      fuse_ino_t parent,
>                                      const char *name,
>                                      size_t namelen);
> ```
>
> **Notify to invalidate parent attributes and the dentry matching parent/name.**
>
> Parameters
> - `se`: the session object
> - `parent`: inode number
> - `name`: file name
> - `namelen`: `strlen()` of file name
>
> Returns
> - `0` for success, `-errno` for failure

<https://libfuse.github.io/doxygen/fuse__lowlevel_8h.html#ab14032b74b0a57a2b3155dd6ba8d6095>

Nydus eventually issues the same kind of notification through `notify_inval_entry(parent, name)`:

```rust
pub fn notify_inval_entry<S: BitmapSlice>(
    &self,
    mut w: FuseDevWriter<'_, S>,
    parent: u64,
    name: &std::ffi::CStr,
) -> Result<usize>
```

and `walk_and_notify_invalidation()` calls it as:

```rust
self.server
    .notify_inval_entry(writer, parent_kernel_ino, cstr_name.as_c_str())
```

In `v2.4.2`, however, `FsService::umount()` starts the invalidation walk from `ROOT_ID` and passes the full mountpoint path string:

<https://github.com/dragonflyoss/nydus/blob/v2.4.2/service/src/fs_service.rs#L187-L195>

```rust
self.walk_and_notify_invalidation(
    ROOT_ID,
    cmd.mountpoint.trim_start_matches('/'),
    root_ino,
    fs_idx,
)
```

For a nested mountpoint such as `/a/b/c`, this means issuing an invalidation equivalent to `(ROOT_ID, "a/b/c")`, which is not equivalent to invalidating the direct dentry `(parent of /a/b/c, "c")`. That mismatch matches the observed behavior after remount.

I also ran a comparison between the affected `v2.4.2` binary and a fixed build using a minimal `/a/b/c` mountpoint. For the remount step, I sent the same HTTP API requests as above, but with `mountpoint=/a/b/c`.

Observed with affected `v2.4.2`:

```text
$ stat -c '%n: %F' /tmp/nydus-v242-mnt/a/b/c
/tmp/nydus-v242-mnt/a/b/c: directory

$ curl --unix-socket /tmp/nydus-v242.sock -X DELETE \
    'http://localhost/api/v1/mount?mountpoint=/a/b/c'
$ curl --unix-socket /tmp/nydus-v242.sock -X POST \
    'http://localhost/api/v1/mount?mountpoint=/a/b/c' \
    -H 'Content-Type: application/json' \
    -d '{...}'

$ stat -c '%n: %F' /tmp/nydus-v242-mnt/a/b/c
stat: cannot stat '/tmp/nydus-v242-mnt/a/b/c': No such file or directory

$ ls -l /tmp/nydus-v242-mnt/a/b >/dev/null
$ stat -c '%n: %F' /tmp/nydus-v242-mnt/a/b/c
/tmp/nydus-v242-mnt/a/b/c: directory

$ grep -m1 'notify_inval_entry failed: a/b/c' /tmp/nydus-v242.log
[2026-04-24 17:28:19.536409 +08:00] WARN [/src/fusedev.rs:385] notify_inval_entry failed: a/b/c InvalidMessage(...)
```

Observed with the fixed build:

```text
$ stat -c '%n: %F' /tmp/nydus-new_fix-mnt/a/b/c
/tmp/nydus-new_fix-mnt/a/b/c: directory

$ curl --unix-socket /tmp/nydus-new_fix.sock -X DELETE \
    'http://localhost/api/v1/mount?mountpoint=/a/b/c'
$ curl --unix-socket /tmp/nydus-new_fix.sock -X POST \
    'http://localhost/api/v1/mount?mountpoint=/a/b/c' \
    -H 'Content-Type: application/json' \
    -d '{...}'

$ stat -c '%n: %F' /tmp/nydus-new_fix-mnt/a/b/c
/tmp/nydus-new_fix-mnt/a/b/c: directory

$ grep -m1 'notify_inval_entry failed: a/b/c' /tmp/nydus-new_fix_rCURRENT.log
# no match
```

With the fixed build, the `notify_inval_entry failed: a/b/c` warning no longer appears, and `stat` on the remounted path succeeds immediately without requiring `ls -l <parent>`.

tedcy:
## Overview

Fix FUSE cache invalidation when unmounting a RAFS backend mounted at a subdirectory path.

Before this change, `FsService::umount()` invalidated the mountpoint root entry by passing `ROOT_ID` and the full mountpoint path to `notify_inval_entry()`. For a mountpoint such as `/a/b/c`, this produced an invalidation request equivalent to `(ROOT_ID, "a/b/c")`. FUSE entry invalidation expects a direct `parent/name` pair, so the remounted path could remain stale and `stat(<mountpoint>)` could return `ENOENT` until the parent directory was read.

## Related Issues

Fix #1928 

## Change Details

For non-root mountpoints, this PR resolves the mountpoint pseudo inode, finds its direct parent inode, and passes the mountpoint basename to `walk_and_notify_invalidation()`. This makes the entry invalidation target match the FUSE `parent/name` semantics.

For the root mountpoint `/`, this PR keeps the previous behavior by using `ROOT_ID` and an empty entry name. This avoids changing the existing `/` mount/unmount behavior while fixing subdirectory mountpoints.

## Test Results

Code checks:

```text
$ cargo fmt --check --package nydus-service
$ cargo check -p nydus-service
```

Workspace unit tests were also run with host-environment-dependent cases skipped on the validation host:

```text
Starting 832 tests across 13 binaries (5 tests skipped)
Summary [  23.696s] 832 tests run: 832 passed, 5 skipped
```

The skipped cases depend on host capabilities or system configuration that were not available in that environment; all executed unit tests passed.

Added a smoke test for the nested mountpoint remount case, then ran it against both the old invalidation target and the fixed build:

```text
Bad build (old invalidation target):
mount nested RAFS backend at /a/b/c
stat nested mountpoint after first mount: <workdir>/mnt/a/b/c
unmount nested mountpoint /a/b/c
remount nested RAFS backend at /a/b/c
stat nested mountpoint immediately after remount without reading parent directory: <workdir>/mnt/a/b/c
Error: Received unexpected error:
  stat <workdir>/mnt/a/b/c: no such file or directory
--- FAIL: TestAPI/TestNestedMountpointStatAfterRemount

Fixed build (direct parent/name invalidation):
mount nested RAFS backend at /a/b/c
stat nested mountpoint after first mount: <workdir>/mnt/a/b/c
unmount nested mountpoint /a/b/c
remount nested RAFS backend at /a/b/c
stat nested mountpoint immediately after remount without reading parent directory: <workdir>/mnt/a/b/c
--- PASS: TestAPI/TestNestedMountpointStatAfterRemount
PASS
```

The bad build was produced by temporarily reverting the invalidation target back to `ROOT_ID + full mountpoint path`. With that behavior, the same smoke test reproduces the remount failure. With this PR applied, the test passes by directly statting `/a/b/c` after remount without reading the parent directory first.

Regression validation for the root mountpoint `/`:

```text
old v2.4.1:
  mount / unmount / exit via the HTTP API all returned 204

fixed build:
  mount / unmount / exit via the HTTP API all returned 204
```

The patched build did not produce `invalid mountpoint basename /`, and root mountpoint behavior remained unchanged.

## Change Type

- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/service/src/fs_service.rs
+++ b/service/src/fs_service.rs
@@ -187,9 +187,11 @@ pub trait FsService: Send + Sync {
         if self.is_fuse() {
             if let Some(rafs) = fs.deref().as_any().downcast_ref::<Rafs>() {
                 let root_ino = rafs.get_root_inode().unwrap();
+                let (mountpoint_parent_ino, mountpoint_name) =
+                    mountpoint_invalidation_target(self.get_vfs(), &cmd.mountpoint)?;
                 self.walk_and_notify_invalidation(
-                    ROOT_ID,
-                    cmd.mountpoint.trim_start_matches('/'),
+                    mountpoint_parent_ino,
+                    mountpoint_name,
                     root_ino,
                     fs_idx,
                 )?;
@@ -248,6 +250,27 @@ pub trait FsService: Send + Sync {
     }
     /// Cast `self` to trait object of [Any] to support object downcast.
     fn as_any(&self) -> &dyn Any;
+}
+
+fn mountpoint_invalidation_target<'a>(vfs: &Vfs, mountpoint: &'a str) -> Result<(u64, &'a str)> {
+    if mountpoint == "/" {
+        // Keep root mountpoint invalidation compatible with the previous behavior.
+        // Path::file_name() does not produce a basename for "/".
+        return Ok((ROOT_ID, mountpoint.trim_start_matches('/')));
+    }
+
+    let name = Path::new(mountpoint)
+        .file_name()
+        .and_then(|name| name.to_str())
+        .ok_or_else(|| {
+            Error::InvalidArguments(format!("invalid mountpoint basename {mountpoint}"))
+        })?;
+    let root_pseudofs = vfs.get_root_pseudofs();
+    let mountpoint_ino = root_pseudofs.path_walk(mountpoint)?;
+    let ino = mountpoint_ino.ok_or(Error::NotFound)?;
+    let parent_ino = root_pseudofs.get_parent_inode(ino).ok_or(Error::NotFound)?;
+
+    Ok((parent_ino, name))
 }
 
 /// Validate prefetch file list from user input.
@@ -440,6 +463,57 @@ mod tests {
             "relative/path".to_string(),
         ]))
         .is_err());
+    }
+
+    #[test]
+    fn it_should_resolve_root_mountpoint_invalidation_target() {
+        let vfs = Vfs::default();
+        let (parent_ino, name) = mountpoint_invalidation_target(&vfs, "/").unwrap();
+
+        assert_eq!(parent_ino, ROOT_ID);
+        assert_eq!(name, "");
+    }
+
+    #[test]
+    fn it_should_resolve_nested_mountpoint_invalidation_target() {
+        let vfs = Vfs::default();
+        let root_pseudofs = vfs.get_root_pseudofs();
+        root_pseudofs.mount("/a/b/c").unwrap();
+
+        let (parent_ino, name) = mountpoint_invalidation_target(&vfs, "/a/b/c").unwrap();
+        let expected_parent_ino = root_pseudofs.path_walk("/a/b").unwrap().unwrap();
+
+        assert_eq!(parent_ino, expected_parent_ino);
+        assert_eq!(name, "c");
+    }
+
+    #[test]
+    fn it_should_fail_to_resolve_missing_mountpoint_invalidation_target() {
+        let vfs = Vfs::default();
+
+        let err = mountpoint_invalidation_target(&vfs, "/a/b/c").unwrap_err();
+        assert_eq!(
+            std::mem::discriminant(&err),
+            std::mem::discriminant(&Error::NotFound)
+        );
+    }
+
+    #[test]
+    fn it_should_fail_to_resolve_relative_mountpoint_invalidation_target() {
+        let vfs = Vfs::default();
+
+        assert!(mountpoint_invalidation_target(&vfs, "a/b/c").is_err());
+    }
+
+    #[test]
+    fn it_should_reject_invalid_mountpoint_invalidation_basename() {
+        let vfs = Vfs::default();
+
+        let err = mountpoint_invalidation_target(&vfs, "/..").unwrap_err();
+        assert_eq!(
+            std::mem::discriminant(&err),
+            std::mem::discriminant(&Error::InvalidArguments(String::new()))
+        );
     }
 
     #[test]
--- a/smoke/tests/api_test.go
+++ b/smoke/tests/api_test.go
@@ -264,6 +264,88 @@ func (a *APIV1TestSuite) TestSubMountCache(t *testing.T) {
 	}
 }
 
+func (a *APIV1TestSuite) TestNestedMountpointStatAfterRemount(t *testing.T) {
+	ctx := tool.DefaultContext(t)
+
+	ctx.PrepareWorkDir(t)
+	defer ctx.Destroy(t)
+
+	config := tool.NydusdConfig{
+		NydusdPath:  ctx.Binary.Nydusd,
+		MountPath:   ctx.Env.MountDir,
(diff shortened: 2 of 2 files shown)


BraveY (approved):
LGTM. Thanks for the PR !

imeoer (approved):
Thanks for this critical fix. Previously, we had only tested it under sub mountpoint located in second-level directories.

By the way, could you please clarify the use case requiring mountpoints with multiple nested levels?

tedcy:
Our case is an internal serverless project. It is a lightweight Docker-in-Docker style environment, but we only use mount namespace isolation and only need a readonly rootfs, so nydus-snapshotter is not really needed in this scenario.

We are migrating from CVMFS because of the maintenance cost. The current model is inherited from CVMFS: images are mounted by their URL, and the URL is split into path components, like:

```
/data/nydus-images/
└── registry.example.com/project/
    ├── team-a/image-a_v1
    ├── team-a/image-a_v2
    └── team-b/image-b_v1
```

This makes image ownership and usage easy to trace.

After you mentioned that it was only tested with two-level mountpoints, I took a quick look at nydus-snapshotter. It seems to only use `/` in dedicated mode and subdirectories like `/snapshotID` in shared mode. Then I realized why others have not hit this issue: most users probably do not use nydusd directly like we do.

imeoer:
@tedcy Thanks, this is an interesting use case, in Ant Group’s envs, there are also docker-in-docker scenarios where nydusd is used to mount the rootfs.

PR #1931, opened 2026-05-06
nydus-image: fix chunkdict deduplicate image logic

shu-pei:
## Overview
This PR fixes two logic issues in the chunkdict `deduplicate_image()` implementation.
The first issue is that `min_test_size` was overwritten before comparison, which made the radius selection condition always true. 
The second issue is that `data_cluster_length` was assigned using `data_cluster.len()`, which represents the total number of data points instead of the number of actually clustered points. This made the early-stop condition ineffective when DBSCAN produced few or no clusters.

## Change Details
This PR makes two small logic fixes in `deduplicate_image()`:
1. Fix radius selection based on `min_test_size`
Before this PR, the code assigned `min_test_size` before comparing it:
```rust
min_test_size = total_test_set_size;
if total_test_set_size <= min_test_size {
    min_test_size = total_test_set_size;
    min_data_dict = data_dict;
}
```
This made the condition always true because min_test_size had already been set to total_test_set_size.
This PR removes the premature assignment so that the dictionary corresponding to the actual minimum test set size can be selected:
```rust
if total_test_set_size <= min_test_size {
    min_test_size = total_test_set_size;
    min_data_dict = data_dict;
}
```
2. Fix clustered point ratio calculation

Before this PR, `data_cluster_length` was calculated as:

```rust
data_cluster_length = data_cluster.len();
```

However, `data_cluster` is still the full `Vec<DataPoint>`, so `.len()` represents the total number of data points, not the number of points that were successfully clustered.

This PR changes the calculation to count only points marked as clustered:

```rust
data_cluster_length = data_cluster
    .iter()
    .filter(|point| point.clustered)
    .count();
```

This better matches the intended early-stop logic:

```rust
if (data_cluster_length as f64 / all_train_length as f64) < 0.2 {
    break;
}
```

## Test Results

I ran the following checks:

```bash
cargo fmt
cargo test test_deduplicate_image
cargo test test_deduplicate_version
```

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/src/bin/nydus-image/deduplicate.rs
+++ b/src/bin/nydus-image/deduplicate.rs
@@ -736,8 +736,7 @@ impl Algorithm<SqliteDatabase> {
             // Adjust the radius size to select the dictionary that tests best.
             while radius <= max_radius {
                 let data_cluster = Self::dbsacn(&mut data_point, radius)?;
-                data_cluster_length = data_cluster.len();
-
+                data_cluster_length = data_cluster.iter().filter(|point| point.clustered).count();
                 let data_dict = Self::aggregate_chunk(data_cluster)?;
 
                 let all_chunks: HashSet<&ChunkdictChunkInfo> =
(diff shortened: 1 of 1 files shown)


Zephyrcf (approved):
Thanks, LGTM.

PR #1945, opened 2026-05-22
Add performance test for Dragonfly SDK Proxy

bergwolf:
Add a daily scheduled perf test for Draognfly SDK proxy. It helps to create a perf-test image that can be reused for perf test in different deployments.

Example run:
https://github.com/bergwolf/nydus/actions/runs/26274407769


BraveY (commented):
It would be nice to have a more intuitive way to view perftest results. Currently you have to download the result.json artifact and dig through the workflow log, which is not very accessible.

PR #1953, opened 2026-06-04
storage: fix backend user io merge for compressed chunks

shu-pei:
## Overview
When testing Nydus startup performance, I found that compression has a very large impact under 4K chunking. An uncompressed image takes about 1 minute to start, while a compressed image takes close to 16 minutes.

The issue seems to be in the previous continuity check, which mixes two different address spaces:
```rust
        // Make sure user io of same region continuous
        if !self.regions.is_empty() && self.joinable(region_type) {
            let region = &self.regions[self.regions.len() - 1];
            if !region.seg.is_empty() && tag.is_user_io() {
                if let BlobIoTag::User(ref seg) = tag {
                    if seg.offset as u64 + start
                        != region.blob_address + region.seg.offset as u64 + region.seg.len as u64
                    {
                        self.commit();
                    }
                }
            }
        }
```

For backend regions:
**start and region.blob_address are compressed offsets.**
**seg.offset and seg.len are uncompressed offsets and lengths.**
As a result, backend reads for compressed chunks cannot be merged correctly and are split into per-chunk requests.

One possible solution is to introduce a separate user-visible logical address, which is uncompressed and used only for the continuity check.

Since this change touches a critical path in Nydus, I would like to discuss whether there is a better approach. I can continue working on the final fix based on the feedback.

## Related Issues

## Change Details

## Test Results

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/storage/src/cache/cachedfile.rs
+++ b/storage/src/cache/cachedfile.rs
@@ -1190,6 +1190,7 @@ impl FileCacheEntry {
                         RegionType::CacheFast,
                         chunk.uncompressed_offset(),
                         chunk.uncompressed_size(),
+                        chunk.uncompressed_offset(),
                         req.tags[i].clone(),
                         None,
                     )?;
@@ -1206,6 +1207,7 @@ impl FileCacheEntry {
                         RegionType::CacheSlow,
                         chunk.uncompressed_offset(),
                         chunk.uncompressed_size(),
+                        chunk.uncompressed_offset(),
                         req.tags[i].clone(),
                         Some(req.chunks[i].clone()),
                     )?;
@@ -1231,7 +1233,14 @@ impl FileCacheEntry {
                 let (start, len) = blob_cci.get_compressed_info(chunk)?;
 
                 // NOTE: Only this request region can read more chunks from backend with user io.
-                state.push(RegionType::Backend, start, len, tag, Some(chunk.clone()))?;
+                state.push(
+                    RegionType::Backend,
+                    start,
+                    len,
+                    chunk.uncompressed_offset(),
+                    tag,
+                    Some(chunk.clone()),
+                )?;
             }
         }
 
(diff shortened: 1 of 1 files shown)


shu-pei:
Since this change touches a critical path in Nydus, I would like to discuss whether there is a better approach. I can continue working on the final fix based on the feedback.

Zephyrcf:
@shu-pei Sorry for being late, I will check it out soon.

storage/src/cache/cachedfile.rs
```diff
@@ -1796,8 +1812,8 @@ impl FileIoMergeState {
             let region = &self.regions[self.regions.len() - 1];
             if !region.seg.is_empty() && tag.is_user_io() {
                 if let BlobIoTag::User(ref seg) = tag {
-                    if seg.offset as u64 + start
-                        != region.blob_address + region.seg.offset as u64 + region.seg.len as u64
+                    if seg.offset as u64 + user_start
+                        != region.user_io_address + region.seg.offset as u64 + region.seg.len as u64
```
Zephyrcf, line 1816: This is indeed a problem. Previous tests implicitly assumed an uncompressed scenario and couldn't trigger the address space mixing issue.                                                        

You could add two tests for the compressed scenario:
1. Compressed offsets are contiguous but uncompressed offsets are not → regions should be split
2. Uncompressed offsets are contiguous but compressed offsets have gaps → regions should be merged

Zephyrcf (commented):
Overall LGTM

shu-pei:
@Zephyrcf I done. Thanks ：）

Zephyrcf (approved):
Good job! LGTM, thanks.

PR #1955, opened 2026-06-12
utils: fix dns timeout cancelled handle

### LINKED ISSUE #1954: [Bug] DNS singleflight gets stuck after a canceled lookup
### Problem Description

In my production setup, the machine was pinned at 100% CPU, so the request timed out even though the DNS service itself was fine. When that registry DNS lookup gets canceled by request timeout, the DNS `singleflight` state can get stuck in the nydusd process.

After that, later requests for the same host keep failing with:

```text
DNS singleflight call was cancelled
```

This keeps happening even after DNS works again. The process does not start a fresh lookup for that host, so the only easy way to recover is to restart nydusd.

### Expected Behavior

If the leader DNS lookup times out or gets canceled, nydus should clear the in-flight `singleflight` entry for that hostname.

After DNS recovers, the next request should be able to start a new lookup and continue normally in the same nydusd process.

### Actual Behavior

The first timed-out lookup leaves an in-flight entry behind.

Later requests hit that stale entry, return `DNS singleflight call was cancelled`, and never become a new lookup leader. In the repro, the DNS drop rule had already been removed, and I could resolve and reach the registry from the host with `getent` and `curl`, but nydusd still did not recover.

The key signal is that after DNS is restored, the second read still returns `DNS singleflight call was cancelled`, and nydusd does not log a new DNS query for the registry host.

### How to reproduce

1. Run nydusd with a registry backend whose host needs DNS resolution.
2. Use a small registry request timeout, for example `timeout=1` and `connect_timeout=1`.
3. Before starting nydusd, use `tc` on the host network device to drop outbound DNS traffic to port 53.
4. Start nydusd and trigger a read from the mounted RAFS filesystem.
5. The first read should time out while DNS lookup is still pending.
6. Remove the `tc` DNS drop rule.
7. Verify host DNS/network is healthy again, for example with `getent` and `curl`.
8. Trigger the same read again in the same nydusd process.

At commit `29ab52f7ee6bf571836741584b402f6fc0df2ea4`, the second read still fails with `DNS singleflight call was cancelled`, and no new DNS query is logged for the registry host.

Any equivalent setup that temporarily drops outbound UDP/TCP DNS traffic works. In my repro, I used `tc` to block port 53 before starting nydusd, then removed the rule before the second read.

### Environment Details

- Nydus-snapshotter version: not involved in the minimal repro
- Baseline commit: `29ab52f7ee6bf571836741584b402f6fc0df2ea4`
- Container runtime: not involved in the minimal repro
- Operating System: Linux
- Kernel version: not tied to a specific kernel
- Repro shape: real nydusd, real registry backend, host DNS failure injected with `tc`

### Additional Information

I also verified the same repro with the proposed fix. After DNS recovered, nydusd started a fresh DNS lookup and the second read succeeded.

imeoer:
Fix https://github.com/dragonflyoss/nydus/issues/1954

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -50,6 +50,7 @@ ignore = [
     { id = "RUSTSEC-2026-0118", reason = "hickory-proto v0.25.2 NSEC3 loop only affects DNSSEC validation (DnssecDnsHandle). Nydus uses TokioResolver for plain IP lookups without DNSSEC. Transitive dep via storage/hickory-resolver 0.25." },
     { id = "RUSTSEC-2026-0119", reason = "hickory-proto v0.25.2 O(n^2) label compression in BinEncoder only triggers with many-record messages. Nydus only sends simple single-record A/AAAA queries; the encoding path cannot be reached by external input. Transitive dep via storage/hickory-resolver 0.25." },
     { id = "RUSTSEC-2026-0104", reason = "rustls-webpki v0.103.12 CRL parsing panic is only reachable when certificate revocation lists are configured. Nydus does not enable CRL verification. Transitive dep from reqwest/hyper-rustls." },
+    { id = "RUSTSEC-2026-0173", reason = "proc-macro-error2 v2.0.1 is unmaintained. Build-time only proc-macro helper, transitive dep via getset (used by neli and oci-spec). No safe upgrade available until upstream migrates away." },
 ]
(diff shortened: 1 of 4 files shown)


Zephyrcf (approved):
LGTM

PR #1957, opened 2026-06-12
fix(nydusify): guard against empty mount slice in Manager.Inspect

immanuwell:
## Overview

`Manager.Inspect` calls `snapshot.Mounts()` and immediately indexes `mount[0]` with no bounds check. If the snapshotter returns an empty slice (no error), it panics with index out of range.

## Related Issues

N/A

## Change Details

Added a `len(mount) == 0` guard right after the nil-error check that returns a clear error with the snapshot key instead of crashing. Also added `TestInspectEmptyMounts` which mocks the snapshotter returning zero mounts to cover this path.

How to hit it: configure `Manager.Inspect` against a container whose nydus snapshot got garbage-collected or is otherwise missing from the snapshotter, `Mounts()` returns `([], nil)`, and you get a panic.

## Test Results

```
=== RUN   TestInspectEmptyMounts
--- PASS: TestInspectEmptyMounts (0.00s)
```

All existing tests still pass.

## Change Type

- [x] Bug Fix

## Self-Checklist

### DIFF
--- a/contrib/nydusify/pkg/committer/manager.go
+++ b/contrib/nydusify/pkg/committer/manager.go
@@ -113,6 +113,9 @@ func (m *Manager) Inspect(ctx context.Context, containerID string) (*InspectResu
 	if err != nil {
 		return nil, errors.Wrapf(err, "get snapshot mount")
 	}
+	if len(mount) == 0 {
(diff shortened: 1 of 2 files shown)


Zephyrcf (approved):
LGTM

PR #1959, opened 2026-06-13
fix: clamp nydusctl backend metric deltas on counter resets

immanuwell:
backend metrics live in process memory, so restart or hot upgrade can reset them.
`nydusctl backend --interval` subtracts counters directly, so the next sample can underflow. kinda nasty.

fix:
use `saturating_sub()` for scalar and per-bucket deltas
add reset regression tests

repro:
1. run `nydusctl backend --interval 1`
2. restart or hot-upgrade `nydusd` or `nydus-snapshotter`
3. wait for the next sample

before:
debug builds can panic on underflow, and wrapped deltas are bogus

after:
reset samples clamp to `0`, then metrics keep going as usual

tests:
`cargo test -q --bin nydusctl`


Zephyrcf (approved):
LGTM

PR #1962, opened 2026-06-15
chore(ci): bump actions for Node.js 24 and fix containerd config compat

Zephyrcf:
-  Upgrade GitHub Actions that rely on Node.js 20 before the forced
-  Downgrade `containerd_config.toml` from `version = 4` to `version = 3` to fix `takeover-test` failures: CI runners ship with containerd v2.2.4 which does not support config version 4. Will revert once runners are upgraded.


imeoer (approved):
LGTM, thanks!

PR #1964, opened 2026-06-15
fix: avoid logger deadlock on broken error channel

### LINKED ISSUE #1958: [Bug] nydusd API hangs after ENOSPC
### Problem Description

In production I hit a disk-full case where `nydusd` blob cache, logs, stdout, and stderr are on the same filesystem. After that filesystem is full, `nydusd` stops answering the daemon API.

The process is still alive. The API socket is still listening. But `/api/v1/daemon` connects and then gets no response. Freeing disk space does not recover it.

### Expected Behavior

`nydusd` should not make the daemon API stop responding forever when the filesystem for cache/logs is full.

### Actual Behavior

Before the filesystem is full, the mount works and `/api/v1/daemon` returns normally.

After that filesystem is filled, reading a larger file from the mount gets stuck. After deleting the filler file, the filesystem has free space again, but `/api/v1/daemon` still times out.

`nydusd` is still running, and the API socket is still in `LISTEN`.

### How to reproduce

Before the filesystem is full, the mount works and `/api/v1/daemon` returns normally.

After that filesystem is filled, reading a larger file from the mount gets stuck. After deleting the filler file, the filesystem has free space again, but `/api/v1/daemon` still times out.

`nydusd` is still running, and the API socket is still in `LISTEN`.

### 1. Build `nydusd` from upstream HEAD

In my test:

```text
nydusd
Version:        v2.4.3-14-g29ab52f7
Git Commit:    29ab52f7ee6bf571836741584b402f6fc0df2ea4
Profile:       release
```

### 2. Prepare a normal registry-backed Nydus image

The image only needs to have one file that is large enough to cause blob cache writes when read. I used an existing image and picked a `.so` file around 24 MB.

### 3. Create a small ext4 filesystem for cache and logs

```bash
truncate -s 512M fullfs.img
sudo losetup --find --show fullfs.img
```

Observed:

```text
/dev/loopX
```

### 4. Format and mount it

Use the device path printed by the previous command:

```bash
sudo mkfs.ext4 -F /dev/loopX
mkdir -p fullfs
sudo mount /dev/loopX fullfs
mkdir -p fullfs/cache
df -h fullfs
```

Observed:

```text
Filesystem      Size  Used Avail Use% Mounted on
/dev/loopX      488M  712K  452M   1% /path/to/fullfs
```

### 5. Start `nydusd`

Put blob cache, log file, stdout, and stderr on `fullfs`.

This is the sanitized config I used. The important part is that blobcache writes to `/path/to/fullfs/cache`.

```bash
cat > nydusd-config.json <<'JSON'
{
  "version": 2,
  "backend": {
    "type": "registry",
    "registry": {
      "scheme": "https",
      "host": "<registry>",
      "repo": "<repo>",
      "timeout": 5,
      "connect_timeout": 5,
      "retry_limit": 0
    }
  },
  "cache": {
    "type": "blobcache",
    "filecache": {
      "work_dir": "/path/to/fullfs/cache"
    }
  }
}
JSON
```

```bash
nydusd \
  --config nydusd-config.json \
  --bootstrap image.boot \
  --mountpoint mnt \
  --apisock api.sock \
  --log-level trace \
  --log-file fullfs/nydusd.log \
  > fullfs/nydusd.stdout 2>&1
```

### 6. Check that stdout/stderr really point to the small filesystem

```bash
pgrep -f 'nydusd .*--apisock api.sock'
```

Observed:

```text
285894
```

Then check fd 1 and fd 2 with that pid:

```bash
readlink /proc/285894/fd/1
readlink /proc/285894/fd/2
```

Observed:

```text
/path/to/fullfs/nydusd.stdout
/path/to/fullfs/nydusd.stdout
```

### 7. Before filling the filesystem, check that the mount and API work

```bash
stat mnt/path/to/a/small/file
sudo curl --unix-socket api.sock http://localhost/api/v1/daemon
```

Observed:

```text
stat: success
curl: returned daemon JSON
curl rc: 0
```

### 8. Fill the filesystem

```bash
dd if=/dev/zero of=fullfs/fill-to-enospc.bin bs=1M conv=fsync
echo $?
df -h fullfs
```

Observed:

```text
dd: error writing 'fullfs/fill-to-enospc.bin': No space left on device
dd rc: 1

Filesystem      Size  Used Avail Use% Mounted on
/dev/loopX      488M  453M     0 100% /path/to/fullfs
```

### 9. Read a larger file from the mount

```bash
timeout 120s dd if=mnt/path/to/a/large/file.so of=/dev/null bs=1M status=progress
echo $?
```

Observed:

```text
read rc: 124
```

### 10. Check the daemon API before freeing space

```bash
sudo timeout 5s curl -sS -m 2 --unix-socket api.sock http://localhost/api/v1/daemon
echo $?
```

Observed:

```text
curl: (28) Operation timed out after 2000 milliseconds with 0 bytes received
curl rc: 28
```

### 11. Remove the filler file and check that space is back

```bash
rm -f fullfs/fill-to-enospc.bin
sync
df -h fullfs
```

Observed:

```text
Filesystem      Size  Used Avail Use% Mounted on
/dev/loopX      488M   59M  394M  13% /path/to/fullfs
```

### 12. Check the daemon API again

```bash
sudo timeout 8s curl -v -m 5 --unix-socket api.sock http://localhost/api/v1/daemon
echo $?
```

Observed:

```text
* Connected to localhost (...) port 80
> GET /api/v1/daemon HTTP/1.1
> Host: localhost
> User-Agent: curl/7.47.0
> Accept: */*
>
* Operation timed out after 5001 milliseconds with 0 bytes received
curl: (28) Operation timed out after 5001 milliseconds with 0 bytes received
curl rc: 28
```

### 13. Check the socket and process

```bash
sudo ss -xanp | grep api.sock
ps -p 285894 -o pid,stat,comm,args
```

Observed:

```text
u_str LISTEN 1 4096 /path/to/api.sock ... users:(("nydusd",pid=...,fd=...))
u_str ESTAB  86 0   /path/to/api.sock ... users:(("nydusd",pid=...,fd=...))

PID   STAT COMMAND
...   Sl   nydusd
```

### Environment Details

- Nydus-snapshotter version: not involved
- Nydus version: `v2.4.3-14-g29ab52f7`
- Nydus commit: `29ab52f7ee6bf571836741584b402f6fc0df2ea4`
- Container runtime: not involved; `nydusd` was started directly
- Operating System: Linux
- Kernel version: `5.4.0-48-generic`
- `flexi_logger`: `0.25.6` from `Cargo.lock`
- Cache backend: `blobcache`
- Remote backend: registry

### Problem analysis

The useful part is the continuous stack of Thread 5, the fuse_server thread that hit the problem. I added short comments inline:

```text
Thread 5 (LWP 654727 "fuse_server"):
#4  std::sys::sync::mutex::futex::Mutex::lock (self=0x7f96f13db820)
#5  std::sync::mutex::Mutex<flexi_logger::writers::file_log_writer::state::State>::lock (self=0x7f96f13db820)
#6  flexi_logger::writers::file_log_writer::state_handle::{impl#2}::write::{closure#0}
    at src/writers/file_log_writer/state_handle.rs:223
    # <-- recursive log path: tries to lock the same flexi_logger State mutex again
...
#16 log_panics::{impl#1}::install_panic_hook::{closure#0}
    msg = "failed printing to stderr: No space left on device (os error 28)"
    thread = "fuse_server"
    # <-- stderr write failed with ENOSPC, then the panic hook logs it
...
#23 std::io::stdio::print_to<std::io::stdio::Stderr>
#24 std::io::stdio::_eprint
#25 flexi_logger::util::try_to_write
#26 flexi_logger::util::eprint_err
    # <-- flexi_logger is trying to print its own write error to stderr
#27 flexi_logger::writers::file_log_writer::state_handle::{impl#2}::write::{closure#0}::{closure#2}
    at src/writers/file_log_writer/state_handle.rs:211
#28 core::result::Result<...>::unwrap_or_else
#29 flexi_logger::writers::file_log_writer::state_handle::{impl#2}::write::{closure#0}
    at src/writers/file_log_writer/state_handle.rs:211
    # <-- outer log write hit ENOSPC while still inside StateHandle::write
...
#33 flexi_logger::writers::file_log_writer::state_handle::StateHandle::write
    at src/writers/file_log_writer/state_handle.rs:198
#35 flexi_logger::primary_writer::multi_writer::{impl#1}::write
#36 flexi_logger::flexi_logger::{impl#1}::log
#39 fuse_backend_rs::api::server::Server<...>::handle_message
#40 nydus_service::fusedev::FuseServer::svc_loop
```

Based on this single-thread stack, my current read is that the bad path is probably:

```text
normal logging path
-> flexi_logger holds its file writer mutex
-> log file write fails with ENOSPC
-> flexi_logger tries to print its own write error to stderr
-> stderr is also on the full filesystem, so that print fails
-> the stderr print failure panics
-> nydusd uses log_panics, so the panic gets logged again
-> the same thread enters flexi_logger again while the same mutex is still held
-> self-deadlock
```

Once this logger mutex is stuck, any path that tries to write a log can also get stuck behind it. That includes the `/api/v1/daemon` request above.

### Local workaround

For my local environment I used a smaller workaround for now:

```rust
.error_channel(flexi_logger::ErrorChannel::DevNull)
```

This drops `flexi_logger`'s own internal error message instead of sending it to stderr. It is not disabling normal Nydus logs.

In my environment, this cuts the stderr ENOSPC -> panic -> recursive logging path. The downside is that logger-internal errors like `[flexi_logger][ERRCODE::Write]` are no longer printed to stderr.

I do not think this is a great long-term fix, so I think the possible full fix below needs Nydus maintainer review.

### Possible full fix

Nydus currently uses `flexi_logger 0.25.6`. That version has `ErrorChannel::DevNull`, but it does not have:

```rust
.panic_if_error_channel_is_broken(false)
```

`flexi_logger` seems to have hit this kind of problem later too. The related discussion is emabee/flexi_logger#155:

https://github.com/emabee/flexi_logger/discussions/155

The maintainer suggested throwing these secondary messages away, and the option was available in `flexi_logger 0.28`:

https://docs.rs/flexi_logger/latest/flexi_logger/struct.Logger.html#method.panic_if_error_channel_is_broken

A cleaner fix may be to upgrade `flexi_logger` and set `.panic_if_error_channel_is_broken(false)`. That should still allow `flexi_logger` to report its own errors when the error channel works, but avoids a panic when the error channel itself is broken.

I cannot tell if upgrading `flexi_logger` is safe for Nydus. It may affect the logger dependency version, error diagnostics, compatibility, and tests. This needs Nydus maintainer review, so I am not sending a PR.

- [ ] Yes I am willing to submit a PR!

tedcy:
## Overview
This bumps `flexi_logger` to `0.28.0` and tells it not to panic when its own error output channel is broken.

That is the path hit in #1958. When cache/log/stdout/stderr are on a full filesystem, a secondary logger error can make `nydusd` stop answering the daemon API.

## Related Issues
Fixes #1958

## Change Details
- Bump `flexi_logger` from `0.25.6` to `0.28.0`.
- Call `.panic_if_error_channel_is_broken(false)` in both nydusd logger setup paths.

## Test Results
- `cargo fmt --check`
- `cargo check --locked -p nydus-rs`
- `cargo test -p nydus-rs logger::tests`
- I also reran the ENOSPC repro from #1958:
  - put blob cache, log file, stdout, and stderr on the same 512 MiB ext4 filesystem
  - filled it to 100%
  - read a real file from the mounted image
  - `/api/v1/daemon` still returned while the filesystem was full
  - `/api/v1/daemon` also returned after space was freed

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1299,15 +1299,14 @@ dependencies = [
 
 [[package]]
 name = "flexi_logger"
-version = "0.25.6"
+version = "0.28.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "84075a94fd76ea9b7a0d2c257444ae30356e58fdac6cd0fe68af33f5279981ae"
+checksum = "f248c29a6d4bc5d065c9e9068d858761a0dcd796759f7801cc14db35db23abd8"
 dependencies = [
  "chrono",
  "flate2",
  "glob",
  "is-terminal",
- "lazy_static",
  "log",
  "nu-ansi-term",
  "regex",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ path = "src/lib.rs"
 [dependencies]
 anyhow = "1"
 clap = { version = "4.0.18", features = ["derive", "cargo"] }
-flexi_logger = { version = "0.25", features = ["compress"] }
+flexi_logger = { version = "0.28.0", features = ["compress"] }
 fuse-backend-rs = { workspace = true }
 hex = "0.4.3"
 hyper = { version = "1.8.1", features = ["client"] }
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -145,6 +145,7 @@ pub fn setup_logging(
             .map_err(|_e| enosys!())?
             .log_to_file(spec)
             .append()
+            .panic_if_error_channel_is_broken(false)
             .format(opt_format);
 
         // Set log rotation
@@ -167,6 +168,7 @@ pub fn setup_logging(
         // can't change log level to a higher level than what is passed to `flexi_logger`.
         Logger::try_with_env_or_str("trace")
             .map_err(|_e| enosys!())?
+            .panic_if_error_channel_is_broken(false)
             .format(colored_opt_format)
             .start()
             .map_err(|e| eother!(e))?;


Zephyrcf (approved):
LGTM, thanks!

PR #1974, opened 2026-06-26
storage: reduce BlobFactory lock scope during blob init

### LINKED ISSUE #1973: [Improvement] Reduce BlobFactory global lock contention during blob init
### Problem Summary

I noticed that `BlobFactory::new_blob_cache() `may suffer from pretty heavy lock contention when there are many blob init requests at the same time.

The lock here:
```rust
let mut guard = self.mgrs.lock().unwrap();
```
can sometimes take 7s or even longer to acquire in our environment.

From the logs, after this lock is acquired, the following logic may still take hundreds of milliseconds to create blob cache. During that time, all other blob init requests are blocked on the same global mgrs lock.

I think we can reduce the lock scope here. Ideally the global lock should only protect the mgrs map itself.  Expensive operations like get_blob_cache() or other blob cache initialization logic should be moved outside the lock if possible.

### Proposed Improvement

- hold mgrs lock only for lookup/insert
- clone/pin the selected manager entry
- release mgrs lock before expensive blob cache operations
- use an active-user guard so GC can skip managers currently in use

### Additional Context

_No response_

Park-Jiyeonn:
## Overview
BlobFactory currently keeps the global mgrs lock while resolving blob caches. When many blob init requests come in, the wait time can become very long.

Keep the lock only around mgr lookup and insert, and use an active-user guard so GC skips managers that are still in use.

## Related Issues
Fixes: #1973

### DIFF
--- a/storage/src/factory.rs
+++ b/storage/src/factory.rs
@@ -12,7 +12,7 @@
 use std::collections::HashMap;
 use std::hash::{Hash, Hasher};
 use std::io::Result as IOResult;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


storage/src/factory.rs
```diff
@@ -174,41 +219,37 @@ impl BlobFactory {
...
         if let Some((config, id)) = victim {
             let key = BlobCacheMgrKey {
                 config: config.clone(),
             };
-            let mgr = self.mgrs.lock().unwrap().get(&key).cloned();
-            if let Some(mgr) = mgr {
-                if mgr.gc(Some(id)) {
-                    mgrs.push((key, mgr.clone()));
+            if let Some(entry) = guard.get(&key).cloned() {
+                if entry.active_users() == 0 && entry.mgr.gc(Some(id)) && entry.active_users() == 0
```
Zephyrcf: The second check `entry.active_users() == 0` is redundant, `pin()` needs to acquire the `self.mgrs` lock, so `active_users` will not increase while the GC holds that lock.

storage/src/factory.rs
```diff
@@ -174,41 +219,37 @@ impl BlobFactory {
...
-                if mgr.gc(None) {
-                    mgrs.push((
-                        BlobCacheMgrKey {
-                            config: key.config.clone(),
-                        },
-                        mgr.clone(),
-                    ));
+            let mut reclaim = Vec::new();
+            for (key, entry) in guard.iter() {
+                if entry.active_users() == 0 && entry.mgr.gc(None) && entry.active_users() == 0 {
```
Zephyrcf: Ditto.

storage/src/factory.rs
```diff
@@ -89,9 +89,51 @@ lazy_static::lazy_static! {
...
+    fn new(mgr: Arc<dyn BlobCacheMgr>) -> Self {
+        Self {
+            mgr,
+            active_users: AtomicUsize::new(0),
+        }
+    }
+
+    fn mgr(&self) -> Arc<dyn BlobCacheMgr> {
+        self.mgr.clone()
+    }
```
Zephyrcf: This can be removed.

storage/src/factory.rs
```diff
@@ -139,7 +181,10 @@ impl BlobFactory {
         };
         let mut guard = self.mgrs.lock().unwrap();
         // Use the existing blob cache manager if there's one with the same configuration.
-        if let Some(mgr) = guard.get(&key) {
+        if let Some(entry) = guard.get(&key).cloned() {
+            let mgr = entry.mgr();
+            let _active_guard = entry.pin();
+            drop(guard);
             return mgr.get_blob_cache(blob_info);
```
Zephyrcf:
```suggestion
            let _active_guard = entry.pin();
            drop(guard);
            return entry.mgr.get_blob_cache(blob_info);
```

storage/src/factory.rs
```diff
@@ -315,8 +356,8 @@ impl BlobFactory {
 
     fn check_cache_stat(&self) {
         let mgrs = self.mgrs.lock().unwrap();
-        for (_key, mgr) in mgrs.iter() {
-            mgr.check_stat();
+        for (_key, entry) in mgrs.iter() {
```
Zephyrcf: Change `mgrs` to `guard`.

Park-Jiyeonn:
Thx for the review! Updated as suggested.

Zephyrcf (approved):
LGTM,thanks!

PR #1977, opened 2026-07-01
fix(upgrade): handle full UDS state transfers

### LINKED ISSUE #1975: [Bug] restore may fail when UDS state data is larger than 32KB
### Problem Description

I noticed a restore issue in the UDS storage backend.

Right now `restore()` reads state data into a fixed 32KB buffer. If the state data is larger than that, only the first 32KB gets read, and the rest is left behind. Then restore tries to use incomplete state data and can fail.

- `recv_with_fd()` only reads up to `MAX_STATE_DATA_LENGTH`
- `MAX_STATE_DATA_LENGTH` is `32 * 1024`
- restore may fail because the data is incomplete

There is a similar risk on the write side too, since `send_with_fd()` may not write the full buffer in one call. 

I think we should keep reading the remaining data after receiving the fd.

Park-Jiyeonn:
## Overview
UDS restore was only taking the first `recv_with_fd()` chunk, so larger state data could be cut off. Keep reading the rest after the fd arrives, and send any bytes left by `send_with_fd()`.

## Related Issues
Fixes #1975

### DIFF
--- a/upgrade/src/backend/unix_domain_socket.rs
+++ b/upgrade/src/backend/unix_domain_socket.rs
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
+    io::{Read, Write},
     os::{fd::RawFd, unix::net::UnixStream},
     path::PathBuf,
 };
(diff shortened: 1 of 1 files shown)


imeoer (approved):
LGTM, Thanks!

PR #1980, opened 2026-07-09
kernel-patches: adapt fscache ondemand IO patch for newer kernel

hwenwur:
find_get_pages_contig() was removed in newer kernels https://github.com/torvalds/linux/commit/48658d8509d2 . Add a follow-up patch converting cachefiles_page_copy() to use filemap_get_folios_contig() instead, so the on-demand IO optimization keeps building on recent kernels.

### DIFF
--- a/contrib/kernel-patches/0002-cachefiles-convert-cachefiles_page_copy-to-use-filem.patch
+++ b/contrib/kernel-patches/0002-cachefiles-convert-cachefiles_page_copy-to-use-filem.patch
@@ -1,0 +1,98 @@
+From 8cded62556a98153a07dacc34198c4e5553e1a12 Mon Sep 17 00:00:00 2001
+From: Wenwu Hou <hwenwur@gmail.com>
+Date: Tue, 7 Jul 2026 17:31:27 +0800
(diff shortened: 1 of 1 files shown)


Zephyrcf:
Thank you for your PR. However, I don't think it's necessary to merge this commit at the moment. As stated in https://github.com/dragonflyoss/nydus/issues/1826, we will be working on supporting fanotify.

imeoer (approved):
If there’s a genuine need, LGTM!

imeoer:
@hwenwur The second commit https://github.com/dragonflyoss/nydus/pull/1980/changes/a2f51abc31b7f59a0a764dfdc6cc94c9dc82ccba appears unnecessary.

hsiangkao:
@hwenwur it would be better to document the kernel versions when applying this patch, otherwise LGTM.

hwenwur:
@imeoer All done

> @hwenwur it would be better to document the kernel versions when applying this patch, otherwise LGTM.

> @hwenwur The second commit [a2f51ab](https://github.com/dragonflyoss/nydus/commit/a2f51abc31b7f59a0a764dfdc6cc94c9dc82ccba) appears unnecessary.

hsiangkao (approved):

PR #1982, opened 2026-07-14
feat: add fanotify pre-content on-demand service for native EROFS mounts

Zephyrcf:
## Overview

Implement the `nydus fanotify` daemon behind `#[cfg(feature = "fanotify")]`.
It serves a multi-device EROFS image on demand through the kernel fanotify
pre-content interface (FAN_CLASS_PRE_CONTENT + FAN_PRE_ACCESS, Linux 6.15+).

The bootstrap is mounted directly as a local EROFS file; each blob's sparse
cache file is a separate `device=` backing file, marked for FAN_PRE_ACCESS.
Cold blob-data fault reads, the daemon fetches and decodes the range into
the cache, and answers FAN_ALLOW — the filled cache file is the same file
the kernel reads, so no separate copy step is needed.

Includes a bounded event coordinator with per-event deadlines, exactly-once
response ownership, a range_ready fast-path that avoids backend I/O when
the authoritative groupmap already covers the faulted range, and a
single-stage graceful shutdown.

Verified end-to-end on kernel 6.15.0-061500-generic with a registry backend.

Also fixes an existing GroupFlight bug where a leader panic left follower
threads permanently blocked (LeaderGuard drop guard).

## Related Issues

## Change Details

## Test Results

[test-fanotify.log](https://github.com/user-attachments/files/30074118/test-fanotify.log)

## Change Type
- [x] Bug Fix
- [x] Feature Addition
- [x] Documentation Update
- [x] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/docs/fanotify.md
+++ b/docs/fanotify.md
@@ -1,0 +1,349 @@
+# Nydus Fanotify Pre-Content Service
+
+## Status
+
+This document describes the current `nydus fanotify` daemon, which serves a
+real kernel EROFS mount on demand through fanotify pre-content hooks.
+Requires Linux 6.15+.
+
+## Overview
+
+The daemon serves a **multi-device** EROFS image. The bootstrap is a real local
+EROFS file that is mounted directly; each data blob is a separate EROFS
+`device=` backed by the accessor's per-blob sparse cache file. The daemon marks
+each blob cache file `FAN_PRE_ACCESS`. Mount, `ls`, and `stat` read the local
+bootstrap and never involve fanotify; only cold blob-**data** reads fault.
+
+```text
+nydus fanotify  --marks each blob cache file-->  FAN_CLASS_PRE_CONTENT group
+
+guest process: cat /mnt/erofs/big.bin
+      |
+      v
+kernel EROFS driver (bootstrap mounted directly; big.bin data is on device N)
+      |
+      | reads blob device N's backing cache file at [offset, +count)
+      v
+backing cache file (sparse hole; pages not present)
+      |
+      | pre-content event {fd -> cache file, RANGE{offset, count}} -- BLOCKS reader
+      v
+nydus fanotify
+      |
+      | fstat(fd) -> (dev, ino) -> blob device
+      | BlobAccessor::fetch(id, offset, count):
+      |   backend fetch + decode + CRC validate
+      |   write decoded data to the blob's cache file
+      v
+write(fan_fd, {fd, FAN_ALLOW}); close(fd)
+      |
+      v
+kernel resumes the blocked read -> EROFS returns bytes
+```
+
+The kernel EROFS driver is unmodified. The filled cache file *is* the file the
+kernel reads, so no separate copy step is needed. A marked file faults on
+**every** read (a pre-content mark also disables readahead on it), so repeat
+reads of a filled range still generate events — but the groupmap fast path
+answers them immediately with no backend I/O.
+
+## Multi-Device Model
+
+The daemon enumerates blob devices from the bootstrap via
+`BlobAccessor::entries()`, which also creates and sizes each blob's sparse cache
+file so the device files exist before mount.
+
+Layout rules:
+
+- The **bootstrap** is a local EROFS file, mounted directly as the mount source.
+  It is never marked and never served through fanotify.
+- Each **blob** is one EROFS `device=` slot, in device-table order. Its backing
+  file is the accessor's decoded cache file (`BlobInfo::cache_path`), sized to
+  the blob's decoded length.
+- Device slots keep their original 1-based index, including **redirect** slots,
+  so the EROFS `device=` index is never renumbered. A read routed to a redirect
+  slot is an invariant violation and is denied.
+- A blob device is identified at fault time by the event fd's `(st_dev, st_ino)`,
+  taken from the same descriptor that was confirmed to be a regular file of the
+  expected size at startup (closes the path-stat/mark/open race).
+
+## Event ABI
+
+The parser mirrors `<linux/fanotify.h>`: a fixed `fanotify_event_metadata`
+header optionally followed by info records. For a pre-content read the kernel
+appends a `FAN_EVENT_INFO_TYPE_RANGE` (type `6`) record carrying the
+`{offset, count}` the reader is about to access.
+
+The RANGE record trails its info header with `{ __u32 pad; __u64 offset;
+__u64 count; }`, so `offset` sits 8 bytes and `count` 16 bytes past the record
+start. `FAN_PRE_ACCESS` is `0x0010_0000`; the metadata version is `3`.
+
+The parser is pure byte handling with no syscalls (unit-tested). It walks each
+event in a `read(2)` batch with checked arithmetic and reports — rather than
+silently dropping — a malformed event. Errors are split by whether the event's
+length was already validated: a **semantic** error (missing/zero/duplicate
+RANGE, malformed info record — the next event's boundary is known) denies just
+that event and keeps parsing the batch, so one odd event cannot take the daemon
+down; a **structural** error (truncated header, bad version, bogus `event_len`
+— the boundary is untrustworthy) stops the batch fail-closed. A queue-overflow
+marker (`fd == FAN_NOFD`) carries no fd or range and is fatal.
+
+## Event Processing
+
+The daemon runs one event loop on a dedicated thread, driven by `epoll` over the
+fanotify fd, a completion-wakeup `eventfd`, and a stop pipe. For each event it
+decides a response purely from the parsed event, then admits the fetch:
+
+1. **Decide.** A non-`FAN_PRE_ACCESS` mask, or a missing/zero RANGE record, is a
+   `FAN_DENY` (without an offset the fetch cannot be bounded).
+2. **Resolve device.** `fstat` the event fd for `(dev, ino)` and look up the
+   blob device. An unknown device, or a redirect slot, is denied.
+3. **Align.** The RANGE is aligned outward to 4 KiB EROFS blocks and clamped to
+   the device size (the accessor requires block-aligned arguments). The device
+   size is the fetch bound; there is no per-event byte cap.
+4. **Fast path.** If the authoritative groupmap already covers the aligned range,
+   the event is allowed immediately with no backend I/O.
+5. **Coalesce or admit.** If an identical `(blob, aligned range)` fetch is
+   already in flight, the event attaches to it and is answered by that one
+   fetch's completion — no duplicate work. Otherwise it takes a slot in an
+   unbounded pending table. Admission never fails; concurrency is bounded by the
+   fetch thread pool (`--fetch-concurrency`), which queues tasks (backpressure)
+   rather than denying.
+6. **Fetch.** `BlobAccessor::fetch(id, offset, count)` runs on a fetch worker
+   thread; it decodes, CRC-validates, dedups (idempotent, group-granular) and
+   writes the blob's cache file in place. The work is wrapped in `catch_unwind`.
+7. **Respond.** On completion the event is answered `FAN_ALLOW` on success, or
+   `FAN_DENY` on a backend/range failure or a worker panic. A fetch in flight
+   when shutdown denies its event still completes and may warm the cache, but
+   as a late completion can no longer change the already-sent response. Bounded
+   fetch time comes from the backend's HTTP timeout + bounded retries, not a
+   per-event deadline; a registry `timeout: 0` (which would disable the HTTP
+   timeout and leave fetches unbounded) is rejected at startup in this mode.
+
+Duplicate work is removed at two layers. The event loop coalesces events sharing
+an identical `(blob, aligned range)` key, so a burst of readers faulting the same
+range dispatches a single fetch. Below that, the accessor de-duplicates at
+blob-meta-group granularity (single-flight), so faults for different ranges in
+the same group also join one fetch. Distinct groups fetch concurrently up to the
+fetch thread pool size (`max(ncpu, 64)`); a saturated pool queues tasks rather
+than denying reads.
+
+### Response protocol
+
+Each event fd is owned exactly once. A decision is selected once, the response
+(`fanotify_response{fd, FAN_ALLOW|FAN_DENY}`) is written to the fanotify fd, and
+only then is the fd closed — the kernel keys the response by fd, and each
+intercepted access dups a fresh fd that must be closed or it leaks. `EINTR` is
+retried; `EAGAIN` yields and retries (the fanotify fd is non-blocking). If a
+permission object is dropped undecided it makes a best-effort `FAN_DENY` before
+closing, so a reader is never left blocked.
+
+### Deny reasons
+
+| Reason | Cause |
+|---|---|
+| `InvalidRange` | missing/zero/out-of-device/overflow range, or non-pre-access mask |
+| `UnknownDevice` | event fd's `(dev, ino)` matched no known blob device |
+| `RedirectRead` | a read targeted a redirect slot the guest must not read |
+| `BackendFailure` | backend fetch, decode, CRC, or cache write failed |
+
+Overload is **not** a deny reason: admission is unbounded (the pending table
+never fills) and a saturated fetch pool queues tasks, so a burst of cold reads
+waits rather than failing with `EPERM`. The per-event denies above fire only on
+malformed or unmappable events, never on load.
+
+A queue-overflow event is handled as fatal (fail-closed): a lost permission
+event cannot be answered safely. The group keeps a **bounded** kernel queue
+(no `FAN_UNLIMITED_QUEUE`) so the overflow safety valve stays intact and
+memory growth is bounded. The kernel default (16384 queued events) is large
+enough for most workloads. Sustained overload (readers arriving faster than
+the backend can drain) eventually overflows and stops the daemon fail-closed;
+a large queue just raises that threshold.
+
+Admission is unbounded — the pending table never fills. Concurrency is
+bounded by `--fetch-concurrency` (the fetch thread pool): a busy pool queues
+tasks, so readers wait rather than seeing `EPERM` (backpressure). To keep the
+unbounded admission from exhausting file descriptors — each in-flight cold read
+pins a dup'd event fd — the daemon raises its `RLIMIT_NOFILE` soft limit to the
+hard limit at startup.
+
+Per-event denies (`InvalidRange`, `UnknownDevice`, `RedirectRead`,
+`BackendFailure`) surface as `EPERM` to the offending `read()`. Most
+applications do not retry `EPERM`, but these only fire on a malformed event or
+a backend failure, not on overload — admission is unbounded and the fetch pool
+backpressures readers rather than denying them.
+
+## Service Lifecycle
+
+1. **Setup.** Raise the `RLIMIT_NOFILE` soft limit to the hard limit (each
+   in-flight cold read pins a dup'd event fd and admission is unbounded). Build
+   the accessor, enumerate and prepare blob devices, create the
+   `FAN_CLASS_PRE_CONTENT` group, and `FAN_MARK_ADD` `FAN_PRE_ACCESS` on each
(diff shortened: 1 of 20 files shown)


joy-allen:
Can we introduce an optimization to skip marking full ready blob files.

Zephyrcf:
> Can we introduce an optimization to skip marking full ready blob files.

This is already implemented in FanotifyService::setup. Fully ready blob files are not marked with fanotify_mark.

scripts/fanotify-e2e.sh
```diff
@@ -0,0 +1,598 @@
...
+        log "  kernel config OK (EROFS file-backed + fanotify perms)"
+    else
+        warn "kernel config not readable; skipping CONFIG_* checks"
+    fi
+
+    [[ -x "$NYDUS_BIN" ]] || die "nydus binary not found/executable: $NYDUS_BIN"
+    "$NYDUS_BIN" fanotify --help >/dev/null 2>&1 || die "nydus binary lacks the fanotify subcommand (build with --features cli,fanotify)"
+    command -v "$NYDUSIFY_BIN" >/dev/null 2>&1 || [[ -x "$NYDUSIFY_BIN" ]] || die "nydusify not found: $NYDUSIFY_BIN"
+    command -v docker >/dev/null 2>&1 || die "docker required for the throwaway local registry"
+    for t in sha256sum dd du mountpoint; do command -v "$t" >/dev/null || die "missing tool: $t"; done
```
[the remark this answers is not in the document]
Zephyrcf: The current e2e script is not integrated with the CI process and is only used for local self-testing. Ignore this comment.

docs/fanotify.md
```diff
@@ -0,0 +1,259 @@
...
+2. **Run.** Register the fd as an `AsyncFd` inside the runtime and signal
+   readiness only once the loop is ready to read events.
+3. **Mount.** After the loop is ready, mount the bootstrap file-backed with each
+   blob cache file as a `device=` option (`mount -t erofs <bootstrap> -o
+   ro,device=<cache>...`, `MS_RDONLY|MS_NODEV|MS_NOSUID`). Because the bootstrap
+   is local, mount and metadata succeed without any fanotify traffic.
+4. **Serve.** Handle events until a termination signal.
+5. **Shutdown.** On `SIGTERM`/`SIGINT`/`SIGHUP`, deny every still-undecided
+   event, then close the group fd; the kernel denies any remaining permission
+   events on its own. The mount is unmounted after the fd is closed, so no read
```
imeoer: **[major] The kernel ALLOWS (not denies) outstanding permission events when the group fd is closed.**

`fanotify_release()` in `fs/notify/fanotify/fanotify_user.c` simulates a userspace reply of **`FAN_ALLOW`** for every unanswered permission event on both the access list and the notification queue — a deliberate fail-open design so that a crashed listener never wedges readers.

Consequence for this daemon: between "group fd closed" and "unmount", any cold read still queued in the kernel (or arriving in that window) is **allowed** and reads zero pages from the sparse cache file — silent data corruption rather than fail-closed. `deny_undecided()` only covers events already read into userspace; it cannot cover events still sitting in the kernel queue. The same applies to a daemon crash.

The same claim appears in `nydus/src/fanotify/service.rs` (L166-169, L511-513) and `nydus/src/bin/nydus/fanotify.rs` (L154-158).

Suggestions:
1. Fix the comment/doc claims in all four places.
2. Reorder shutdown to **unmount first, then close the group fd**, so no new reads can reach the marked device files once the fd stops being serviced.
3. Document the crash-window semantics explicitly (fanotify is inherently fail-open on listener death).

nydus/src/fanotify/service.rs
```diff
@@ -0,0 +1,746 @@
...
+}
+
+#[allow(clippy::too_many_arguments)]
+async fn admit_batch(
+    fan: &AsyncFd<OwnedFd>,
+    core: &Arc<FanotifyCore>,
+    options: &FanotifyOptions,
+    semaphore: &Arc<Semaphore>,
+    writer: &Arc<dyn ResponseWriter>,
+    completion_tx: &mpsc::UnboundedSender<Completion>,
```
imeoer: **[major] A mid-batch parse error silently drops all remaining events in the same `read(2)` buffer → fd leaks + readers blocked forever.**

`EventIter` sets `done = true` on any error, so after this branch denies the one fd and `continue`s, the iterator yields nothing more — but the rest of the batch may contain valid events whose fds were **already dup'd into this process at `read()` time**. Those fds are never closed (leak) and their readers never get a response (they block indefinitely; permission events have no kernel-side timeout).

Two acceptable fixes:
- Treat any mid-batch parse error as fatal (return `Err` and enter the fail-closed shutdown path, same as the no-fd "batch corruption" branch below), or
- For error kinds where the event boundary is still trustworthy (`event_len` already validated, e.g. `InvalidVersion`, `MissingRange`), let the iterator skip the bad event and continue.

The current "deny one, silently drop the rest" is the worst of both options.

nydus/src/fanotify/service.rs
```diff
@@ -0,0 +1,746 @@
...
+            "fanotify: event range {count} exceeds max_event_bytes {}",
+            options.max_event_bytes
+        );
+        return respond(fan, &mut permission, Response::Deny).await;
+    }
+    match core.range_ready(&device.id, offset, count) {
+        Ok(true) => {
+            debug!(
+                "fanotify: range [{}, +{}) already ready for blob {}; allowing immediately",
+                offset, count, device.id
```
imeoer: **[major] Denying on transient overload turns bursts into application-visible read errors (EPERM).**

A burst of more than `queue_depth`/`max_pending_events` concurrent cold reads (common during bulk container startup) makes innocent `read()` calls fail outright.

A more natural backpressure model: when at capacity, **stop draining the fanotify fd** and let events queue in the kernel — readers then wait instead of failing. That interacts with the "queue overflow is fatal" policy, so it may need `FAN_UNLIMITED_QUEUE` or a raised `max_queued_events`.

If fail-fast is intentional, please at least document the trade-off in docs/fanotify.md (currently only the deny reason is listed, not the operational consequence).

nydus/Cargo.toml
```diff
@@ -36,6 +36,12 @@ uffd = [
   "dep:sendfd",
   "dep:tokio",
 ]
+# Serve an EROFS multi-device image on demand through fanotify pre-content
+# hooks (FAN_CLASS_PRE_CONTENT + FAN_PRE_ACCESS). Requires Linux >= 6.15,
```
imeoer: **[minor]** Truncated comment: `Requires Linux >= 6.14,` — the sentence ends with a dangling comma. Also the minimum kernel version is inconsistent across the PR: this says 6.14, `docs/fanotify.md` L7 says 6.15+, L215 says 6.14 for the mask / 6.15 for RANGE, and `event.rs` L14/L20 splits them too. Since the daemon hard-requires the RANGE record (missing RANGE ⇒ deny), the effective minimum is whichever kernel ships `FAN_EVENT_INFO_TYPE_RANGE` — please pin one number everywhere.

nydus/src/fanotify/response.rs
```diff
@@ -0,0 +1,297 @@
...
+impl FdResponseWriter {
+    pub fn new(fan_fd: RawFd) -> Self {
+        Self { fan_fd }
+    }
+}
+
+// SAFETY: this type does not own the descriptor. The service keeps the fanotify
+// AsyncFd alive until all PendingPermission values have been drained or dropped.
+unsafe impl Send for FdResponseWriter {}
+unsafe impl Sync for FdResponseWriter {}
```
imeoer: **[minor]** The `unsafe impl Send/Sync` safety argument ("the service keeps the fanotify AsyncFd alive until all PendingPermission values have been drained or dropped") currently holds only through **implicit local-variable drop order** in `run()` (`pending_events` is declared after `fan`, so it drops first). That's correct but fragile — a refactor that reorders declarations turns the Drop best-effort deny into a write to a closed (or worse, reused) fd number. Suggest either a comment in `run()` pinning the ordering, or having the writer hold an `Arc<OwnedFd>` so the invariant is structural.

nydus/src/fanotify/core.rs
```diff
@@ -0,0 +1,422 @@
...
+    /// Blocking; run under `spawn_blocking`.
+    pub fn fetch(
+        &self,
+        id: &BlobID,
+        cache_size: u64,
+        offset: u64,
+        count: u64,
+    ) -> Result<(), FetchError> {
+        let (aligned_off, aligned_len) =
+            align_fetch_range(offset, count, cache_size).map_err(FetchError::Range)?;
```
imeoer: **[minor]** `admit_event` already calls `align_fetch_range` (service.rs L359) and then this `fetch` re-aligns the same range. Idempotent so harmless, but it obscures who owns the invariant. Consider making `fetch` take pre-aligned arguments and `debug_assert!` alignment instead.

nydus/src/fanotify/core.rs
```diff
@@ -0,0 +1,422 @@
...
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+use anyhow::{Context, Result};
+
+use crate::{BlobID, Config, NydusAccessor};
+
+use super::event::{PreContentEvent, Range};
+
+const EROFS_BLOCK_SIZE: u64 = 4096;
```
imeoer: **[nit]** `EROFS_BLOCK_SIZE` is redefined here; nydus-accessor already has it. Re-export from the accessor to avoid drift.

nydus/src/fanotify/core.rs
```diff
@@ -0,0 +1,422 @@
...
+    /// The blob devices, in device-table order (the order they must be passed as
+    /// EROFS `-o device=` options). Includes redirect slots as placeholders.
+    pub fn devices(&self) -> &[BlobDevice] {
+        &self.devices
+    }
+
+    /// Find the blob device an event fd refers to, by its `(dev, ino)`.
+    pub fn device_for(&self, dev: u64, ino: u64) -> Option<&BlobDevice> {
+        self.devices.iter().find(|d| d.dev == dev && d.ino == ino)
+    }
```
imeoer, line 192: **[nit]** `device_for` is a linear scan per event. Fine for typical blob counts; a `HashMap<(dev, ino), usize>` would make it O(1) if images with many layers are expected.

nydus/src/fanotify/coordinator.rs
```diff
@@ -0,0 +1,184 @@
...
+/// Fixed-capacity table of in-flight permission events.
+///
+/// Job ids are monotonic and never reused, so a completion correlates to its
+/// event by `job_id` alone — no generation counter is needed.
+pub struct PendingTable {
+    entries: HashMap<u64, Entry>,
+    next_job_id: u64,
+    max_pending: usize,
+}
+
```
imeoer: **[design note, non-blocking]** `PendingTable` and the `pending_events: HashMap<u64, PendingEvent>` in service.rs are two parallel maps keyed by the same `job_id`, where this table only adds a `decided` bool. The split buys pure unit-testable logic, which is a fair trade — just noting that merging them (entry holds `Option<PendingPermission>` + decided flag) would remove the "completion has no permission event" impossible-state error paths.

nydus-accessor/src/storage/cache/local.rs
```diff
@@ -120,32 +173,64 @@ impl LocalBlobCache {
...
+                crate::metrics::inc_cache_hit_group();
+                return Ok(());
+            }
+            if let Some(recorder) = self.trace_recorder.as_ref() {
+                recorder.record_group_access(self.blob_index, group_index as u32);
+            } else {
+                crate::metrics::trace::record_group_access(self.blob_index, group_index as u32);
+            }
+
+            let mut buffers = BlobCacheBuffers::default();
```
imeoer, line 216: The `LeaderGuard` fix is correct and necessary — the complete-before-remove ordering avoids the new-leader race, and idempotent `complete()` keeps the normal and panic paths from clobbering each other. Nice.

Two notes worth capturing in the commit message:
1. This is also a **behavior change**: the old `fetch_lock` serialized *all* group fetches per blob; now distinct groups fetch concurrently. Peak memory becomes up to `queue_depth × BlobCacheBuffers` and backend concurrency rises accordingly.
2. `BlobCacheBuffers::default()` is now allocated per cold-group fetch instead of reused via the old mutex-guarded buffer — a reasonable concurrency-for-allocation trade, just making it explicit.

imeoer (changes requested):
Overall this is a high-quality PR: the byte-level event parser is pure and thoroughly tested, `PendingPermission` gives clean exactly-once response ownership, the `O_NOFOLLOW` + fd-identity check closes the path/mark race, and the data path fully reuses nydus-accessor (the daemon itself never copies blob bytes — the kernel reads the same cache file the accessor fills).

Requesting changes for: (1) the incorrect "kernel denies on close" claim and the resulting fail-open shutdown window, (2) the silent drop of packed events after a mid-batch parse error, and (3) at minimum documenting the deny-on-overload behavior. Details in the inline comments.

Verified locally: `cargo clippy -p nydus --features fanotify --all-targets` clean, 38 unit tests pass.

imeoer:
@Zephyrcf Thanks for the PR, could we add a performance comparison between fanotify and fuse in the doc? Using wordpress or openclaw image as examples?

imeoer:
I have suggestions for the 4 options of the nydus fanotify CLI to simplify user decisions and reduce complexity:

The four FanotifyOptions knobs can all be removed or derived, which also simplifies the coordinator considerably:

--queue-depth → fold into --threads. Fetches run on the spawn_blocking pool, so the semaphore duplicates what runtime.max_blocking_threads(N) already provides. Drop the semaphore and let the blocking pool bound fetch concurrency; when the pool is busy, jobs queue instead of being denied — which is exactly the backpressure semantics we want. (Note: fetches are network-bound, so the default should be something like max(ncpu, 64), not the CPU count.)

--max-pending-events → internal constant + pause-draining. Replace "table full ⇒ FAN_DENY" with "table full ⇒ stop reading the fanotify fd". Unread events stay queued in the kernel and readers simply wait — no application-visible EPERM. Truly unlimited isn't safe (each pending event pins a dup'd fd, so RLIMIT_NOFILE is the real bound), but with pause-draining the limit never needs user tuning.

--max-event-bytes → delete the check entirely. align_fetch_range already clamps to the device size, which is the true upper bound; kernel readahead produces small ranges in practice, so this check only rejects mid-size ranges that are perfectly legal. (Clamping-then-allowing is not an option — allowing with a partially-filled range would let the reader see zero pages.)

--event-timeout-secs → rely on backend timeouts. The registry backend config already has connect/read timeouts and bounded retries, so a fetch always returns in bounded time and the event is answered through the normal completion path. This removes the per-event deadline, and with it the entire late-completion machinery (PendingTable's Decide/Late/Unknown states, expire_events, the deadline timer). One caveat: the cache-file write (write_all_at) has no timeout, so a wedged disk would leave readers hanging — but that matches normal POSIX semantics for hung local I/O, and a generous hard-coded safety deadline could be kept internally if desired. Requires backend timeout defaults to be enforced (not "unset = infinite").

Net result: the CLI shrinks to --bootstrap/--config/--mountpoint (+ logging), overload never surfaces as EPERM to applications, and coordinator.rs mostly disappears.

imeoer:
Also, I recommend switching `scripts/fanotify-e2e.sh` to Go’s e2e implementation, such as `tests/integration/fanotify_test.go`.

nydus/src/fanotify/service.rs
```diff
@@ -0,0 +1,1099 @@
...
+    // Loop until submitted or fatal error. The fanotify fd is non-blocking;
+    // EAGAIN just means spin a bit and retry — the kernel drains responses
+    // quickly and this never blocks for long.
+    loop {
+        match permission.try_submit() {
+            Ok(true) => return,
+            Ok(false) => thread::yield_now(),
+            Err(_) => return, // best-effort
+        }
+    }
```
imeoer: **[non-blocking]** `respond()` swallows fatal submit errors (`Err(_) => return, // best-effort`), while `submit()` propagates them and stops the daemon. The asymmetry matters: closing an event fd without a response does **not** unblock the reader (only group release fail-opens), so a fatal write error on the immediate-answer path (ready-range allow, invalid-range deny, etc.) leaves that reader blocked until the daemon exits — and the daemon keeps running, so the shutdown cleanup that would rescue it never triggers.

Suggest propagating fatal errors here too, so both paths converge on the fail-closed shutdown (which now correctly unmounts before dropping the fd). Benign `ENOENT` (event already answered) could be exempted.

nydus/src/fanotify/service.rs
```diff
@@ -0,0 +1,1099 @@
...
+                                thread::yield_now();
+                            }
+                        }
+                        Err(err) => {
+                            if let Some(fd) = err.event_fd() {
+                                if let Ok(owned) = owned_event_fd(fd) {
+                                    let mut permission =
+                                        PendingPermission::new(owned, writer.clone());
+                                    permission.decide(Response::Deny)?;
+                                    while !permission.try_submit()? {}
```
imeoer: **[nit]** This EAGAIN retry loop has no `yield_now()`, so it's a pure busy-spin (unlike the loops in `submit()` and the sibling branch above at L841). Rare in practice, but worth adding the yield for consistency.

nydus/src/fanotify/service.rs
```diff
@@ -0,0 +1,1099 @@
...
+                    let work: Box<dyn FnOnce() + Send> = {
+                        let guard = rx.lock().unwrap();
+                        match guard.recv() {
+                            Ok(w) => w,
+                            Err(_) => return,
+                        }
+                    };
+                    work();
+                })
+                .expect("spawn fetch worker");
```
imeoer: **[nit]** `FetchPool::new` panics via `expect` if thread spawning fails (resource exhaustion at startup). Returning `Result` and letting the caller exit with a normal error message would be friendlier than a panic backtrace.

nydus/src/bin/nydus/fanotify.rs
```diff
@@ -0,0 +1,265 @@
...
+use anyhow::{Context, Result};
+use clap::Args;
+use nydus::config::Config;
+use nydus::fanotify::{mount_erofs, unmount_erofs, FanotifyCore, FanotifyService};
+use nydus::tracing::init_tracing;
+use signal_hook::consts::{signal::SIGHUP, TERM_SIGNALS};
+use signal_hook::iterator::Signals;
+use tracing::{debug, error, info, warn, Level};
+
+const FANOTIFY_PIPE_BUF_SIZE: usize = 8;
```
imeoer: **[nit]** `FANOTIFY_PIPE_BUF_SIZE` is a misleading name — `PIPE_BUF` has a specific kernel meaning (atomic pipe write limit, 4096), while this is just the stop-pipe wake payload. A single byte would do; something like `STOP_WAKE_BYTES` would read more honestly.

imeoer (approved):
Thanks for the thorough rework — this revision addresses all three majors and essentially every minor from my previous review:

- **Fail-open on close**: `run()` now hands the group fd back on both clean and fatal paths, shutdown unmounts **before** dropping the fd with bounded EBUSY retries and deny-draining in between, and the docs/comments now state the kernel's FAN_ALLOW-on-release behavior correctly.
- **Mid-batch parse errors**: the new `is_recoverable()` classification denies-and-skips boundary-safe errors and treats structural corruption as fatal — exactly the right split.
- **Deny-on-overload**: admission is now unbounded with `RLIMIT_NOFILE` raised at startup, and the read-coalescing (`FetchKey`) is a nice bonus for container-startup storms.
- CLI collapsed to a single `--fetch-concurrency` knob, timeouts delegated to the backend with `timeout: 0` rejected at startup, `FdResponseWriter` now holds `Arc<OwnedFd>` (no more `unsafe impl`), O(1) device lookup, shared `EROFS_BLOCK_SIZE`, `debug_assert!`ed pre-aligned fetch. Dropping tokio for raw epoll + a std thread pool also slims the dependency tree considerably.

Approving. A few small non-blocking comments inline. Verified locally: clippy clean, 52 unit tests pass (the Go integration tests need a 6.15 kernel + root, so I did not run them).

PR #1984, opened 2026-07-20
builder: avoid coalescing hard-linked symlinks

### LINKED ISSUE #1983: [Bug] RAFS v5 bootstrap validation fails for hard-linked symlinks: invalid parent inode
### Problem Description

When building a RAFS v5 image from a directory containing multiple directory entries that are hard links to the same symbolic-link inode, `nydus-image create` succeeds but produces a bootstrap that cannot pass validation or be merged.

Image

### Expected Behavior

nydus-image create should generate a valid RAFS v5 bootstrap. The resulting bootstrap should pass both `nydus-image check` and `nydus-image merge`.

### Actual Behavior

create exits successfully, but both check and merge fail with:
```
Error:
    "invalid parent inode"
    at rafs/src/metadata/direct_v5.rs:446
Error: Invalid argument (os error 22)
```

### How to reproduce

```
# 1. Build nydus-image from the affected upstream commit.
rm -rf /tmp/nydus-source /tmp/nydus-hardlinked-symlink-repro
git clone https://github.com/dragonflyoss/nydus.git /tmp/nydus-source
cd /tmp/nydus-source
git checkout 18f9c2f8d7c4d2905c992202a62b63868c492ae6
cargo build --release --bin nydus-image

NYDUS_IMAGE=/tmp/nydus-source/target/release/nydus-image
WORKDIR=/tmp/nydus-hardlinked-symlink-repro

# 2. Create two directory entries that are hard links to one symlink inode.
mkdir -p "$WORKDIR/src/b"
ln -s target "$WORKDIR/src/a"
ln -P "$WORKDIR/src/a" "$WORKDIR/src/b/a"

stat -c 'inode=%i nlink=%h type=%F path=%n target=%N' \
  "$WORKDIR/src/a" "$WORKDIR/src/b/a"

# 3. Create a RAFS v5 bootstrap.
"$NYDUS_IMAGE" create \
  --blob "$WORKDIR/blob" \
  --bootstrap "$WORKDIR/bootstrap" \
  --source-type directory \
  --whiteout-spec none \
  --fs-version 5 \
  --compressor lz4_block \
  "$WORKDIR/src"
CREATE_RC=$?

# 4. Validate and merge the generated bootstrap.
RUST_BACKTRACE=1 "$NYDUS_IMAGE" check -v "$WORKDIR/bootstrap"
CHECK_RC=$?

RUST_BACKTRACE=1 "$NYDUS_IMAGE" merge \
  --bootstrap "$WORKDIR/merged-bootstrap" \
  "$WORKDIR/bootstrap"
MERGE_RC=$?

printf 'create=%s check=%s merge=%s\n' \
  "$CREATE_RC" "$CHECK_RC" "$MERGE_RC"
```

### Environment Details

Nydus source: dragonflyoss/nydus master
Git commit: 18f9c2f8d7c4d2905c992202a62b63868c492ae6
Build profile: release
Rust: rustc 1.94.0
Operating system: Debian GNU/Linux 10 (buster)
Kernel: Linux 5.4.250.bsk.9-amd64 x86_64

### Additional Information

Impact
Affected input: directory sources containing hard-linked symbolic links.
nydus-image create may silently produce an invalid RAFS v5 bootstrap.
Downstream validation and merge fail.
Regular-file hardlinks and independent symbolic links are not affected.

A proposed fix is to coalesce hardlinks only for regular files, assign distinct RAFS inodes to non-regular entries, and track the actual generated inode count separately.

zaliasc:
Directory builds used the source inode and device to coalesce all file types. Hard-linked symlinks under different directories could therefore reuse an inode lower than the parent inode, causing RAFS v5 bootstrap validation and merge to fail.

Limit hardlink coalescing to regular files and track the number of generated inodes separately. Add regression coverage for hard-linked symlinks while verifying that regular-file hardlinks remain coalesced.

## Overview
Fix RAFS v5 bootstrap validation failures caused by coalescing hard-linked symbolic links into the same RAFS inode.

When two symlink directory entries share the same source inode but are located under different parent directories, the builder may reuse a child inode number that is lower than its parent inode. Since RAFS only treats regular files with nlink > 1 as hardlinks, the symlink is validated as a normal child and fails with invalid parent inode.

## Related Issues
Fix #1983 

## Change Details
1. Limit source-inode based hardlink coalescing to regular files.
2. Assign a distinct RAFS inode with nlink=1 to each non-regular directory entry, including hard-linked symlinks.
3. Add inode_count to track the actual number of generated filesystem inodes independently from the regular-file hardlink cache.
4. Use the generated inode count in the RAFS v5 superblock.
5. Add a regression test covering both hard-linked symlinks and regular-file hardlinks.

## Test Results

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/builder/src/core/bootstrap.rs
+++ b/builder/src/core/bootstrap.rs
@@ -38,14 +38,7 @@ impl Bootstrap {
         root_node.index = index;
         root_node.inode.set_ino(index);
         ctx.prefetch.insert(&self.tree.node, root_node.deref());
-        bootstrap_ctx.inode_map.insert(
-            (
-                root_node.layer_idx,
-                root_node.info.src_ino,
-                root_node.info.src_dev,
-            ),
-            vec![self.tree.node.clone()],
-        );
+        bootstrap_ctx.inode_count += 1;
         drop(root_node);
 
         Self::build_rafs(ctx, bootstrap_ctx, &mut self.tree)?;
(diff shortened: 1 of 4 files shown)


imeoer:
Thanks for the detailed analysis and the fix! Before settling on the builder-side change, I'd suggest considering an alternative direction:

**The builder's original coalescing is arguably the faithful behavior** — two directory entries sharing one symlink inode in the source filesystem should map to one RAFS inode with `nlink=2`. The actual bug is on the reader side: the parent-ordering exemption in the v5 validator (`is_hardlink() = is_reg() && nlink > 1`) is narrower than what the builder legitimately produces.

Note that this validation only runs in `nydus-image check` / `nydus-image merge`, and in nydusd only when `validate: true` is set in the RAFS config (default is `false`). In clusters where nydusd has never enabled `validate`, the runtime already serves such images correctly today — lookup/readdir/readlink all work, and `st_ino`/`st_nlink` stay faithful to the source filesystem.

So a smaller fix that keeps full fs consistency would be to leave the builder untouched and only relax the three validation sites (`rafs/src/metadata/direct_v5.rs:445`, `:480` and `cached_v5.rs:377`):

```rust
let coalesced = inode.i_nlink > 1 && !inode.is_dir();
if !coalesced && inode.i_parent >= inode.i_ino {
    return Err(einval!("invalid parent inode"));
}
```

(Directories must stay non-exempt — `parent < ino` remains the loop/corruption guard for them. Regular-file hardlinks are already exempt, so this adds no new attack surface.)

Benefits over the builder-side change:

- Full source-fs fidelity is preserved (shared `st_ino`, `nlink=2`) instead of silently materializing independent symlinks — a lossy conversion.
- Already-built images containing hard-linked symlinks retroactively become valid for `check`/`merge`; a builder-side fix cannot heal those.
- Builder output stays byte-identical for all existing inputs, so there is zero risk to the common path.

The only caveat: older binaries running `check`/`merge` (or nydusd with `validate: true`) would still reject such images — but that combination never worked anyway, so nothing regresses.

imeoer:
After offline discussion, upgrading and fixing this issue at runtime (nydusd) would incur high upgrade costs for existing cluster nodes. Thus, we’ll prioritize a build-side workaround, then improve it in nydus v3 to ensure consistency with the original filesystem.

imeoer (approved):
LGTM!

PR #1995, opened 2026-07-27
perf(fanotify): drop fanotify marks from fully-ready blobs

Zephyrcf:
## Overview
Once a blob's groupmap latches its sticky ALL_READY flag, the fanotify mark on its cache file has no further purpose: every subsequent read is already cache-resident. Remove the mark so the kernel stops generating FAN_PRE_ACCESS events for that file entirely, and warmed reads go straight to the page cache with zero daemon involvement.

## Related Issues

## Change Details

## Test Results

## Change Type
- [ ] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/nydus/src/fanotify/core.rs
+++ b/nydus/src/fanotify/core.rs
@@ -20,18 +20,33 @@
 //! (they need a real image / kernel).
 
 use std::collections::HashMap;
+use std::ffi::CString;
 use std::os::fd::{AsRawFd, FromRawFd, OwnedFd, RawFd};
+use std::os::unix::ffi::OsStrExt;
 use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use anyhow::{Context, Result};
+use tracing::{debug, warn};
 
 use crate::{BlobID, Config, NydusAccessor};
 
-use super::event::{PreContentEvent, Range};
+use super::event::{PreContentEvent, Range, FAN_PRE_ACCESS};
 
 /// EROFS block size as u64 — reuses the canonical constant from the accessor.
 const BLOCK_SIZE: u64 = crate::metadata::EROFS_BLOCK_SIZE as u64;
+
+/// `fanotify_init(2)` flag: close-on-exec on the group descriptor.
+pub(crate) const FAN_CLOEXEC: u32 = 0x0000_0001;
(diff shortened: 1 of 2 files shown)


imeoer (commented):
Overall: the idea is right and the implementation is clean — the "Acquire fast-path load → O(1) `fully_ready` probe → CAS claim" sequence in `try_unmark` guarantees the probe and the `FAN_MARK_REMOVE` syscall run at most once per blob, failure degrades gracefully (the mark just stays and events keep flowing through the `range_ready → Allow` fast path), and the add/remove `(mask, path, AT_FDCWD)` pairs are symmetric. Clippy is clean. A few comments before merging:

1. **[medium] Shutdown race: raw-fd use-after-close.** The fetch worker closure captures a bare `RawFd` integer, but `drain_completions` uses `try_recv` and does **not** wait for in-flight fetches, while the daemon drops the `Arc<OwnedFd>` right after unmount. A slow fetch worker can therefore call `fanotify_mark(fd, FAN_MARK_REMOVE, ...)` after the group fd is closed — and if the fd number has been recycled, the syscall hits an unrelated fd. Suggest capturing a `Weak<OwnedFd>` in the closure (skip the unmark when `upgrade()` fails). Capturing an `Arc` clone would be wrong here since it would delay the fail-open drop semantics on shutdown.

2. **[minor] Cross-process warm case is not covered.** When another process sharing the cache directory fetches the last group, this daemon never unmarks the blob, because `try_unmark` only runs after a fetch performed by this daemon. Consider also calling `try_unmark` on the `range_ready → Allow` fast path in `admit_event` — the probe is O(1), so the cost is negligible.

3. **[minor] Doc nit.** `try_unmark`'s doc says "Returns `true` if this call actually removed the mark", but it also returns `true` when the syscall fails with a non-`ENOENT` error. "Returns true if this call claimed the unmark" would be accurate.

4. **[minor] No unit tests.** The CAS-claim / `mark_unmarked` state machine is testable. Splitting the claim logic from the syscall (e.g. `fn claim_unmark(&self, id) -> Option<usize>` with the syscall in a thin wrapper) would allow testing the concurrent-claim semantics without a kernel dependency.

imeoer:
Thanks, LGTM!

PR #1998, opened 2026-07-29
ci: add tag-triggered static release workflow

imeoer:
## What this does

Adds `.github/workflows/release.yml`, which publishes static release
binaries whenever a `v*` tag is pushed. There is currently no release
automation on the `v3` branch.

- **Targets**: `linux/amd64` (`ubuntu-24.04`) and `linux/arm64`
  (`ubuntu-24.04-arm`), built natively so no cross toolchain is needed.
- **Rust**: `cargo build -p nydus --release --target <musl>` with
  `cli,uffd,fanotify,backend-dragonfly-proxy`, fully static against musl.
- **Go**: `nydusify` built with `CGO_ENABLED=0`.
- **Artifacts**: `nydus-static-<tag>-linux-<arch>.tgz` (containing
  `nydus`, `nydusify` and the sample `config.yaml`) plus a matching
  `.sha256sum`, following the naming used on `master`.
- **Publishing**: the release job is guarded by
  `if: startsWith(github.ref, 'refs/tags/')`; tags containing
  `-rc`/`-alpha`/`-beta` are marked as pre-releases. `workflow_dispatch`
  is kept as a dry run that only uploads workflow artifacts.

## Notes

- Unlike `master`, this does not use `cross`: the `cross` v0.2.5 musl
  images ship no OpenSSL, and `backend-dragonfly-proxy` pulls in
  `openssl-sys` through `dragonfly-client-util`. That crate already
  declares `openssl` with the `vendored` feature, so OpenSSL is built
  from source for the musl target — `perl`/`make` are installed for that.
- Go is pinned to `1.25.5` (same as `ci.yml`): `docker/cli` guards
  `cli/config/configfile` with `//go:build go1.25`, so building
  `nydusify` with the `go 1.24.2` directive from `nydusify/go.mod` fails
  with `undefined: configfile.ConfigFile`.

## Testing

Validated end to end on a fork by pushing a `v3.0.0-alpha.0` tag: both
arch jobs built, the static checks and `--help` smoke runs passed, and
the four assets were published to the release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,158 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+*
+  # Manual dry run: builds and uploads workflow artifacts, but never
+  # publishes a GitHub release.
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+  NYDUS_FEATURES: cli,uffd,fanotify,backend-dragonfly-proxy
+
+permissions:
+  contents: read
+
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  build:
+    name: Build ${{ matrix.arch }} static binaries
+    runs-on: ${{ matrix.runner }}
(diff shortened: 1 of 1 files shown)


BraveY (approved):
LGTM

PR #2002, opened 2026-07-30
fix: support relative registry redirects

pantherman594:
## Overview
Resolve relative redirect URLs when fetching blobs from registries. GCP Artifact Registry returns relative Location headers (e.g. `/artifacts-downloads/...`) on blob redirects, which caused nydusd to fail with `RelativeUrlWithoutBase`.

## Related Issues
N/A

## Change Details
In `_try_read`, the redirect Location header was parsed with `Url::parse()`, which requires an absolute URL. Relative redirect URLs fail. The fix constructs a base URL from the registry's scheme and host fields and uses Url::join() to resolve the `Location` header against it.

`join()` handles both cases: absolute URLs are returned as-is (the base is ignored), and relative URLs are resolved against the base origin.

## Test Results
N/A

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/storage/src/backend/registry.rs
+++ b/storage/src/backend/registry.rs
@@ -913,8 +913,11 @@ impl RegistryReader {
             // Handle redirect request and cache redirect url
             if need_redirect {
                 if let Some(location) = resp.headers().get("location") {
-                    let location = location.to_str().unwrap();
-                    let mut location = Url::parse(location)
+                    let location = location.to_str().map_err(|e| {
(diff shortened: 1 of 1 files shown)


Zephyrcf:
@pantherman594 plz squash the commits.

Zephyrcf (approved):
LGTM, thanks!

PR #2019, opened 2026-08-07
fix: correct ten filesystem and build defects found by a new test suite

imeoer:
xfstests turns out to give a read-only FUSE filesystem no coverage at all. Of its 794 generic cases, 530 call `_scratch_mkfs`, which for FSTYP=fuse means `rm -rf $SCRATCH_MNT/*` and can only fail here. Running all 264 remaining candidates yields 4 passes, and all 4 are vacuous: generic/349, 350 and 351 never reference TEST_DIR, and generic/308 only asserts that no BUG_ON fired. The harness is removed and replaced by a pure-Go suite that builds an image, mounts it, and diffs the mount against the source tree. It runs 569 cases in 20s, so it fits on the pull request path where the fstests run never could.

The suite found the following, each fixed here:

- Directory entries were serialized with "." and ".." pushed first and the children appended, never sorting the combined list. Any name beginning below 0x2E broke the byte order the kernel binary-searches on; fsck.erofs reported filesystem corruption while our own reader tolerated it.

- Filenames were carried as `String` and converted with `to_string_lossy`, so every byte that is not valid UTF-8 became U+FFFD. Two distinct names could collapse into one directory entry, and because the sort ran on the original bytes while the conversion ran afterwards, the entry order was destroyed as well. Names are now raw bytes end to end.

- A name longer than 255 bytes returned ENOENT. The kernel only enforces FUSE_NAME_MAX, so a filesystem advertising f_namelen=255 has to check for itself.

- A symlink target that did not fit behind its inode failed the build outright, at 4065 bytes, where mkfs.erofs accepts up to PATH_MAX. Oversized targets now take a data block of their own and the extended inode layout, whose i_nb really is the block address rather than the link count.

- Unmounting a path the daemon no longer owned tore down whichever filesystem had since been mounted there, because fuser's Mount::drop unmounts by path. Ownership is now established from the session guard and the mountinfo device number.

- Nothing but the daemon's own uid could reach the mount, not even a world-readable file: fuser rejects other uids in user space before the kernel's permission check, and its default ACL was in force. This matches nydus v2, which mounts with allow_other and without nosuid.

- The root directory's permission bits were whatever umask staged the layer, so a umask of 0077 locked every container process out of its own rootfs. They are pinned to 0755 at runtime, as nydus v2 does.

- An mtime beyond the epoch's 32-bit range wrapped silently, turning year 2106 into 1970. Such an inode now moves to the extended layout, the same rule already applied to size, uid, gid and nlink.

- Unpacking a nydus layer back to an OCI tar failed outright on any symlink whose target ran past 100 bytes, the room a ustar header has for one, where a target may run to PATH_MAX. Links are now appended through the tar builder so an oversized target becomes a GNU long link. The round trip that covers this had no Makefile target and no CI job, so it had never run; both are added here.

- Two builds of the same tree could differ. The root's mtime is the build time rather than a property of the content, an empty tree fell through to the wall clock, and listxattr returns attributes in the order they were set. All three are now normalized, and the build reads no clock at all.

## Overview
_Please briefly describe the changes your pull request makes._

## Related Issues

## Change Details

## Test Results

## Change Type
- [ ] Bug Fix
- [x] Feature Addition
- [ ] Documentation Update
- [x] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist


Zephyrcf (approved):
LGTM

PR #2021, opened 2026-08-07
builder: stop requiring device numbers on tarball fifo entries

gliush:
## Overview

Converting a tarball that contains a fifo aborts the build. The tarball builder parses `devmajor`/`devminor` on every special entry, but POSIX only defines those fields for character and block special files, and GNU tar leaves them NUL for a fifo.

## Related Issues

None.

## Change Details

POSIX ustar defines `devmajor` and `devminor` for character and block special files only, so an implementation is free to leave them NUL for a fifo. GNU tar does exactly that in all three of its `gnu`, `ustar` and `posix` formats, and a v7 entry carries no such fields at all. Converting such an archive fails:

```
tarball: failed to get device major from tar entry
numeric field was not a number:  when getting device_major for fifo-entry
```

A fifo has no device number to carry. `stat()` reports `st_rdev` of 0 for one, so the directory builder stores 0, and the tarball builder now stores 0 too instead of parsing fields the entry is not required to write. The same tree built from a directory and from a tarball produces the same inode.

Character and block special files keep parsing both fields, so an entry which converts today converts to the same bytes: the only entries whose result changes are fifos, which used to abort the build.

Also fixes a copy-pasted error context inside the touched block — the `device_minor()` parse reported "device major".

## Test Results

```
cargo test -p nydus-builder --lib            110 passed, 0 failed
cargo clippy -p nydus-builder --all-targets  clean
```

New tests:

- `test_v5_fifo_without_device_numbers` / `test_v6_fifo_without_device_numbers` — a fifo whose device fields are NUL converts, and its `rdev` is 0. Reverting the production change makes both fail with the error above.
- `test_v5_device_files_keep_rdev` / `test_v6_device_files_keep_rdev` — character and block special files still round-trip their device number.

## Change Type

- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/builder/src/tarball.rs
+++ b/builder/src/tarball.rs
@@ -279,20 +279,26 @@ impl<'a> TarballTreeBuilder<'a> {
             RafsVersion::V6 => RafsInodeFlags::default(),
         };
 
-        // Parse special files
-        let mut rdev = if entry_type.is_block_special()
-            || entry_type.is_character_special()
-            || entry_type.is_fifo()
-        {
+        // Parse special files.
+        //
+        // POSIX ustar defines `devmajor` and `devminor` for character and block special files
+        // only, so an implementation is free to leave them NUL for a fifo, and GNU tar does
+        // exactly that in all of its `gnu`, `ustar` and `posix` formats. A v7 entry carries no
+        // such fields at all. A fifo has no device number anyway: `stat()` reports `st_rdev` of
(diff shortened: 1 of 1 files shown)


Zephyrcf (approved):
LGTM

PR #2033, opened 2026-08-14
[v3] tests: add comprehensive userfaultfd integration test suite

bergwolf:
This commit introduces a UFFD CI gate with eight test cases that validate the nydus UFFD service against real userfaultfd instances:

- C1: Runner capability preflight that fails closed when NYDUSFS_UFFD_REQUIRE is set
- C2: Managed-copy fault completion with cold/warm byte equality, EROFS superblock check, tail zero pages, and loop-mounted EROFS whole-tree diff
- C3: Managed zero-page completion for the device tail hole backed by /dev/zero
- C4: Zerocopy fault completion with RANGE_RESPONSE FDs, fixed mappings, and UFFDIO_WAKE
- C5: Handshake prefault plus multi-batch RANGE_RESPONSE coverage on a merged multi-blob image
- C6: Malformed/unsupported input fails closed without hanging
- C7: Backend corruption fails closed with zero wrong bytes
- C8: Disconnect/stop/restart with outstanding faults, bounded cleanup, no stale socket reuse

Every fault has an explicit deadline; no case may hang the suite. When a case fails and NYDUSFS_UFFD_EVIDENCE_DIR is set, structured diagnostic evidence is written for CI to upload.

Add CI workflows for the UFFD test suite and update the Makefile with necessary test infrastructure.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,6 +244,56 @@ jobs:
         run: |
           make test-bench
 
+  uffd-tests:
+    name: Run UFFD tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
+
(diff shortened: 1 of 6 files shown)


gaius-qi (approved):
LGTM

PR #2045, opened 2026-08-28
[v3]fix(builder): preserve non-regular hardlinks across layer merge

bergwolf:
Loading a layer for merge only assigned a hardlink-group key to regular files, so hardlinked fifos, sockets, device nodes and symlinks were split into separate inodes in the merged image while the single-layer build path (FsTreeNode) correctly deduplicates every non-directory by (dev, ino). Key merge nodes on (layer, nid) for all non-directories.

Extend the build/merge equivalence test with hardlinks to a fifo and to a symlink; it fails without this fix.


jiangliu (approved):

gaius-qi (approved):
LGTM

PR #2058, opened 2026-09-01
feat(fuse): add hot upgrade and crash failover

Zephyrcf:
## Description
 
 This PR adds service continuity to the Nydus FUSE frontend through zero-interruption hot upgrade and opt-in crash failover.
 
 The implementation:
 
 - preserves the existing kernel FUSE mount and live `/dev/fuse` connection across daemon generations;
 - adds a bounded, same-host control protocol for instance discovery, peer authentication, and predecessor-to-successor handoff;
 - introduces stable Session Identity and a fixed two-fd Session Transfer containing the live FUSE connection and a process-shared inflight request journal;
 - supports Fresh, Upgrade, and Recover startup modes with explicit ownership and readiness rules;
 - allows only Failover-Protected Sessions created with a Recovery Holder to use hot upgrade or crash recovery, while Standalone Sessions continue to expose `INFO` but reject handoff;
 - parks workers at request boundaries during hot upgrade so existing file and directory handles remain valid and new requests experience only a bounded pause;
 - recovers requests left by a crashed daemon by using `FUSE_NOTIFY_RESEND` when the kernel advertises `FUSE_HAS_RESEND`, or by replying `EIO` to journaled request IDs on older kernels;
 - makes teardown ownership-aware so a retired predecessor never unmounts the successor's live session;
 - keeps inode and file-handle reconstruction deterministic from the same EROFS image, avoiding a serialized userspace VFS snapshot;
 - adds real-mount E2E coverage for successful hot upgrade, pre-READY rollback, and hot upgrade followed by two crash recoveries;
 - documents the continuity model, wire contracts, operational requirements, failure semantics, and current scope limits.

 ## Related Issue
 
 - upstream proposal issue: [cberner/fuser#753](https://github.com/cberner/fuser/issues/753)
 - fuser dependency pr: [Zephyrcf/fuser#1](https://github.com/Zephyrcf/fuser/pull/1)
 
 ## Motivation and Context
 
 A mounted container root filesystem should not have to be detached and recreated merely because its userspace FUSE daemon is upgraded or crashes. Remounting changes the kernel mount identity, invalidates continuity assumptions, disrupts existing file and directory handles, and creates a visible outage for readers.
 
 Hot upgrade now hands the initialized FUSE connection directly from the serving predecessor to a prepared successor. The predecessor pauses at a request boundary, the successor adopts the same kernel session without repeating `FUSE_INIT`, and `READY` establishes the irreversible ownership cutover.
 
 Crash failover extends this model by allowing an external Recovery Holder to retain the live FUSE connection and inflight journal independently of any daemon process. A supervisor can then start one authorized recovery successor, which reconstructs the same deterministic `ErofsFs`, reconciles unfinished requests, and resumes the original mount.
 
 The shared journal is required for kernels without resend support: it preserves the unique IDs of requests read but not completed by the failed daemon, allowing the successor to unblock those callers with explicit `EIO` replies instead of leaving them indefinitely pending.
 
 ## Screenshots (if appropriate)
 
 N/A
 
 ## Types of changes
 
 
 
 - [x] Breaking change (fix or feature that would cause existing functionality to change)
 
 The breaking CLI change is that `nydus fuse` now requires an explicit `--control-socket` (or `NYDUS_FUSE_CONTROL_SOCKET`) so the daemon, its manager, and future generations share one stable coordination endpoint.
 
 ## Testing
 
 - Added `TestFuseHotUpgrade` to verify unchanged mount identity and Session Identity, retained file and directory handles, concurrent new opens/reads/readdir operations, zero I/O errors, and bounded handoff stalls.
 - Added `TestFusePreReadyRollback` to verify that a successor failure before `READY` resumes the predecessor and drains requests queued during the pause.
 - Added `TestFuseFailoverMatrix` to verify hot upgrade followed by two forced daemon crashes and recoveries, journaled inflight request handling, stable mount/rootfs semantics, retained handles, Holder ownership, and explicit shutdown.
 - Added unit coverage for protocol framing, descriptor validation, lifecycle arbitration, startup locking, path validation, transfer identity, inflight recovery, and teardown ownership.
 - Added the `make test-fuse-continuity` target and a dedicated Linux CI job.
 - Ran `cargo fmt --all -- --check`.
 - Ran `go test ./...` in `nydusify/`.
 - Cross-compiled the complete Go E2E package for Linux.
 
 ## Checklist
 
 
 
 
 - [x] My change requires a change to the documentation.
 - [x] I have read the **CONTRIBUTING** document.


nydus-core/src/reader/mod.rs
```diff
@@ -300,6 +301,17 @@ impl ErofsReader {
...
 
+    /// Stable commitment to the opened filesystem image metadata.
+    ///
+    /// A standalone bootstrap is hashed in full. For a full blob the mapped
+    /// region starts at its embedded bootstrap, so this covers the bootstrap,
+    /// blob metadata, and footer without re-reading the potentially large data
+    /// region. The bootstrap device slots commit that data by SHA-256, and
+    /// `open_blob` verifies the actual data region against the slot digest.
+    pub fn image_digest(&self) -> [u8; EROFS_BLOB_ID_SIZE] {
+        sha256_bytes(&self.mmap[self.image_offset..])
```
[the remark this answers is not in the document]
Zephyrcf: This only affects compressed full-blob direct mounts; production continuity uses  open_bootstrap() , which hashes the complete standalone bootstrap without remapping it. Therefore, it is not a blocker for the production hot-upgrade/failover path.

nydus/src/fuse/mount.rs
```diff
@@ -1,113 +1,211 @@
...
+    mount_dev: Option<&str>,
+) -> std::io::Result<()> {
+    // If mountinfo is unreadable, do not unmount what cannot be identified.
+    let owned_mount_present = mount_dev
+        .is_some_and(|dev| mount_dev_of(mountpoint).ok().flatten().as_deref() == Some(dev));
+
+    if !handed_off && owned_mount_present {
+        info!("unmounting {}", mountpoint.display());
+        if origin == SessionOrigin::Adopted {
+            umount2_detach(mountpoint).map_err(|err| {
```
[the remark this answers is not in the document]
Zephyrcf: Done. I added a path-based API in fuser so an adopted session without a Mount object uses the same direct-unmount and fusermount fallback as a fresh session.

docs/upgrade.md
```diff
@@ -0,0 +1,364 @@
...
+---
+
+## 1. Contract
+
+Nydus preserves one kernel FUSE session in two ways:
+
+- **Hot upgrade** is a direct predecessor/successor handoff on the live
+  `/dev/fuse` connection.
+- **Crash failover** is external resource escrow through a Recovery Holder that
+  retains one Failover-Protected Session's Session Transfer.
```
bergwolf, line 19: upgrade and failover workflows are mostly the same and share the same state machine. I would prefer to use the external fd holder (e.g., the snapshotter) to hold the session FDs for nydus FUSE connections. On hot upgrade and failover, new nydus daemon fetches FUSE connection fd from the external fd hold.

docs/upgrade.md
```diff
@@ -0,0 +1,364 @@
...
+the successor inherits the existing Session Transfer. A Standalone Session
+supports direct hot upgrade but not crash recovery. Converting it to a
+Failover-Protected Session requires stopping it, unmounting, and creating a new
+protected mount. Passing `--supervisor-socket` with `--upgrade` does not contact
+the Holder or change the inherited protection.
+
+| Fresh mount configuration | Hot upgrade | Crash failover |
+| --- | --- | --- |
+| `--control-socket` only | Direct predecessor/successor handoff | Not available |
+| Also `--supervisor-socket` | The same direct handoff; protection is inherited | Coordinated by the external Supervisor and Holder |
```
bergwolf, line 40: Then we can just introduce a `--supervisor-socket` and use it for both upgrade and failover.

docs/upgrade.md
Zephyrcf: I'd like hot upgrade to work without a Holder across frontends. The planned hot-upgrade designs for NBD, ublk, and fanotify don’t require an external Holder. FUSE can also upgrade directly by receiving the live connection from its predecessor; it needs external retention for crash recovery.

For protected FUSE sessions, the Holder already keeps its resource copies throughout the upgrade. Fetching from it for both upgrade and failover would unify the source, but would make FUSE upgrades depend on a Holder while the other frontends don’t. Keeping planned upgrades independent of that requirement is why I chose the direct path.

bergwolf: If a user ever cares about live upgrade, he/she would care about crash failover as well. So I want to look at upgrade and failover design together. FWIW, FUSE upgrade does not **require** external FD holder either. But we added it in nydus anyway because failover would need it, and it unified the code path of upgrade and failover (well, mostly).

Now I checked about NBD/ublk/fanotify. Only ublk can support crash failover without FD keeper with kernel >= 6.8). For NBD and fanotify, they would require an external FD keeper as well. So FUSE/NBD/fanotify (and ublk with < 6.7 kernels) crash failover all rely on an external FD keeper.

However, that still doesn't justify the need of external FD keeper for the upgrade case. Given that AI coding is so much easy, I think I can live with two separate path for upgrade and failover.

docs/nydus.md
```diff
@@ -697,6 +704,14 @@ Options:
 		Specify the directory path to mount nydus filesystem [env: NYDUS_FUSE_MOUNTPOINT=]
 	--apiserver <APISERVER>
 		Specify the address to serve Prometheus metrics over a Unix socket, e.g. `unix:///run/nydus/api.sock`. The metrics are exposed at `/metrics` [env: NYDUS_FUSE_APISERVER=]
+	--upgrade
+		Hot-upgrade the instance already serving this mountpoint directly, preserving its failover protection. Does not contact a supervisor; requires pidfd process control [env: NYDUS_FUSE_UPGRADE=]
+	--supervisor-socket <SUPERVISOR_SOCKET>
+		Connect to this Recovery Holder socket to send or receive the FUSE Session Transfer. On a fresh mount, configuring a supervisor enables failover protection [env: NYDUS_FUSE_SUPERVISOR_SOCKET=]
+	--recover
```
bergwolf, line 711: Please do not rely on the control plane to decide if it is upgrade or recovery. In the extreme cases, we may have them together. So nydus should decide it internally. The control plane just launch new daemon with proper supervisor and control sockets. The nydus daemon first check if it can do an upgrade, then if it should do a crash recover, and then fallback to a normal startup.

imeoer:
Can we change `--control-socket` to reuse `--apiserver`, and implement the send / recv of fd and json data through `HTTP 101 Switching Protocols`.

imeoer:
Can we merge `--upgrade / --recover` to `--takeover` ?

imeoer:
I recommend adding a `version` number to the JSON format.

PR #2062, opened 2026-09-02
feat: add redirect flag to blob metadata

Zephyrcf:
- Add a `REDIRECT` incompat flag to `BlobMetadataFlags`.         
 - Add `BlobMetadata::new_redirect()` for building ondemand metadata.                   
 - Use the metadata header as the source of truth for blob type.                  
 - Validate that every block group's `source_blob_index` matches the header flag.              
 - Remove runtime block-group traversal from `BlobMetadata::is_redirect()`.

### DIFF
--- a/docs/nydus.md
+++ b/docs/nydus.md
@@ -1222,9 +1222,11 @@ Header details:
 - `flags` is split EROFS-style: the low 16 bits are incompatible features — a
 	reader that does not know a set bit must reject the file (like
 	`feature_incompat`); the high 16 bits are compatible features — unknown
-	bits are ignored (like `feature_compat`). `COMPRESSOR_ZSTD` (`1 << 0`)
-	means zstd is the blob's default compressor; no compressor bit means
-	stored plain. `DIGESTER_BLAKE3` (`1 << 1`) is mandatory for chunk digests.
+	bits are ignored (like `feature_compat`). `COMPRESSOR_ZSTD` (`1 << 0`) or
+	`COMPRESSOR_LZ4` (`1 << 1`) names the blob's default compressor; no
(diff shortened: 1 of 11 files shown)


gaius-qi (approved):
LGTM

imeoer (changes requested):
Thanks for switching to the header-flag approach, this is the direction I had in mind. A few comments before merge:

**1. `BlobMetadataBlockGroup::new(..., is_redirect: bool)` is redundant.**
The bool must always agree with `source_blob_index != 0`, so callers state the same fact twice and every call site becomes `(..., 0, 0, false)`. I'd keep the original `new` / `new_redirect` split for entries; the blob-level flag in `BlobMetadata::new` is what matters. `from_bytes(bytes, is_redirect)` is fine since it checks against the header flag.

**2. Add a byte-level negative test.**
In `mutated_bytes_follow_the_read_rules`, set bit 3 of `flags` on a sealed regular metadata and assert `from_bytes` fails with "does not match the redirect flag". That's the case the flag is meant to catch (writer bug / hand-crafted file) and it's a one-line table entry.

**3. Remove the now-dead defensive branches in `caches.rs`.**
`prefetch_redirect_blob` still has two `if !block_group.is_redirect()` checks (with a `warn!`). The format layer now guarantees this can't happen, so they can go.

**4. Minor:**
- `block_group_block_count_bits = 0` is also what a regular blob with a single 1-block group gets, so the header doc should say "left zero when there is no uniform span" rather than implying zero means redirect. The `block_group_block_count == 0` check in `validate_block_groups` is dead code (`1 << bits >= 1`).
- The last commit message mentions "wider block group entries", but the on-disk entry is still 40 bytes. The first commit is missing the blank line between subject and body. PR title should be updated (e.g. `feat(format): add REDIRECT flag to blob metadata header`), and I'd squash to 1–2 commits.

PR #2063, opened 2026-09-03
perf: stream FUSE directory entries

Zephyrcf:
## Summary

- Stream mmap‑backed directory entries directly into FUSE `readdir` and
  `readdirplus` replies when the kernel uses the dummy `fh == 0` handle.
- Stop parsing as soon as the FUSE reply buffer is full.
- Avoid materializing the complete directory as `Vec<RawDirEntry>` and copying
  every name into an owned `Vec<u8>` for each paginated request.

## Background

With `FUSE_NO_OPENDIR_SUPPORT`, the kernel skips `opendir` and sends directory
requests with `fh == 0`.

The previous path called `ErofsReader::read_dir()` for every reply page. It
materialized the complete directory and copied every mmap‑backed name, even
though the current FUSE reply needed only one page of entries.

The new path consumes `ErofsReader::for_each_dir_entry()` directly, passes each
borrowed name to `reply.add()`, and stops when the reply is full. The owned
`read_dir()` API remains available for callers that need entries beyond the
callback lifetime.

## Performance

The FUSE daemon was profiled with `perf record` while walking the same synthetic
high‑fanout filesystem before and after the change:

- local backend with prefetch disabled
- cold kernel page/dentry cache
- approximately 181,400 directory entries
- `cycles:P` at 499 Hz with frame‑pointer call graphs
- zero lost samples before and after

| Metric | Before | After | Change |
|---|---:|---:|---:|
| Aggregate sampled cycles | 16.736B | 0.637B | **‑96.19%** |
| Perf first‑to‑last sample span | 2,765.2 ms | 235.9 ms | **‑91.47%** |
| Samples | 2,438 | 327 | ‑86.59% |
| Lost samples | 0 | 0 | — |

The sample span is profiling context rather than an end‑to‑end wall‑clock
benchmark. The aggregate cycles and removed allocator call stacks are the
primary evidence.

Before the change, the profile was dominated by per‑name allocation and
release through `read_dir()`:

```text
ErofsFs::dir_entries
└─ ErofsReader::read_dir
   └─ parse_dir_entries
      └─ name.to_vec() / Vec<RawDirEntry> growth

Arc<Vec<RawDirEntry>>::drop_slow
└─ per‑name free
```
After the change, this materialization stack disappears. The remaining CPU is primarily the required EROFS dirent parsing and  readdirplus  reply construction.  
before：
nydus-fuse-metadata-walk-before

after:
nydus-fuse-metadata-walk-after


imeoer (commented):
Nice improvement, the allocation removal is real and the offset/cookie semantics are preserved. A few comments:

**1. Consider byte-offset cookies to avoid re-parsing from entry 0 on every page (follow-up is fine).**
Each page still walks from the first dirent and skips `offset` entries, so a large directory costs O(N × pages) parsing. EROFS dirents are laid out in 4 KiB blocks and each block's entry count is `first_dirent.nameoff / 12` (`dir_block()` already uses this for lookup). The FUSE `d_off` cookie is opaque to the kernel, so encoding it as a directory byte offset (what kernel EROFS does with `ctx->pos`) would let each page seek straight to its block.

**2. `DirHandle::entries` can probably go now.**
`fh != 0` only happens on kernels without `FUSE_NO_OPENDIR_SUPPORT` (< 5.1). Since streaming from mmap is cheap, both paths could share the streaming code and `DirHandle` would only need `ino`. That removes the `Arc<Vec<RawDirEntry>>` cache, its lazy `Mutex<Option<..>>`, and the "entries materialize on first readdir" comment.

**3. Please bring back the doc comment.**
The old `dir_entries` explained the `fh == 0` dummy handle; `for_each_dir_entry` has none. Also note that the callback returning `false` stops iteration.

**4. Add a pagination test.**
There is no test covering readdir paging, and page boundaries are exactly what this change touches. A unit test that builds a multi-block directory, drives `for_each_dir_entry` with a fake reply accepting K entries per call while advancing `offset`, and asserts the concatenation equals `read_dir()` (no dups, no gaps, empty dir, `.`/`..`) would be a good safety net, especially if (1) lands later.

**5. Benchmark.**
The perf cycle numbers are convincing, but a simple end-to-end wall time (`find . > /dev/null` on the same tree, cold cache, before/after) would be easier to compare against in the future.

Also needs a rebase onto current `v3`.

nydus-core/src/reader/metadata.rs
```diff
@@ -40,18 +41,61 @@ impl ErofsReader {
...
+    ) -> io::Result<()>
+    where
+        F: FnMut(u64, u8, &[u8], u64) -> io::Result<bool>,
+    {
+        let dir_size = inode.size();
+        if offset >= dir_size {
             return Ok(());
         }
+        let block_size = EROFS_BLOCK_SIZE as u64;
+        let mut pos = offset / block_size * block_size;
```
bergwolf: Please use bit shifting to avoid unncessary */ operations.

bergwolf (approved):
lgtm, nice work!

PR #2077, opened 2026-09-07
[v3] fix(build): validate merge sources using blob footers

bergwolf:
The merge path treated a source as a metadata-only bootstrap when its file size matched the primary EROFS image size. This is not a reliable distinction: valid full blobs can have the same size after region alignment and were incorrectly rejected.

Validate the trailing Nydus blob footer instead. This verifies the footer checksum and anchors the declared data, bootstrap, and blob metadata regions to the actual file layout. Keep the separate requirement that each merge source contains exactly one external blob.

Add a regression test covering a valid full blob whose file size collides with the old heuristic.

### DIFF
--- a/nydus/src/build/merge.rs
+++ b/nydus/src/build/merge.rs
@@ -1,5 +1,4 @@
 use std::collections::{BTreeMap, HashMap};
-use std::fs;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


imeoer (approved):
LGTM!

PR #2078, opened 2026-09-07
[v3] layout: remove EROFS_FEATURE_COMPAT_NYDUS_NO_XATTR

bergwolf:
layout: remove EROFS_FEATURE_COMPAT_NYDUS_NO_XATTR and add automatic no-xattr marker tracking

    This commit introduces automatic tracking of the no-xattr marker during image
    building and merging operations. The marker is now properly maintained when
    flattening inode trees, rendering bootstraps, and performing layer merges.

    The implementation ensures that the trusted.nydus.no_xattr xattr is correctly
    set on the root inode based on whether any visible xattrs exist in the
    filesystem tree. This marker is preserved through all build operations
    including bootstrap rendering, layer merging, and optimization.

    EROFS_FEATURE_COMPAT_NYDUS_NO_XATTR is dropped so that we do not
    introduce unnecessary EROFS feature bits that might conflict with EROFS
    upstream some day.

    Key changes:
    - Remove EROFS_FEATURE_COMPAT_NYDUS_NO_XATTR feature bit from superblock
    - Add set_root_no_xattr_marker function to manage the marker xattr
    - Update flatten_tree to track no-xattr state during tree traversal
    - Modify bootstrap rendering to preserve the marker through all operations
    - Add comprehensive tests covering marker behavior in various scenarios
    - Update FUSE filesystem to derive no-xattr state from root marker

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ GO_TEST_ENV = $(SUDO) env "PATH=$(CURDIR)/target/release:$(dir $(GO_BIN)):$(PATH
 	"EROFS_C_FUSE=$(EROFS_C_FUSE)" \
 	"EROFS_MKFS=$(EROFS_MKFS)"
 TEST_SUPPORT_FILES = harness.go optimize.go diff.go
-E2E_TEST_FILES = roundtrip_test.go $(TEST_SUPPORT_FILES)
+E2E_TEST_FILES = roundtrip_test.go no_xattr_test.go $(TEST_SUPPORT_FILES)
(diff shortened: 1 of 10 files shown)


imeoer (commented):
Adding the `--no-xattr` flag feels like it would increase the adaptation burden on snapshotter management and the usage burden of the nydus binary. Could we consider adding `trusted.nydus.no_xattr` on the root inode, or using a `trusted.nydus.hints` bitmap to replace the original `EROFS_FEATURE_COMPAT_NYDUS_NO_XATTR` ?

imeoer (approved):
Thanks, LGTM!

PR #2082, opened 2026-09-09
[v3] layout: target upstream Linux 5.16 EROFS metadata

bergwolf:
The FS version 7 bootstrap is meant to be mounted by the kernel EROFS
driver directly (RunD guests over pmem/rawblock, erofsfuse, fsck.erofs),
not only by our own userspace reader. Auditing the on-disk writer against
fs/erofs/erofs_fs.h showed that several pieces of metadata only had the
meaning we assumed on very recent kernels, or on none at all:

  - Block addresses, device slots and chunk indexes were written with
    48-bit semantics. EROFS_FEATURE_INCOMPAT_48BIT only exists since
    Linux 6.15 (with follow-up fixes after 7.1), while 32-bit addresses
    already cover 16 TiB per device, far beyond any container image.

  - Compact inodes carried their mtime as a 32-bit delta from the
    superblock time in the field at offset 12. Every kernel able to
    mount this format (5.16 through 6.14) defines that field as
    i_reserved and reports sb->build_time for all compact inodes; 6.15
    redefined it as i_mtime. The same image therefore showed real
    timestamps under Nydus FUSE and on 6.15+, but a single collapsed
    timestamp on the kernels before 6.15.

  - The superblock declared build_time as 8 bytes instead of __le32,
    so the tail of the struct (build_time, rootnid_8b, reserved) was
    misaligned relative to the specification.

  - symlink_is_inline() sized the inode header as compact for every
    inode, so an extended inode with xattrs could be laid out inline
    when its payload no longer fit the block.

Since we already require DEVICE_TABLE, upstream Linux 5.16 (4 KiB
blocks, CHUNKED_FILE | DEVICE_TABLE) is the natural metadata baseline.
This series makes the writer produce exactly what that kernel reads,
makes the reader agree with the kernel instead of with a private
encoding, and rejects images that would need newer on-disk features
rather than silently truncating.


imeoer (commented):
Thanks, the direction is right: matching what 5.16 actually reads, and rejecting instead of truncating. I verified the null-chunk encoding and compact-time handling against both 5.16 and current mainline, they agree. Two things I'd like addressed, the rest is minor.

**1. Epoch selection: use the most common mtime, and apply it to merge/optimize too (please do this in the PR).**
Compact inodes now require `mtime == epoch`, so the epoch choice decides how many inodes stay 32 bytes. `build` uses `min()`: one file with mtime 0 (common from build tools) drags the epoch to 0 and turns *every* inode into 64 bytes. `merge` and `optimize` hardcode `epoch = 0`, so the merged bootstrap, the one we actually mount, has all inodes extended. Suggest a shared `choose_epoch(&inodes)` = the most frequent `mtime` among inodes with `mtime_nsec == 0` (ties → smallest, for determinism), used by build, merge and optimize. It's one HashMap pass. Also worth putting a before/after bootstrap size number in the PR description.

**2. Leftover symlink promotion.**
`alloc_inodes` still forces non-inline symlinks to extended. The reason (no room for `startblk_hi` in a compact inode) was removed from inode.rs and no longer applies with 32-bit addresses. Either drop the rule or put the reason back as a comment.

**Minor:**
- The 32-bit commit also fixes compact directories writing `startblk_hi` (0) into the field the kernel reads as `i_nlink`, so native mounts saw nlink=0 for compact dirs. Worth stating in the commit message.
- The first commit's body still describes the "32-bit delta" timestamp check that a later commit removes; consider squashing/rewording.
- CI: `modprobe erofs || true` plus `t.Skip` means the native test can silently skip. A `grep -qw erofs /proc/filesystems` step would make that loud.

PR #2083, opened 2026-09-09
feat(build): a series of optimizations for nydus on block device

imeoer:
## Overview

This PR reworks how nydus v3 builds and serves EROFS images. One builder now
produces either a **chunk-based image** that the daemons fetch on demand
(FUSE, NBD, ublk, fanotify, UFFD) or a **native EROFS layer** that the kernel
mounts directly from a block device or a local store.

> Revision, September 16, 2026: `c7b37842` on `upstream/v3` at
> `657c94a94d0e9b534a58dc4663f06fd8fbc1e3dd`, four commits that each build and
> pass their own tests. The blob meta is a private format at version `1`:
> images, hot blobs, traces and cache sidecars from upstream/v3 or earlier
> revisions of this branch must be rebuilt; no compatibility shims are kept.

### Design

**Fixed-size chunk groups.** Files are cut into chunks of at most
`--chunk-size` (default 1 MiB) and packed densely into chunk groups of that
same size: a chunk that fills a group is a group of its own; smaller chunks
and file tails are bin-packed into shared groups, each on its own block
inside the group's slot. Group `i` owns address slot `[i·S, (i+1)·S)`, so an
address maps to its group by a shift and the group table is one 16-byte row
per group (`compressed_offset`, `first_chunk`, `crc32c`) plus a terminator.
Chunk lengths and blake3 digests are separate dense tables; the blob meta
header is 32 bytes. Each group is compressed on its own (zstd or LZ4, stored
plain when it does not shrink below 70%), verified by crc32c on every fetch
and by digest when `skip_verify_checksums` is false.

**Runtime.** Readiness is one bit per group in a shared `.group.map` sidecar;
byte offset `N` of the same file is the cross-process OFD fetch lock for
group `N`. A miss fetches, in one
backend read, every group whose compressed range overlaps the
`storage.fetch_size`-aligned cell (default 2 MiB of **compressed** bytes)
holding the missed group, trimmed at groups already cached or in flight, and
decodes group by group into the block-aligned plaintext cache. The fetch is
single-flighted in process and its whole range is claimed across processes,
so peers missing on neighbouring groups of one cell wait and then fetch only
what is still missing. Prefetch reads
in batches of the same size through up to `prefetch.concurrent_blob_count`
workers with a single-group ramp at the head; under the default
`prefetch.scope: ondemand` every blob cache is opened by a background pool so
neither the daemon nor a block-device probe pays the blob meta round trips
one blob at a time. The fetch size is a runtime knob, so operators trade GET
count against transferred bytes without rebuilding.

**Optimize.** `/trace` records `{blob_index, chunk_group_index}` for every
chunk group a read touched, hit or miss. `nydus optimize` copies exactly
those groups byte for byte, in access order, into a `REDIRECT` blob together
with their chunk lengths and digests; an 8-byte redirect table
(`source_blob_index`, `source_chunk_group_index`) names the source of every
copy. Nothing is decoded or recompressed, and the bootstrap keeps its chunk
indexes — it only gains the ondemand device and the prefetch xattr. At mount
time prefetch streams the ondemand blob once and writes each decoded group
into the *source* blob's cache at the source slot, skipping groups another
process already cached; the ondemand blob never builds a cache of its own.
This is upstream's redirect design carried over to fixed-size groups.

**Native layers and streaming.** `--compressor erofs-none | erofs-lz4 |
erofs-zstd` build a raw EROFS device (`[layer data][bootstrap][footer]`,
footer flag `RAW_DEVICE`, no blob meta) that the kernel mounts directly; the
daemons pass reads through and the registry backend refuses to serve them.
`nydus build --source-type tar` streams an OCI layer tarball into the blob
without staging a rootfs; nydusify pipes decompressed layers into it.
Compact EROFS metadata, inline directory tails and contiguous child allocation
shrink the merged bootstrap about 4x.

**Daemon memory.** Fetch and decode buffers of 1 MiB or more are anonymous
huge-page-advised mappings unmapped when the job ends; zstd and LZ4 decode
straight into the output slice; the mount services limit glibc to two arenas
with fixed mmap and trim thresholds unless `MALLOC_*` / `GLIBC_TUNABLES` are
set.

### Commits

1. `feat: build and serve dense and native EROFS images` — format, builder,
   runtime, optimize, CLI, config, daemon memory trims.
2. `feat(nydusify): stream layers through the builder, check native images`.
3. `test: validate kernel mounts and conversions end to end`.
4. `docs: describe chunk groups, fetch size and native layers`.

### CLI and config differences from upstream/v3

| | upstream/v3 | this PR |
| --- | --- | --- |
| `nydus build --compressor` | `none`, `zstd`, `lz4-block` | `none`, `zstd`, `lz4`, `erofs-none`, `erofs-lz4`, `erofs-zstd` |
| `nydus build --chunk-size` | 1 MiB | 1 MiB, also the chunk group size |
| `nydus build --block-group-size` | present | removed |
| `nydus build --source-type dir\|tar`, `--digester`, `--blob-id`, `--erofs-data-alignment` | — | new |
| `storage.fetch_size` | — | compressed bytes per on-demand read, default 2 MiB |
| blob meta | `REDIRECT` flag, source blob / block group fields in every block group entry | `REDIRECT` flag kept, separate 8-byte redirect table (`source_blob_index`, `source_chunk_group_index`) after the digest table |
| cache sidecars | `.group.map` (bitmap) | `.group.map` (bitmap, byte `N` also the OFD lock for group `N`) |
| metrics / `/trace` | `cache_*_block_group`, `block_group_index` | `cache_*_chunk_group`, `{blob_index, chunk_group_index}` |
| `nydusify convert --block-group-size` | present | removed |

## Measurements against upstream/v3

Same machine (Linux 7.0 aarch64, 10 cores, 32 GiB), both binaries built with
Rust 1.88.0 `--release --features cli` (plus `ublk`), each with its own
defaults. Workload: OpenClaw 2026.9.3 arm64 (31 layers, 26 data blobs,
3.2 GiB, 123k files) built from identical directory inputs. Runtime: fresh
cache, daemon and runc container per sample; blobs served by an isolated
loopback HTTP Range origin (0 ms; 20 ms per GET; 5 ms per GET plus a
200 Mbit/s body limit). `raw` = no prefetch; `opt` = each binary's own
`optimize` output from its own trace, `prefetch.scope: ondemand`,
`concurrent_blob_count: 4`. FUSE runs with 10 threads on both. End-to-end
time is daemon spawn to the first passing container health check; medians of
three interleaved samples (six for the FUSE/opt/20 ms cell, whose first
round was noisy). Both trees (139,276 entries) are identical.

### Build, size and optimize

| | upstream/v3 | this PR | Δ |
| --- | ---: | ---: | ---: |
| Build wall, all layers | 5.99 s | **5.29 s** | -12% |
| Build user CPU | 10.56 s | **9.62 s** | -9% |
| Build peak RSS (largest layer) | 110 MiB | **63 MiB** | -43% |
| Merge wall / peak RSS | 0.24 s / 131 MiB | **0.14 s / 62 MiB** | |
| Full blobs (data + embedded meta) | 858.8 MiB | 889.2 MiB | +3.5% |
| Blob meta (`.blob.meta`, 26 files) | 5.89 MiB | **4.42 MiB** | -25% |
| Merged bootstrap | 66.8 MiB | **17.0 MiB** | -75% |
| Blobs + bootstrap + meta | 931.5 MiB | **910.6 MiB** | -2.2% |
| `optimize` wall / hot blob | 2.46 s / 118.2 MiB | **0.42 s / 75.3 MiB** | -83% / -36% |

Blobs grow 3.5% because 1 MiB groups compress a little worse than 4 MiB ones
(2 MiB chunks: 869.1 MiB, +1.2%, at 82 MiB build RSS). Everything else
shrinks. The hot blob is a byte-exact copy of the 331 traced 1 MiB groups,
so optimize is a range-read-and-append pass with no decode or recompression;
upstream copies 4 MiB block groups for the same trace.

### Cold start, traffic, cache and memory

`e2e` in seconds; `GET` / `MiB` are origin requests and body bytes; `cache`
is the allocated size of the daemon's cache files; `RSS` is the daemon
process at health check (this PR's RSS is mostly file-backed pages of the
`mmap`ed cache — anonymous memory is the number in parentheses).

| Frontend | Mode | Network | e2e upstream | e2e this PR | GET | MiB | cache MiB | RSS MiB (anon) |
| --- | --- | --- | ---: | ---: | ---: | ---: | ---: | ---: |
| FUSE | raw | 0 ms | 6.04 | 6.37 (+5%) | 156 → 105 | 124 → 170 | 529 → 834 | 182 (105) → **23** (9) |
| FUSE | raw | 20 ms | 9.42 | **8.23** (-13%) | 156 → 105 | 124 → 170 | 529 → 834 | 184 (107) → **21** (7) |
| FUSE | raw | 200 Mbit/s | 15.32 | 15.74 (+3%) | 155 → 105 | 122 → 170 | 525 → 834 | 180 (103) → **21** (7) |
| FUSE | opt | 0 ms | 5.73 | **5.42** (-5%) | 109 → 107 | 134 → 89 | 529 → **343** | 202 (124) → **33** (13) |
| FUSE | opt | 20 ms | 6.05 | **5.67** (-6%) | 108 → 108 | 133 → 92 | 529 → **348** | 207 (130) → **33** (12) |
| FUSE | opt | 200 Mbit/s | 9.30 | **8.35** (-10%) | 116 → 108 | 147 → 97 | 529 → **361** | 232 (154) → **30** (9) |
| ublk | raw | 0 ms | 8.06 | **7.22** (-10%) | 189 → 138 | 125 → 171 | 533 → 837 | 559 (59) → **289** (22) |
| ublk | raw | 20 ms | 11.35 | **9.66** (-15%) | 189 → 138 | 125 → 171 | 533 → 837 | 551 (51) → **291** (25) |
| ublk | raw | 200 Mbit/s | 17.88 | 17.59 (-2%) | 189 → 138 | 125 → 171 | 533 → 837 | 558 (60) → **292** (26) |
| ublk | opt | 0 ms | 5.14 | **4.67** (-9%) | 99 → 99 | 124 → 81 | 529 → **334** | 538 (50) → **299** (27) |
| ublk | opt | 20 ms | 6.21 | **4.60** (-26%) | 99 → 99 | 124 → 81 | 529 → **334** | 542 (54) → **300** (28) |
| ublk | opt | 200 Mbit/s | 7.53 | **6.10** (-19%) | 104 → 104 | 129 → 90 | 529 → **351** | 552 (64) → **297** (24) |

- **Optimized images** (the deployment mode): ublk starts 9–26% faster, FUSE
  5–10% faster, with the same GET count (44 prefetch reads of the 75 MiB hot
  blob in 2 MiB batches plus one meta round trip per blob on both sides);
  32–35% fewer bytes because the 1 MiB groups carry less untraced data than
  4 MiB block groups; cache 529 → 334–361 MiB; daemon anonymous memory
  124–154 → 9–13 MiB (FUSE) and 50–64 → 24–28 MiB (ublk). Under ublk the
  trace covers the boot completely: one on-demand read per start on both
  binaries.
- **Raw images** with the default 2 MiB compressed fetch size: 27–33% fewer
  GETs, so 10–15% faster over 20 ms latency and under ublk locally, within
  noise (+5% / +3%) for FUSE at 0 ms and 200 Mbit/s, at +37% bytes and a
  +58% plaintext cache (the blobs compress about 3.7x, so a fetch of `W`
  compressed bytes decodes to roughly `3.7 W` of cache). `storage.fetch_size`
  moves along this curve at runtime, measured in an earlier round of the same
  harness: 1 MiB → 144 GETs / 137 MiB / 662 MiB cache (upstream parity),
  4 MiB → 80 / 228 / 1123.
- Not measured here: NBD/fanotify/UFFD startups (functional e2e only), native
  `erofs-*` runtime, real EBS volumes, x86_64. Harness, raw JSON and logs are
  kept out of the repository.

## Change Type

- [x] Feature Addition
- [x] Documentation Update
- [x] Code Refactoring
- [x] Performance Improvement
- [ ] Bug Fix
- [ ] Other

## Self-Checklist

- [x] fmt, Clippy (`-D warnings`, all features), Rust workspace tests and Go vet/race tests pass on every commit.
- [x] Unit, real-builder and kernel-mount e2e regression tests added and run.
- [x] Documentation updated for behavior, format and compatibility changes.
- [x] Build, size, startup, traffic, cache and memory measured against the current `upstream/v3` on FUSE and ublk; regressions reported alongside the gains.
- [x] Benchmark data and harness kept out of the commit chain.


gaius-qi (approved):
LGTM

nydus-format/src/blob/metadata.rs
```diff
@@ -18,827 +18,769 @@ use std::path::Path;
...
+        crate::utils::align_up_u64(self.used_size(), BLOCK).expect("blob meta size overflowed")
     }
 }
 
-/// One chunk entry: the digest of a fixed-size span of the blob's dense
-/// uncompressed address space. Chunk entries are packed back to back in the
-/// chunk table right after the header.
-///
-/// The entry's 48 bytes (integers little-endian):
+fn align8(value: u64) -> u64 {
```
gaius-qi: `nydus-format/src/utils/align.rs`

gaius-qi (approved):

nydus-format/src/blob/metadata.rs
imeoer: Done in c7b37842: `align8` is gone, the header offsets and the checked size use `utils::align_up_u64`. Same sweep replaced the zero-padding loop in `BlobWriter::write_z_padding` with `utils::write_zeros`.

PR #2087, opened 2026-09-10
[v3] scripts: remove nydus crate from publish list

bergwolf:
The nydus crate is for nydus binary. There is no need to publish it.


imeoer (approved):
LGTM!

PR #2088, opened 2026-09-10
[v3] config: re-run the build script when the checked-out commit changes

bergwolf:
The build script embedded the git commit into GIT_COMMIT_SHORT_HASH and GIT_COMMIT_DATE without declaring any rerun-if-changed inputs, so cargo only re-ran it when a file inside nydus-config changed. Checking out or committing elsewhere in the tree kept the cached environment, and `nydus --version` reported the commit of an earlier build while the rest of the workspace had been recompiled from the new sources.

Declare HEAD, the refs/heads directory, packed-refs and reftable as build-script inputs, resolving each through `git rev-parse --git-path` so the `gitdir:` indirection of submodules and worktrees is followed and shared entries land in the common directory while HEAD stays per-worktree. The refs/heads directory is declared as a whole rather than the single branch file so a loose ref written after `git gc` packed the branch is still noticed. Only existing paths are declared because cargo treats a missing input as always stale and would re-run on every build. Building outside a source checkout still emits nothing and falls back to "unknown".

### DIFF
--- a/nydus-config/build.rs
+++ b/nydus-config/build.rs
@@ -1,4 +1,5 @@
-use std::path::Path;
+use std::env;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
 /// Represents the git commit information.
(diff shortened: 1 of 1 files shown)


imeoer (approved):
LGTM!

PR #2089, opened 2026-09-11
[v3] tests: strengthen EROFS mount coverage and replace erofsfuse dependency with kernel erofs

bergwolf:
This commit removes the erofsfuse compatibility layer and replaces it with direct kernel EROFS mount validation. The changes eliminate the need for the external erofsfuse tool while maintaining comprehensive mount testing.

The key improvements include:
- Removing setupCErofsFuse() and related mkfs lookup functions that were only used for erofsfuse compatibility
- Refactoring verifyNativeErofsTree() to extract mountNativeErofs() as a reusable helper
- Adding TestMergedMountKernelErofsMatchesNydusFuse() that directly mounts EROFS images using the kernel driver
- Updating cachedBlobDataDevicesForBlobs() documentation to reflect the new kernel mount approach

This change simplifies the test infrastructure by relying on the kernel's native EROFS support rather than an external FUSE implementation, making tests more reliable and reducing external dependencies.

### DIFF
--- a/tests/e2e/harness.go
+++ b/tests/e2e/harness.go
@@ -24,28 +24,7 @@ var blobMetaFilenamePattern = regexp.MustCompile(`^[0-9a-f]{64}\.blob\.meta$`)
 
 const (
 	erofsCFuseEnv = "EROFS_C_FUSE"
-	erofsMkfsEnv  = "EROFS_MKFS"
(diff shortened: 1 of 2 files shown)


imeoer (approved):
LGTM!

PR #2102, opened 2026-09-18
feat: content-defined chunk groups on a dense blob meta layout

imeoer:
## Summary

Use content-defined chunk groups so independently built images can share the units fetched, verified and cached by nydus. Dense blob metadata stores one digest per group and a direct granule index for O(1) address lookup, without a derived runtime index.

Defaults are independent 2 MiB file chunks and a 2 MiB group minimum: chunks at least the minimum stand alone; smaller chunks form 2-8 MiB packs. Only the final group may be smaller. The experimental metadata version stays at 1; older layouts must be rebuilt.

## Performance

Fresh measurements on Openclaw 2026.9.2, arm64 manifest `sha256:992c1a931f2c82063d008ba8539b44ceb181f17dc914abedb352a24ef0638edc`. All columns convert the **same merged 2.79 GiB rootfs tar**, using zstd/BLAKE3 and each builder's default chunk geometry. This replaces the earlier per-layer performance measurements; image download/export is excluded.

Baselines: **RAFS v6** uses the checksum-verified official [v2.4.5 release](https://github.com/dragonflyoss/nydus/releases/tag/v2.4.5), commit `e3190057`, not current master HEAD; **upstream/v3** is `9ba51335`; **this PR** is `465b0de5`. Host: Ubuntu arm64 VM, Linux 7.0.0, 10 CPUs, 32 GiB RAM, ext4. Both v3 binaries use Rust 1.88 release builds; the official v6 binaries were built with Rust 1.94.

Three-run medians, sequential rotated variants. **Runtime E2E = FUSE mount start through overlay/runc startup to the first HTTP 200 from `/healthz`**, the image's healthcheck endpoint. Bootstrap is already local; blobs come from the same loopback HTTP range origin, without injected delay/rate limits. Every run uses a fresh mount and cache, with best-effort file-specific page-cache eviction, not a global cache drop. Raw prefetch is off; optimized runs enable native prefetch with concurrency one. Fetch/amplify target is 2 MiB; runtime digest verification is disabled on both sides.

| Metric | RAFS v6 (v2.4.5 release) | upstream/v3 | This PR |
|---|---:|---:|---:|
| Build wall / CPU (s) | 15.02 / 15.02 | 5.09 / 11.56 | 4.90 / 10.88 |
| Build peak RSS (MiB) | 275.8 | 105.3 | 132.8 |
| Image artifacts: blobs + bootstrap (MiB) | 933.0 | 899.3 | 864.9 |
| Standalone bootstrap (MiB) | 24.39 | 16.40 | 16.39 |
| Sum of cache `.blob.meta` files (MiB) | 1.57 | 4.18 | 0.53 |
| Bootstrap + `.blob.meta` total (MiB) | 25.96 | 20.58 | 16.92 |
| Raw E2E median [min, max] (s) | 6.59 [6.54, 6.77] | 4.69 [4.67, 5.11] | 5.31 [5.06, 7.18] |
| Raw daemon peak RSS / CPU (MiB / s) | 25.4 / 4.67 | 28.6 / 1.30 | 31.7 / 1.48 |
| Raw application peak RSS / CPU (MiB / s) | 465.8 / 3.65 | 466.5 / 3.17 | 495.8 / 3.74 |
| Raw cache allocated / logical (MiB) | 624.8 / 2947.8 | 915.6 / 3090.2 | 976.6 / 3085.2 |
| Raw backend GET MiB / requests | 154.5 / 213 | 175.9 / 83 | 185.1 / 77 |
| Optimize wall / CPU (s) | 1.56 / 1.52 (invalid output) | 0.43 / 0.38 | 0.58 / 0.52 |
| Optimize peak RSS (MiB) | 332.5 (invalid output) | 307.7 | 433.7 |
| Additional optimized blob (MiB) | 57.7 (invalid output) | 72.9 | 115.9 |
| Optimized E2E median [min, max] (s) | N/A: mount rejected | 3.85 [3.56, 3.92] | 4.02 [3.96, 4.18] |
| Optimized daemon peak RSS / CPU (MiB / s) | N/A | 26.6 / 0.78 | 33.6 / 1.01 |
| Optimized application peak RSS / CPU (MiB / s) | N/A | 462.9 / 3.22 | 465.1 / 3.27 |
| Optimized cache allocated / logical (MiB) | N/A | 322.9 / 3405.0 | 497.9 / 3578.5 |
| Optimized backend GET MiB / requests | N/A | 82.0 / 48 | 131.8 / 77 |

- CPU means user + system CPU-seconds, not CPU percentage. Build/optimize RSS uses `wait4`; daemon CPU/peak RSS are captured at readiness; application CPU uses its dedicated cgroup and application RSS is a sampled process sum. Builds do not include an explicit fsync gate. Shared-host noise remains; notably the PR raw E2E range is wide.
- Image artifacts exclude sidecar duplicates. Bootstrap size is the standalone uncompressed file; `.blob.meta` is the sum of materialized cache metadata files at raw readiness. V6 does not emit these sidecars during the normal compressed build. Cache allocation is `st_blocks * 512` at readiness, including metadata/readiness files; logical sparse size is shown separately. Do not add the metadata diagnostic rows again to image size.
- Optimize uses local source blobs and excludes trace collection/upload. V6 uses the 5,338-file read set exported by its daemon; v3 uses native group traces (314 upstream groups, 169 PR groups). These are different optimization granularities. **All three v2.4.5 optimize commands exited successfully but produced invalid RAFS v6 metadata**: official `nydus-image check` and `nydusd` reject it with `RafsV6Blob: idx 1 invalid fields, ci_d_size 1ffe0, chunk_count 1554`. Its command cost is shown only as failure evidence; there is no valid optimized E2E result.

On this sample the PR reduces build time and image size versus RAFS v6, but uses more fetched bytes and allocated cache. Versus upstream/v3 it reduces metadata size substantially, while raw startup, optimized fetch volume and optimize peak RSS regress. These are workload-specific measurements, not general performance guarantees.

## Deduplication

Separate earlier swebench measurement, unchanged by this performance rerun: three independently built sibling images, order 4021, 3947, 4182 (17 layers each, 20 unique layers total). Values are additional compressed bytes required for each subsequent image; they are not measurements of the merged Openclaw input above. RAFS v6 deduplication was not remeasured here.

| Deduplication unit | 2nd image | 3rd image |
|---|---|---|
| OCI layers (baseline) | 200 MiB | 199 MiB |
| Final chunk groups (C = M = 2 MiB; 1,664 groups total) | 15.1 MiB | 29.7 MiB |

## Validation

All three raw FUSE application starts and both v3 optimized paths reached readiness. Build and runtime smoke tests checked actual file contents; native kernel EROFS mounting was not part of this benchmark. Earlier independent full-read verification matched 139,233 entries and SHA256 hashes of 121,162 files against upstream/v3; that full-tree check was not repeated for this measurement.


gaius-qi (approved):
LGTM

PR #2103, opened 2026-09-19
backend: bound the shared tokio runtime and pin its netns

bergwolf:
The registry backend bridges the synchronous BlobBackend trait into async
HTTP through one process-wide tokio runtime built on first use. With
tokio's defaults that runtime spawns one worker per CPU eagerly and keeps
them for the life of the process (192 threads per embedding VMM on a
192-core node), and every runtime thread inherits the network namespace
of whichever thread touched the runtime first, so an embedder that must
reach the registry from the host namespace has to order its first read
carefully and can never verify that it did.

Add `RuntimeOptions` and `configure_runtime()`: worker and blocking
thread bounds plus an optional netns descriptor that every runtime thread
enters as it starts (probed on a scratch thread first, so an unusable
descriptor fails the call rather than a worker). The first call wins,
repeating it with the same options is a no-op, and different options or a
call after the runtime exists report `AlreadyExists` instead of silently
doing nothing. Without a call the runtime now defaults to 2 workers and
at most 8 blocking threads (DNS lookups) released after 10 s idle; the
workers only drive socket I/O for the connection pool, callers block on
their own threads.


nydus-backend/src/registry/rt.rs
```diff
@@ -0,0 +1,352 @@
...
+        ),
+    }
+    runtime
+});
+
+/// Access the shared backend runtime, building it on first use.
+pub(super) fn runtime() -> &'static Runtime {
+    &RUNTIME
+}
+
```
Zephyrcf, line 201: The connection-establishing futures in `request_http()` and `request_http_once()` are still polled on the calling thread via `Runtime::block_on()`.

If that thread is in a different netns, new sockets are created there rather than in the configured namespace.

Could we run connection setup on the configured workers too, including the fallback and Dragonfly paths?
```suggestion
/// Run connection setup on a worker in the configured network namespace.
/// `Runtime::block_on` alone polls its future on the caller's thread instead.
pub(super) fn block_on_worker<T>(future: impl Future<Output = T> + Send + 'static) -> io::Result<T>
where
    T: Send + 'static,
{
    let runtime = runtime();
    runtime
        .block_on(runtime.spawn(future))
        .map_err(|err| io::Error::other(format!("backend runtime task failed: {err}")))
}
```

bergwolf: Fixed. HTTP, fallback, and Dragonfly construction/requests now use block_on_worker, which spawns their futures onto runtime workers.

Zephyrcf (approved):
LGTM

PR #2104, opened 2026-09-20
prefetch: add the auto scope and stream ondemand blobs first

bergwolf:
`prefetch.scope` gains `auto`, now the default: when a priority blob is an ondemand (REDIRECT) blob the prefetch behaves like `ondemand` (only that blob is streamed), otherwise like `all`, so an unoptimized image streams every blob in the background the way a classic nydusd mount does instead of serving every read on demand. Under `all` the ondemand blobs are streamed before the other priority blobs regardless of declared order, so whole-blob pulls skip the groups the redirect already filled.

The auto decision runs on the prefetch thread after the blob opener pool starts, so it never delays the mount and reuses the meta round trips the openers already pay.


Zephyrcf (approved):
LGTM

PR #2107, opened 2026-09-22
refactor(blob-metadata): express group span in bytes and always verify seal on load

gaius-qi:
## Overvie
Blob metadata now records the chunk-group span as a byte shift instead of a chunk count, and the crc32 seal is verified unconditionally when the metadata is loaded rather than only on request. Reserved and padding bytes in the footer are ignored during parsing so they no longer affect validation.

The footer parser rejects unknown blob footer versions instead of silently accepting them, and the registry backend anchors the footer layout so the trailing metadata is located at a fixed position.

Accessor methods for crc and offset fields are renamed for consistency, and the storage cache, prefetch, chunk-group map, build, check, and optimize paths are updated to use the byte-based span and renamed accessors. Documentation and tests are adjusted accordingly.
## Related Issues

## Change Details

## Test Results

## Change Type
- [ ] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist


imeoer:
Two suggestions on the on-disk format design:

1. **Keep `maximum_group_span_block_shift` and its block-based encoding.** This is the maximum padded address-space span of any group, not a fixed group size or its tightly packed payload size. `maximum` makes that distinction explicit, and 4 KiB blocks match the GroupTable addressing and alignment constraints. Changing header byte 24 from `4096 << shift` to `1 << shift` adds no representational capability, but changes the default 8 MiB encoding from 11 to 23 while the format version remains 1. The new reader rejects the old default value, and the old reader rejects the new default value. Byte-oriented in-memory accessors or constructor arguments can still be supported without changing the wire encoding.

2. **Use `compat` / `incompat` feature bits as the compatibility contract, rather than rejecting solely because `version != 1`.** In particular, the footer previously treated the version as informational; adding an exact-version gate prevents otherwise compatible evolution. With a stable base layout, an unknown compat bit should be ignored, an unknown incompat bit must be rejected, and a higher informational version alone should not make an otherwise readable blob fail. Please make this contract explicit and consistent for the footer and metadata; the metadata version gate already exists, so this is also a broader design clarification.

This does not mean accepting arbitrary future layouts: CRC, bounds, and layout validation must remain mandatory. A change that alters how existing fields are interpreted needs an explicit incompat feature and decoder support; unrecognizable framing must still be rejected. Silently redefining an existing field with the same flags is not a compatible extension.

Could we keep the current block-span encoding, retain the CRC/layout validation improvements, and add regression coverage for a higher informational version with unchanged layout, unknown compat/incompat bits, and reading the existing v1 block-span encoding?

